PR #20, opened 2021-01-11
Avoid escapes to heap in frequently used functions

suzaku:
1. Use `strconv.FormatUint` instead of `Sprintf` to avoid escapes to heap
2. Call `Bytes` directly instead of getting a string `Result` and copying to a `[]byte`.


davies (approved):
Great！

PR #23, opened 2021-01-12
fix compile commands

xyb:


davies (approved):

davies:
@xyb Good catch, could you also fix the one in Chinese version?

xyb:
@davies sure, fixed.

PR #25, opened 2021-01-12
Use miniredis to avoid starting a redis server for unit tests

suzaku:


davies:
This is cool!

One question: Does the miniredis have the same sematics comparing to the real redis? For example, the transaction support. If not, it may hide some bugs.

Since it's pretty light way to start redis-server,  and we have not run into problem, so I'd like to keep it.

Once thing we can do, is to launch a redis process rather than depending on external one.

davies:
Also, we need to have more tests to cover the cases on retrying failed request, that will require a real redis.

suzaku:
If we separate unit tests and integration tests, is it OK to use a lightweight mock for unit tests?

suzaku:
Or we can use something like docker-compose and make the test running container depend on a redis container. Either way, the main goal is to make it easier to run tests in the development environment.

davies:
Miniredis support transactions, that's good. Then we may add checking redis config, which use `INFO` and `CONFIG`, they are not supported by Miniredis. so we still need the real redis to run unit test.

Do you have difficulty to install redis-server in your developing enviroment?

suzaku:
No, I can easily start a redis-server with Docker.

davies:
Thanks, we can revisit this in future.

PR #29, opened 2021-01-12
The lock should not be released immediately after it is acquired

suzaku:


davies (approved):

davies:
After a closer look, this is not bug actually, we should remove the lock, right now, i will dead lock. Sorry for that.

PR #36, opened 2021-01-12
Rename the var which conflict with the libary 'fmt'

201341:


cmd/format.go
```diff
@@ -52,12 +52,12 @@ func fixObjectSize(s int) int {
 	return s
 }
 
-func createStorage(fmt *meta.Format) (object.ObjectStorage, error) {
-	blob, err := object.CreateStorage(strings.ToLower(fmt.Storage), fmt.Bucket, fmt.AccessKey, fmt.SecretKey)
+func createStorage(fmat *meta.Format) (object.ObjectStorage, error) {
```
suzaku, line 55: Using `format` may be better in this case, because the word is just two letters longer.

201341: @Suave fix it.

PR #47, opened 2021-01-13
fix broken block cache

davies:
A line was removed by mistake.

This PR also cleanup the code to to download a object.


angristan (approved):
I can confirm data cache has been fixed with this PR. Thanks!

davies:
@angristan Thanks to get back to us.

PR #63, opened 2021-01-14
Check Redis Server configuration

### LINKED ISSUE #16: Check settings on Redis

After connected to Redis, check the following settings on redis:

1. AOF is ON
2. RDB is ON
3. no in cluster mode (cluster_enabled is 0)
4. maxmemory_policy is set to noeviction
5. show a warning if it's not replicated
6. version should be >= 2.2

**Why is this needed**:

Redis is responsible for persistence of metadata, so it should NOT lose any data, otherwise the data in JuiceFS will be lost.

suzaku:
Fix #16.

### DIFF
--- a/pkg/meta/interface.go
+++ b/pkg/meta/interface.go
@@ -142,8 +142,4 @@ type Meta interface {
 	Setlk(ctx Context, inode Ino, owner uint64, block bool, ltype uint32, start, end uint64, pid uint32) syscall.Errno
 
 	OnMsg(mtype uint32, cb MsgCallback)
-
-	// IsConfigSafe returns true if the configuration of the meta server
-	// met the persistence requirement for storing metadata.
-	IsConfigSafe() bool
 }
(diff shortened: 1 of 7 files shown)


pkg/meta/interface.go
```diff
@@ -142,4 +142,8 @@ type Meta interface {
 	Setlk(ctx Context, inode Ino, owner uint64, block bool, ltype uint32, start, end uint64, pid uint32) syscall.Errno
 
 	OnMsg(mtype uint32, cb MsgCallback)
+
+	// IsConfigSafe returns true if the configuration of the meta server
+	// met the persistence requirement for storing metadata.
+	IsConfigSafe() bool
```
davies, line 148: If we find that Redis is misconfigured, we should show an warning or even exit with -1 in inmediaetly, for example, 

```
 if !r. aofEnabled {
   logger.Warnf("AOF is not enabled, you may lose data if Redis is not shutdown properly.")
}
```

All of these checking could be done in pkg/redis/, so we don't need this API here.

davies (approved):

PR #70, opened 2021-01-15
delete xattr data when removing files

### LINKED ISSUE #69: XAttr data still exists in meta server after removing the file.
**What happened**:
The XAttr data are still in redis after removing the file.

The XAttr data of deleted file will also disappear.

After run following commands:
```
# running in macOS Catalina 
./juicefs format localhost test
sudo ./juicefs mount localhost ~/jfs
cp ~/*.jpg ~/jfs
rm -f ~/jfs/*.jpg
```
The keys I would like to see in redis should be:
```
$ redis-cli keys "*" | sort
i1
nextchunk
nextinode
nextsession
sessions
setting
totalInodes
usedSpace
```
But there is a lot of junk XAttr data:
```
$ redis-cli keys "*" | sort
i1
nextchunk
nextinode
nextsession
sessions
setting
totalInodes
usedSpace
x2
x3
x4
x5
x6
x7
x8
```

**Anything else we need to know?**:

**Environment**:
- JuiceFS version (use `./juicefs --version`): 0.9.2 (2021-01-15 d0aa162)
- Cloud provider or hardware configuration running JuiceFS: MacBook Air (13-inch, 2017)
- OS (e.g: `cat /etc/os-release`): macOS 10.15.7
- Kernel (e.g. `uname -a`): Darwin Kernel Version 19.6.0
- Object storage (cloud provider and region): none
- Redis info (version, cloud provider managed or self maintained): Redis 6.0.5 (00000000/0) 64 bit
- Network connectivity (JuiceFS to Redis, JuiceFS to object storage): localhost
- Others: none

xyb:
fix #69

### DIFF
--- a/pkg/redis/redis.go
+++ b/pkg/redis/redis.go
@@ -717,6 +717,7 @@ func (r *redisMeta) Unlink(ctx Context, parent Ino, name string) syscall.Errno {
 		_, err = tx.TxPipelined(c, func(pipe redis.Pipeliner) error {
 			pipe.HDel(c, r.entryKey(parent), name)
 			pipe.Set(c, r.inodeKey(parent), r.marshal(&pattr), 0)
+			pipe.Del(c, r.xattrKey(inode))
 			if attr.Nlink > 0 {
 				pipe.Set(c, r.inodeKey(inode), r.marshal(&attr), 0)
 			} else {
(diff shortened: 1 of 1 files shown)


davies:
@xyb Good catch!  There are other places also need this, rmdir, rename (overwrite a file/dir)

xyb:
@davies I have fixed the rmdir case but did not have enough time to test and fix the rename feature.

p.s. I actually can not see the files except directories in mount point directory on macOS using Finder, which is strange and I don't know how to fix it yet.

davies:
@xyb, I will look into these.

Issue #65, opened 2021-01-14
In daemon mode, the log should be output to syslog by default


zyfjeff:
**What would you like to be added**:

In non-daemon mode, log output is always to stderr, while in daemon mode, log output is to Syslog by default, unless --nosyslog is specified

**Why is this needed**:
I'm currently annoyed by the fact that the log output to Syslog is only available when the -quiet parameter is added.

PR #71, opened 2021-01-15
Optimized logging level by default only logging to syslog in daemon mode

zyfjeff:
This issue fixes three problems:

1. Even for output to Syslog, I think we need to record levels like trace, debug, etc., which is usually used for problem debug
2. Remove duplicate log level code in format and mount.
3. By `--quiet` definition, it should be able to log warnings so its log level should be `warning`, not `error`

### DIFF
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -82,7 +82,7 @@ func mount(c *cli.Context) error {
 	setLoggerLevel(c)
 	if !c.Bool("nosyslog") {
 		// The default log to syslog is only in daemon mode.
-		utils.InitLoggers(c.Bool("b"))
+		utils.InitLoggers(c.Bool("d"))
 	}
 	if c.Args().Len() < 1 {
 		logger.Fatalf("Redis URL and mountpoint are required")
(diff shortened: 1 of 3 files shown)


zyfjeff:
#65

cmd/mount.go
```diff
@@ -80,14 +79,11 @@ func mount(c *cli.Context) error {
...
-	if c.Bool("trace") {
-		utils.SetLogLevel(logrus.TraceLevel)
-	} else if c.Bool("debug") {
-		utils.SetLogLevel(logrus.DebugLevel)
-	} else if c.Bool("quiet") {
-		utils.SetLogLevel(logrus.ErrorLevel)
+	setLoggerLevel(c)
+	if !c.Bool("nosyslog") {
+		// The default log to syslog is only in daemon mode.
+		utils.InitLoggers(c.Bool("b"))
```
davies, line 85: b -> d

davies (changes requested):
The change make sense to me, please fix the `-d` flag

cmd/mount.go
zyfjeff: done

davies (approved):

PR #72, opened 2021-01-15
Support expand file path

zyfjeff:
Support expand file path starting with `~`

### DIFF
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -23,7 +23,9 @@ import (
 	"os"
 	"os/exec"
 	"os/signal"
+	"os/user"
 	"path"
+	"path/filepath"
 	"runtime"
 	"strings"
 	"syscall"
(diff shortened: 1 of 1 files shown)


davies:
@zyfjeff I tried current version, it already can expend path with `~/`, what's the shell you are using?

zyfjeff:
> @zyfjeff I tried current version, it already can expend path with `~/`, what's the shell you are using?

The current version does not work on macOS or it doesn't work on ZSH.

suzaku:
Isn't this something done by your shell?

chnliyong:
> > @zyfjeff I tried current version, it already can expend path with `~/`, what's the shell you are using?
> 
> The current version does not work on macOS or it doesn't work on ZSH.

I tried zsh, it also works in an interactive shell. Are you use it in a script like [this](https://stackoverflow.com/questions/56951712/why-is-zsh-not-able-to-read-tilde-from-a-path-in-a-script)?

zyfjeff:
@chnliyong @davies 
The shell is supposed to do path expansion ahead of time, but the script approach is problematic, and I run it with the startup parameters configured by Goland and it doesn't automatically expand.

zyfjeff:
image

chnliyong:
> @chnliyong @davies
> The shell is supposed to do path expansion ahead of time, but the script approach is problematic, and I run it with the startup parameters configured by Goland and it doesn't automatically expand.

Conventionally, the `~` expansion is done by shell, not the work for application process.
I think one point is that the interactive shell always opened by a logined user, but the application may not run by an logined user(may a system user which may doesn't have home directory).

davies:
@zyfjeff  This is useful improvement, but we still think it's better handled by shell or other tools. Let's close it for now, we can revisit later.

PR #80, opened 2021-01-16
Use bytes.Buffer to speedup logHandle.Format

suzaku:


davies:
How much is the speed up?

In general it don't have much logging, otherwise the logging itself is the problem. With this assumption, we'd not to optimize this corner case, prefer readability rather than raw performance.

PR #91, opened 2021-01-19
mutex to RWMutex

last-saiyan:
updated `sync.Mutex` to `sync.RWMutex`
in the function  `UsedMemory()` we don't do any mutations to the data structures, so i think read lock is better, it can allows multiple readers simultaneously


davies:
sync.RWMutex is much heavier than Mutex, if it's not necessary, we don't use  RWMutex.

Issue #95, opened 2021-01-20
Support directories with millions of files.


davies:
**What would you like to be added**:

Currently, we fetch the attributes of files in single directory with single batch request to Redis, that could be slow or fail, and block other requests. 

We can split those into small batches, for example, 1000 per batch.

**Why is this needed**:

The number of files could  be millions, we don't want people be bited by that.

### Backlog

- [x] Call `MGET` with small batches #110 
- [x] Use `HSCAN` instead of `HGETALL` #128

suzaku:
Any more detail for this issue? In what kind of operation are the file attributes fetched?

davies:
When you do `ls`,  a Readdir() will  be called. When plus is true, it will fetch all the attributes in one batch.

suzaku:
Got it, I can take a try on this one.

suzaku:
If there are millions of files, we might run into trouble when the first `HGetAll` call is reached. Any idea how to work around it?

davies:
Yes. Right now, we don't have a work around, that could be the next challenge.

davies:
The second part is done by #128

PR #110, opened 2021-01-22
Avoid calling mget with massive number of keys in Readdir

suzaku:
This PR is related to #95 .

In the original implementation, `mget` is called once with all the keys which correspond to files in a directory. When there are many files in a directory, this call might block the Redis server process.

In this PR, I changed it to call `mget` in smaller fixed batch. The consequence is that we reduce the chance of blocking the Redis server, but `ReadDir` is made slower when the number of files in a directory exceeds the batch size (which is set to 4096 now).

For small directories, the latency difference is trivial:

* Before
```python
In [24]: %timeit os.listdir("/Users/satoru//jfs/some-files/")
1.26 ms ± 2.57 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

* After
```python
In [25]: %timeit os.listdir("/Users/satoru/jfs/some-files/")
1.27 ms ± 11.3 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

When I run the benchmark in a directory with more than 200,000 files, the new version is obviously slower:

* Before
```python
In [15]: %timeit os.listdir("/Users/satoru/jfs/many-files/")
446 ms ± 2.87 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

* After
```python
In [19]: %timeit os.listdir("/Users/satoru/jfs/many-files/")
488 ms ± 8.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### DIFF
--- a/pkg/meta/redis.go
+++ b/pkg/meta/redis.go
@@ -1200,29 +1200,45 @@ func (r *redisMeta) Readdir(ctx Context, inode Ino, plus uint8, entries *[]*Entr
 		i++
 	}
 	if plus != 0 {
-		batchSize := 40960
+		batchSize := 4096
 		if batchSize > len(*entries) {
 			batchSize = len(*entries)
 		}
-		keysBatch := make([]string, 0, batchSize)
-		for i := 0; i < len(*entries); i += batchSize {
-			end := i + batchSize
-			if end > len(*entries) {
-				end = len(*entries)
+		nEntries := len(*entries)
+		indexCh := make(chan int, 10)
+		go func() {
+			for i := 0; i < nEntries; i += batchSize {
+				indexCh <- i
 			}
-			for _, e := range (*entries)[i:end] {
-				keysBatch = append(keysBatch, r.inodeKey(e.Inode))
-			}
-			rs, _ := r.rdb.MGet(ctx, keysBatch...).Result()
-			for j, re := range rs {
-				if re != nil {
-					if a, ok := re.(string); ok {
-						r.parseAttr([]byte(a), (*entries)[i+j].Attr)
+			close(indexCh)
+		}()
+		var wg sync.WaitGroup
+		for i := 0; i < 2; i++ {
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				keysBatch := make([]string, 0, batchSize)
+				for idx := range indexCh {
+					end := idx + batchSize
+					if end > len(*entries) {
+						end = len(*entries)
 					}
+					for _, e := range (*entries)[idx:end] {
+						keysBatch = append(keysBatch, r.inodeKey(e.Inode))
+					}
+					rs, _ := r.rdb.MGet(ctx, keysBatch...).Result()
+					for j, re := range rs {
+						if re != nil {
+							if a, ok := re.(string); ok {
+								r.parseAttr([]byte(a), (*entries)[idx+j].Attr)
+							}
+						}
+					}
+					keysBatch = keysBatch[:0]
 				}
-			}
-			keysBatch = keysBatch[:0]
+			}()
 		}
+		wg.Wait()
 	}
 	return 0
 }
--- a/pkg/meta/redis.go
+++ b/pkg/meta/redis.go
@@ -1167,10 +1167,6 @@ func (r *redisMeta) Readdir(ctx Context, inode Ino, plus uint8, entries *[]*Entr
 	if err := r.GetAttr(ctx, inode, &attr); err != 0 {
 		return err
 	}
-	vals, err := r.rdb.HGetAll(ctx, r.entryKey(inode)).Result()
-	if err != nil {
-		return errno(err)
-	}
 	*entries = []*Entry{
 		{
 			Inode: inode,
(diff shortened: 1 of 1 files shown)


pkg/meta/redis.go
```diff
@@ -1185,26 +1181,47 @@ func (r *redisMeta) Readdir(ctx Context, inode Ino, plus uint8, entries *[]*Entr
...
+	newAttrs := make([]Attr, len(newEntries))
+	var i int
 	for name, val := range vals {
 		typ, inode := r.parseEntry([]byte(val))
-		*entries = append(*entries, &Entry{
-			Inode: inode,
-			Name:  []byte(name),
-			Attr:  &Attr{Typ: typ},
-		})
+		ent := newEntries[i]
```
davies, line 1193: ent should be a pointer

pkg/meta/redis.go
```diff
@@ -1185,26 +1181,47 @@ func (r *redisMeta) Readdir(ctx Context, inode Ino, plus uint8, entries *[]*Entr
...
+		ent.Inode = inode
+		ent.Name = []byte(name)
+		attr := newAttrs[i]
```
davies, line 1196: attr should be a pointer

pkg/meta/redis.go
```diff
@@ -1185,26 +1181,47 @@ func (r *redisMeta) Readdir(ctx Context, inode Ino, plus uint8, entries *[]*Entr
...
-		for _, e := range *entries {
-			keys = append(keys, r.inodeKey(e.Inode))
-		}
-		rs, _ := r.rdb.MGet(ctx, keys...).Result()
-		for i, re := range rs {
-			if re != nil {
-				if a, ok := re.(string); ok {
-					r.parseAttr([]byte(a), (*entries)[i].Attr)
+		batchSize := 4096
+		if batchSize > len(*entries) {
```
davies, line 1204: not needed

suzaku:
I've changed the code to call `mget` in two goroutines, it turns out to be faster than the original implementation when reading a folder with 300,000 files.

* Before

```python
In [2]: %timeit os.listdir("/Users/satoru/jfs/many-files/")
657 ms ± 10.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [3]: %timeit os.listdir("/Users/satoru/jfs/many-files/")
654 ms ± 9.33 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [4]: %timeit os.listdir("/Users/satoru/jfs/many-files/")
648 ms ± 8.35 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

* After

```python

In [5]: %timeit os.listdir("/Users/satoru/jfs/many-files/")
525 ms ± 10.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [6]: %timeit os.listdir("/Users/satoru/jfs/many-files/")
613 ms ± 5.61 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [7]: %timeit os.listdir("/Users/satoru/jfs/many-files/")
617 ms ± 9.69 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

suzaku:
For folders with less files (100 in my benchmark), it's slower than the original implementation:

* Before

```python
In [18]: %timeit os.listdir("/Users/satoru/jfs/some-files/")
685 µs ± 8.8 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [19]: %timeit os.listdir("/Users/satoru/jfs/some-files/")
666 µs ± 12.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [20]: %timeit os.listdir("/Users/satoru/jfs/some-files/")
656 µs ± 2.74 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

* After

```python
In [15]: %timeit os.listdir("/Users/satoru/jfs/some-files/")
752 µs ± 6.62 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [16]: %timeit os.listdir("/Users/satoru/jfs/some-files/")
748 µs ± 4.29 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [17]: %timeit os.listdir("/Users/satoru/jfs/some-files/")
722 µs ± 38.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

davies:
@suzaku Based on the result from benchmark, we should have a fast path for small directorym, thanks.

davies:
Could you also add a benchmark in Go (for both of small and large directory)?

suzaku:
@davies Is it safe here to use multiple `HSCAN` instead of a huge `HGETALL` here?

davies:
@suzaku It's OK to use HSCAN. Right now, what's behavior for 5 millions files?

suzaku:
If `HSCAN` is OK, we can set up a goroutine to do HSCAN on smaller batches instead of calling `HGETALL` upfront.
I haven't tested 5 million files yet. Creating files can be quite time consuming, I didn't wait that long...

davies:
Let's merge this one first, then optimize the hgetall() later.

PR #115, opened 2021-01-24
Add swift object storage support

morrishe:
1. Add swift object storage support, current only support "v1" auth method

2. The bucket url (--bucket option) format is:   http://BucketName.1.2.3.4[:port]

### DIFF
--- a/pkg/object/swift.go
+++ b/pkg/object/swift.go
@@ -1,0 +1,104 @@
+package object
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/url"
+	"strings"
+
+	"github.com/ncw/swift"
+)
+
+type swiftOSS struct {
+	conn       *swift.Connection
+	region     string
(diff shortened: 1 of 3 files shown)


pkg/object/swift.go
```diff
@@ -0,0 +1,101 @@
...
+
+type swiftOSS struct {
+	conn       *swift.Connection
+	region     string
+	storageUrl string
+	container  string
+}
+
+func (s *swiftOSS) String() string {
+	return fmt.Sprintf("region:[%s], %s/%s", s.region, s.storageUrl, s.container)
```
davies, line 22: short format:   swift://container.region/

pkg/object/swift.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func (s *swiftOSS) Create() error {
+	err := s.conn.ContainerCreate(s.container, nil)
```
davies, line 26: Will this return error if the container exists?

PR #120, opened 2021-01-26
Fix wrong value assignment for StorageConfig.SecretKey

Cosss7:
I can't find the format of commit message.
If I got sth wrong, plz correct me. : )


Cosss7:
Looks like something wrong.

davies:
@Cosss7 The unit test is broken, we will fix that.

The author in commit message does not match with the email in your account, could you update that?

Cosss7:
@davies  fix it

Issue #96, opened 2021-01-20
Comprehensive User Guide


davies:
**What would you like to be added**:

Add a user guide including all the details of command line argument, especially for the object store.

**Why is this needed**:

People may find it difficult to connect JuiceFS with some object store, for example, Ceph, we should have a guide on how to specify the arguments for every object store we support.

### Backlog

- [x] Command reference #119 #122 
- [x] How to setup different object storage #135 #162 
- [x] Best practice to deploy Redis #171 #174 #177 
- [x] Comparison with other projects #179 
- [x] Hadoop SDK user guide #141 #144 #149 #150 #154 #167
- [x] How to auto-mount JuiceFS #163 
- [x] Use JuiceFS on K8s #184 #191
- [x] Cache management #193
- [x] Fault diagnosis and analysis #197 
- [x] Monitoring and Grafana template #202 
- [x] S3 gateway #203 
- [x] More FAQs #198 
- [x] FUSE mount options #215
- [x] Windows client #271

PR #135, opened 2021-01-28
Add doc: How to Setup Object Storage

xiaogaozi:
Ref to #96. More object storage will be added later in separated PR.

### DIFF
--- a/docs/how_to_setup_object_storage.md
+++ b/docs/how_to_setup_object_storage.md
@@ -6,6 +6,8 @@ This is a guide about how to setup object storage when format a volume. Differen
 
 For authentication, the access key and secret key are needed. You could specify them through `--access-key` and `--secret-key` options. Or you can set `ACCESS_KEY` and `SECRET_KEY` environment variables.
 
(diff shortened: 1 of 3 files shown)


docs/how_to_setup_object_storage.md
```diff
@@ -0,0 +1,53 @@
...
+
+## MinIO
+
+[MinIO](https://min.io) is an open source high performance object storage. It is API compatible with Amazon S3. You need set `--storage` option to `minio`. Currently, JuiceFS only supports path-style URI when use MinIO storage. For example (`<endpoint>` may looks like `1.2.3.4:9000`):
+
+```bash
+$ ./juicefs format \
+    --storage minio \
+    --bucket http://<endpoint>/<bucket> \
+    --access-key XXX \
```
davies, line 50: Since we already cover `--access-key` in the beginning, so we don't need to repeat them here.

docs/how_to_setup_object_storage.md
```diff
@@ -0,0 +1,53 @@
+# How to Setup Object Storage
+
+This is a guide about how to setup object storage when format a volume. Different object storage may has different option value. Check the specific object storage for your need.
+
+## Access key and secret key
+
+For authentication, the access key and secret key are needed. You could specify them through `--access-key` and `--secret-key` options. Or you can set `ACCESS_KEY` and `SECRET_KEY` environment variables.
```
davies, line 7: Mention AMI_Role, for example, s3 and oss. When the machine can access S3 or OSS using role, we don't need to specify them.

davies (approved):

PR #175, opened 2021-02-04
s3: use path-style for s3 compatible storage

iliul:


xiaogaozi:
In my opinion, we should support both virtual hosted-style and path-style.

pkg/object/s3.go
```diff
@@ -326,14 +327,16 @@ func newS3(endpoint, accessKey, secretKey string) (ObjectStorage, error) {
 			bucketName = hostParts[0]
 			ep = hostParts[1]
 			region = awsDefaultRegion
+			usePathStyle = true
```
davies, line 330: We should be able to parse a uri like this:  http://host:port/bucket

If the endpoint is a domain, then we don't know it's vhost style or path style, so we can only support path style if the host is an IP address.

PR #176, opened 2021-02-04
fix broken image link

zhyon404:


xiaogaozi:
Could you also fix image link in [`mdtest.md`](https://github.com/juicedata/juicefs/blob/main/docs/en/mdtest.md)?

PR #185, opened 2021-02-16
encrypt data at-rest

davies:
This PR add encryption for data at-rest.


angristan:
Would it be possible to use an ECC-based algorithm instead of RSA for improved performance? 👀

davies:
@angristan Currently, RSA is only used to encrypt the secret per object (about 1ms per object), there is still some room to improve. Can you provide more details on ECC-based algorithm?

PR #221, opened 2021-03-05
prometheus metrics: add namespace prefix

angristan:
Prefix all prometheus metric names with `juicefs_` in order to set a namespace. Otherwise it's hard to know what a metric is referring to on a Prometheus instance with lots of metrics.

From https://prometheus.io/docs/practices/naming/:

> A metric name should have a (single-word) application prefix relevant to the domain the metric belongs to. The prefix is sometimes referred to as namespace by client libraries. For metrics specific to an application, the prefix is usually the application name itself.

### DIFF
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -45,23 +45,23 @@ import (
 
 var (
 	cpu = prometheus.NewGauge(prometheus.GaugeOpts{
-		Name: "cpu_usage",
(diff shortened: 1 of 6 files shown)


davies:
@angristan LGTM, thanks for contributing it.

The CLA is a way to help us governor the project, especially when we need to change the license of it (if AGPL is not the best license), otherwise it will be impossible to do that.

PR #224, opened 2021-03-05
support path-style URL for S3 (#134)

### LINKED ISSUE #134: Support path-style URL for S3 (or S3-compatible) storage

Support path-style URL for S3 (or S3-compatible) storage

**Why is this needed**:

Currently, JuiceFS only support [virtual hosted-style URI](https://docs.aws.amazon.com/AmazonS3/latest/dev/VirtualHosting.html). The difference between virtual hosted-style and path-style is:

- Virtual hosted-style: `https://<bucket>.s3.<region>.amazonaws.com`
- Path-style: `https://s3.<region>.amazonaws.com/<bucket>`

Although AWS would [deprecate](https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story) path-style URI in the future, but some S3-compatible storage still use path-style (e.g. [Ceph RGW](https://docs.ceph.com/en/latest/radosgw/s3/commons)). So we may need support this type URI.

luohy15:
When uri.Path is not empty, parse uri as path-style, otherwise parse uri as virtual hosted-style.

Passed [TestS3](https://github.com/juicedata/juicefs/blob/5a92bdfa6091042cc1fa21076b5069be1ce14af5/pkg/object/object_storage_test.go#L216) with following values:

```
juicefs-image.s3.ap-northeast-1.amazonaws.com
s3.ap-northeast-1.amazonaws.com/juicefs-image
play.minio.io:9000/juicefs
localhost:443/test
```

Closes #134

### DIFF
--- a/pkg/object/s3.go
+++ b/pkg/object/s3.go
@@ -311,11 +311,11 @@ func parseRegion(endpoint string) string {
 }
 
 func newS3(endpoint, accessKey, secretKey string) (ObjectStorage, error) {
+	endpoint = strings.Trim(endpoint, "/")
 	uri, err := url.ParseRequestURI(endpoint)
 	if err != nil {
 		return nil, fmt.Errorf("Invalid endpoint %s: %s", endpoint, err.Error())
 	}
-	hostParts := strings.SplitN(uri.Host, ".", 2)
 
 	var (
 		bucketName string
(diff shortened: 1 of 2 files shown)


pkg/object/s3.go
```diff
@@ -306,39 +323,46 @@ func newS3(endpoint, accessKey, secretKey string) (ObjectStorage, error) {
 		ep         string
 	)
 
-	if len(hostParts) == 1 { // take endpoint as bucketname
+	if len(hostParts) == 1 {
```
davies, line 326: It could be a valid hostname in path-style, for example, http://localhost/bucket

pkg/object/s3.go
```diff
@@ -353,6 +377,8 @@ func newS3(endpoint, accessKey, secretKey string) (ObjectStorage, error) {
 	}
 	if ep != "" {
 		awsConfig.Endpoint = aws.String(ep)
+	}
+	if uri.Path != "" {
```
davies, line 381: For backwards compatibility,  we want to use S3ForcePathStyle with a virtual hosted-style bucket.

pkg/object/s3.go
```diff
@@ -306,39 +323,46 @@ func newS3(endpoint, accessKey, secretKey string) (ObjectStorage, error) {
...
-				endpoint = endpoint[len("dualstack."):]
-			}
-			if endpoint == "amazonaws.com" {
-				endpoint = awsDefaultRegion + "." + endpoint
-			}
-			region = strings.Split(endpoint, ".")[0]
-			if region == "external-1" {
+	} else {
+		// get region or endpoint
+		if uri.Path != "" {
```
davies, line 334: What will happen if the path is `/` ?

luohy15: We can trim `/` before check to prevent this situation

pkg/object/s3.go
luohy15: Added

pkg/object/s3.go
```diff
@@ -293,67 +293,91 @@ func autoS3Region(bucketName, accessKey, secretKey string) (string, error) {
...
 
 	ssl := strings.ToLower(uri.Scheme) == "https"
 	awsConfig := &aws.Config{
-		Region:     aws.String(region),
-		DisableSSL: aws.Bool(!ssl),
-		HTTPClient: httpClient,
+		Region:           aws.String(region),
+		DisableSSL:       aws.Bool(!ssl),
+		HTTPClient:       httpClient,
+		S3ForcePathStyle: aws.Bool(true),
```
davies, line 374: Can we change to previous behavior? At least, no regression.

luohy15: I changed it back. It seems we need to add some regression tests.

davies (approved):
LGTM

Issue #230, opened 2021-03-06
Add new metric: juicefs_object_request_count


angristan:
Basically `juicefs_object_request_data_bytes`:

![screenshot_2021-03-06_15-42-38](https://user-images.githubusercontent.com/11699655/110210602-ae540900-7e92-11eb-9fc8-7be8ec7507ad.png)

But for request count.

I will try to make a PR for this.

angristan:
Closing because it's already here under the name `juicefs_object_request_durations_histogram_seconds_count` 😄

PR #231, opened 2021-03-06
Add new metric: juicefs_object_request_count

angristan:
Close https://github.com/juicedata/juicefs/issues/230.

Preview:

screenshot_2021-03-06_17-41-24
screenshot_2021-03-06_17-41-10
image

### DIFF
--- a/pkg/object/metrics.go
+++ b/pkg/object/metrics.go
@@ -28,6 +28,10 @@ var (
 		Help:    "Object requests latency distributions.",
 		Buckets: prometheus.ExponentialBuckets(0.01, 1.5, 20),
 	}, []string{"method"})
+	reqCount = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "juicefs_object_request_count",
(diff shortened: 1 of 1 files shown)


pkg/object/metrics.go
```diff
@@ -28,6 +28,10 @@ var (
 		Help:    "Object requests latency distributions.",
 		Buckets: prometheus.ExponentialBuckets(0.01, 1.5, 20),
 	}, []string{"method"})
+	reqCount = prometheus.NewCounterVec(prometheus.CounterOpts{
```
davies, line 31: The latency histogram already have a counter for the number of requests:
```
juicefs_object_request_durations_histogram_seconds_count{method="GET"} 1
juicefs_object_request_durations_histogram_seconds_count{method="PUT"} 1
```

angristan: Oh ok, Thanks! The metric name does not make sense to me (durations? seconds? count?), but since it's already here...

Issue #232, opened 2021-03-06
Add new metric: FUSE ops count


angristan:
To complement `juicefs_fuse_written_size_bytes_sum` and `juicefs_fuse_read_size_bytes_sum`, it would be nice to have a metric for the number of operations FUSE is handling.

angristan:
Oops, wrong type of issue 😢 I can't edit it.

angristan:
Closing because it's already here under a different name 🙂

PR #233, opened 2021-03-06
Add metric juicefs_fuse_ops_{read,write}_count

angristan:
Close https://github.com/juicedata/juicefs/issues/232

Add two new counters:

- `juicefs_fuse_ops_read_count`
- `juicefs_fuse_ops_write_count`

image

image

(the left one)

### DIFF
--- a/pkg/vfs/vfs.go
+++ b/pkg/vfs/vfs.go
@@ -63,6 +63,14 @@ var (
 		Help:    "size of write distributions.",
 		Buckets: prometheus.LinearBuckets(4096, 4096, 32),
 	})
+	readCount = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "juicefs_fuse_ops_read_count",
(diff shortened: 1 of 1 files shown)


pkg/vfs/vfs.go
```diff
@@ -63,6 +63,14 @@ var (
 		Help:    "size of write distributions.",
 		Buckets: prometheus.LinearBuckets(4096, 4096, 32),
 	})
+	readCount = prometheus.NewCounter(prometheus.CounterOpts{
```
davies, line 66: the distribution already has counter for read/write
```
juicefs_fuse_read_size_bytes_count 1
juicefs_fuse_written_size_bytes_count 1
```

angristan: Same as for the other PR, the metric name doesn't make sense to me (byte count ?).

PR #234, opened 2021-03-07
Upgrade scs sdk

Arvintian:
scs-go-sdk have a memory leak bug v1.0.0 version,  fix in v1.1.0 version. This pr upgrade scs sdk to v1.1.0. By the way, fix sync subcommand debug flag.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.14
 
 require (
 	cloud.google.com/go v0.39.0
-	github.com/Arvintian/scs-go-sdk v1.0.0
+	github.com/Arvintian/scs-go-sdk v1.1.0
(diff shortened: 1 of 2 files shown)


pkg/sync/config.go
```diff
@@ -59,7 +59,7 @@ func NewConfigFromCli(c *cli.Context) *Config {
 		Manager:     c.String("manager"),
 		BWLimit:     c.Int("bwlimit"),
 		NoHTTPS:     c.Bool("no-https"),
-		Verbose:     c.Bool("verbose"),
```
davies, line 62: Good catch! We have added `--verbose` as an alias to make it consistent with `juicesync`, so we don't need change it here.

Arvintian: I have revoked the fix commit.

PR #239, opened 2021-03-08
add vol_name and mp labels for metrics

chnliyong:
Add `vol_name` and `mp` labels for metrics.

As using [`prometheus`](https://github.com/prometheus/client_golang/blob/master/prometheus) package's wrap fuction [`WrapRegistererWith`](https://github.com/prometheus/client_golang/blob/babeb356a51b1277535260e0122ad8c46240ab96/prometheus/wrap.go#L46) to add labels to registry, all `prometheus.MustRegister()` call should be run after the wrap call.

### DIFF
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -221,7 +221,7 @@ func mount(c *cli.Context) error {
 
 	mntLabels := prometheus.Labels{
 		"vol_name": format.Name,
-		"mp": mp,
+		"mp":       mp,
 	}
 	// Wrap the default registry, all prometheus.MustRegister() calls should be afterwards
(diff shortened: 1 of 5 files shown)


cmd/mount.go
```diff
@@ -218,6 +218,15 @@ func mount(c *cli.Context) error {
 		}
 	}()
 	installHandler(mp)
+
+	mntLabels := prometheus.Labels{
+		"vol_name": format.Name,
+		"mp":       mp,
+	}
+	// Wrap the default registry, all prometheus.MustRegister() calls should be afterwards
```
davies, line 226: Also  `prometheus.Register`

blob = object.WithMetrics(blob)

chnliyong: OK, fix in next commit.

angristan:
thanks!

PR #256, opened 2021-03-16
add metrics for block cache

### LINKED ISSUE #216: Metrics for block cache
We have metrics for FUSE operations and object requests, it's better to have more metrics for block cache. 

1. number of cached blocks
2. total cached size
3. hit rate
4. others

davies:
Closes #216

### DIFF
--- a/pkg/chunk/cached_store.go
+++ b/pkg/chunk/cached_store.go
@@ -31,6 +31,7 @@ import (
 	"github.com/juicedata/juicefs/pkg/compress"
 	"github.com/juicedata/juicefs/pkg/object"
 	"github.com/juicedata/juicefs/pkg/utils"
+	"github.com/prometheus/client_golang/prometheus"
 )
 
 const chunkSize = 1 << 26 // 64M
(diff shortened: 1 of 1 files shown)


pkg/chunk/cached_store.go
```diff
@@ -39,6 +40,23 @@ const SlowRequest = time.Second * time.Duration(10)
...
+		Help: "missed read from cached block",
+	})
+	cacheHitBytes = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "juicefs_blockcache_hit_bytes",
+		Help: "read bytes from cached block",
+	})
+	cacheMissBytes = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "juicefs_blockcache_miss_bytes",
+		Help: "missed bytes from cached block",
+	})
```
angristan, line 59: 👋 This is nitpicking but for the first two I think it should be `juicefs_blockcache_hits` & `juicefs_blockcache_misses` (or `juicefs_blockcache_hit_count` & `juicefs_blockcache_miss_count`)

Right now `hits` is plural but `miss` is singular 

for example: https://github.com/oliver006/redis_exporter/blob/d0597c841d2c9fa30ce8b6ded6251d1994822e27/exporter/exporter.go#L168-L169

PR #259, opened 2021-03-17
support IPv6 address for sftp

chnliyong:
Support IPv6 address for sftp server.

When we _sync_ files to IPv6 sftp server:
`./juicefs sync ./localdir/ user:pass@[IPv6 address]:/remotedir/`

### DIFF
--- a/pkg/object/sftp.go
+++ b/pkg/object/sftp.go
@@ -383,7 +383,7 @@ func newSftp(endpoint, user, pass string) (ObjectStorage, error) {
 	var hostEnd int
 	// IPv6 address like [fe80::148a:d031:823a:fff]
 	if strings.HasPrefix(endpoint, "[") {
-		hostEnd = strings.Index(endpoint, "]") + 1
+		hostEnd = strings.Index(endpoint, "]:") + 1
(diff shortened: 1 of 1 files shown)


pkg/object/sftp.go
```diff
@@ -380,7 +380,17 @@ func (f *sftpStore) ListAll(prefix, marker string) (<-chan *Object, error) {
...
+	// IPv6 address like [fe80::148a:d031:823a:fff]
+	if strings.HasPrefix(endpoint, "[") {
+		hostEnd = strings.Index(endpoint, "]:") + 1
+	} else { // IPv4 address or host name
+		hostEnd = strings.Index(endpoint, ":")
+	}
+	if hostEnd <= 0 {
+		return nil, fmt.Errorf("unable to parse host from endpoint: %s", endpoint)
+	}
+	parts := strings.Split(endpoint[hostEnd:], ":")
```
davies, line 393: Should we use net.SplitHostPort? https://golang.org/pkg/net/#SplitHostPort

chnliyong: The port may not exist like: `[IPv6]:/dir` and we use `22` as default.

Issue #219, opened 2021-03-03
accept named parameters after positional arguments 


davies:
Right now, named parameter after positional arguments are ignored, but they are useful and easy to add.

We can recognize named parameters and re-order them before parsing.

For example, the should work:

```bash
$ ./juicefs mount localhost /jfs -d -q
```

PR #267, opened 2021-03-21
Accept named parameters after positional arguments #219

voidint:


davies:
@voidint The named arguments will be ignored. Can you add a few test cases to verify that?

voidint:
Sorry, it really doesn't work.

davies:
Since this patch does not work, we'd like to close it, so other people can work on it.

Issue #205, opened 2021-02-26
Show the internal information for a file or directory


davies:
We can provide a tool to inspect the internal details of a file or directory:

```
$ ./juicefs info /jfs/syslog
/jfs/syslog:
 inodes: 	280
  directories: 	2
  files: 	278
 chunks: 	278
 length: 	654645174
 size: 		655237120

```

201341:
hi，I will try this, please assign it to me.

davies:
@201341  `juicefs rmr` could be an example to follow.

201341:
one question about the implement is whether i can get the info through redis or just through the os library? If using the os library, the performance isn't good when there are millions files in the directory.

davies:
I think we only need the information for current directory or a file, so performance will not be a problem. The os library will not give us the internal information, so we have to look into Redis, or call a RPC to do that (similar to `juicefs rmr`) .

PR #284, opened 2021-04-06
Add info to show internal infomation for directory or file

201341:
#205 

### DIFF
--- a/pkg/meta/redis.go
+++ b/pkg/meta/redis.go
@@ -1149,6 +1149,18 @@ func (r *redisMeta) Rmr(ctx Context, parent Ino, name string) syscall.Errno {
 	return r.emptyEntry(ctx, parent, name, inode, concurrent)
 }
 
+func (r *redisMeta) Info(ctx Context, parent Ino, summary* Summary) syscall.Errno {
+	if st := r.Access(ctx, parent, 3, nil); st != 0 {
+		return st
+	}
+
+	if st := r.Summary(ctx, parent, summary); st != 0 {
+		return st
+	}
+
+	return 0
+}
+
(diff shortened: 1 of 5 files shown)


201341:
@davies

pkg/meta/redis.go
```diff
@@ -1149,6 +1149,18 @@ func (r *redisMeta) Rmr(ctx Context, parent Ino, name string) syscall.Errno {
 	return r.emptyEntry(ctx, parent, name, inode, concurrent)
 }
 
+func (r *redisMeta) Info(ctx Context, parent Ino, summary* Summary) syscall.Errno {
+	if st := r.Access(ctx, parent, 3, nil); st != 0 {
+		return st
+	}
+
+	if st := r.Summary(ctx, parent, summary); st != 0 {
```
davies, line 1157: The Summary() will walk through all the directories, and will not return the details of chunk/slices of a file.

I think we don't need to a new API, getattr and read should be enough

201341: ok, I will try.  
For the example you give, I don't know the meaning of the info. 
For the file `syslog`
`Inode ` is the inode number of the file or directory.
`directory ` is the directory numbers of the file
`files` is the chunk files? so the directory doesn't have the info?
`chunks` is the same meaning as the `files `?
`length` is the file or dir length which is the same as the results `os.stat` get
`size` is the file size or dir size, which is the same as the results `os.stat` get


```
/jfs/syslog:
 inodes: 	280
  directories: 	2
  files: 	278
 chunks: 	278
 length: 	654645174
 size: 		655237120
```

davies: It's not easy to know the total number of files and sub-dir, we can show the chunks/slices of a file.

davies:
@201341 We would like to have this into 0.12, so I have created #288 to take over this one, can you review it?

201341:
> @201341 We would like to have this into 0.12, so I have created #288 to take over this one, can you review it?

@davies it looks good to me.

PR #297, opened 2021-04-13
Remove redundant break

suzaku:
`resolved` is only set to `true` when `i == len(ss)-1`, but when this
condition holds the loop will terminate without `break`.


davies (approved):
LGTM

PR #298, opened 2021-04-13
Store Lua script in a separate file

suzaku:
So that we can have editor support when writing embeded Lua scripts.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/juicedata/juicefs
 
-go 1.14
+go 1.16
 
 require (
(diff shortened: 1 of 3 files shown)


davies:
@suzaku we should support at least last three versions of Go, so we can't depend on 1.6 right now.

suzaku:
@davies Three versions may be too many for Go. According the the Go Developer Survey 2020, about 60% of the respondents begin evaluating a new version before or within 2 months of release. After all, unlike Python, developers are not blocked by the version of Go in deployment environment.

PR #301, opened 2021-04-14
Use Lua script to speed up path resolving in `FileSystem.lookup`

### LINKED ISSUE #199: Speed up path resolving using Lua script
The Hadoop SDK and S3 gateway are heavy dependent on path resolving (lookup the inode for a path step by step), which is slow if there are many level of directories.

This could be speed up using Lua script to resolve the path in Redis (also check the permissions).

This is similar to the recursive Lookup with permission checking.

suzaku:
Fix #199 .

Per the discussion in #199 , we'll introduce this optimization by first adding a new method `Resolve` to the `Meta` interface.
If there's no way to implement a high performance `Resolve`, a `Meta` implementation can just return `syscall.ENOTSUP`. In this case, the current implementation will be used which only depends on `Meta.Lookup` being implemented.

With `redisMeta`, we can implement this with the help of Lua scripting do the resolving in one Redis request.

### DIFF
--- a/pkg/fs/fs.go
+++ b/pkg/fs/fs.go
@@ -530,9 +530,17 @@ func (fs *FileSystem) RemoveXattr(ctx meta.Context, p string, name string) (err 
 }
 
 func (fs *FileSystem) lookup(ctx meta.Context, p string, followLastSymlink bool) (fi *FileStat, err syscall.Errno) {
+	var inode Ino
+	var attr = &Attr{}
+
+	if err = fs.m.Resolve(ctx, p, followLastSymlink, &inode, attr); err != syscall.ENOTSUP {
+		return
+	}
+
+	// Fallback to the default implementation that calls `fs.m.Lookup` for each directory along the path.
+	// It might be slower for deep directories, but it works for every meta that implements `Lookup`.
 	parent := Ino(1)
 	ss := strings.Split(p, "/")
-	var attr = &Attr{}
 	for i, name := range ss {
 		if len(name) == 0 {
 			continue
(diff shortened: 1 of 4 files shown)


pkg/fs/fs.go
```diff
@@ -530,9 +530,17 @@ func (fs *FileSystem) RemoveXattr(ctx meta.Context, p string, name string) (err
 }
 
 func (fs *FileSystem) lookup(ctx meta.Context, p string, followLastSymlink bool) (fi *FileStat, err syscall.Errno) {
+	var inode Ino
+	var attr = &Attr{}
+
+	if err = fs.m.Resolve(ctx, p, followLastSymlink, &inode, attr); err != syscall.ENOTSUP {
+		return
```
davies, line 537: If err is 0, we should return an FileStat.

suzaku: We also need the last `name` to create the FileStat, maybe I can list it as another pointer parameter to `Resolve`.

davies: The last name could be parsed from `p`.

suzaku: What if we follow a symlink to a different name? This would happen inside the Lua script and we have no idea if it has followed a symlink to a different path, right?

davies: We could stop to resolve if any symlink found.

suzaku: Can't we following symlink in the Lua script?

davies: This script is only used for performance, so it will enough if it can handle the common case (without a symlink).

davies: Also, this is only used for Java SDK, Windows or S3 gateway, there is usually no symlink in these cases.

suzaku:
Since the Lua interpreter embedded in Redis is of version 5.1, we can't use functions like `string.unpack`.

But we can copy the implementation of unpack from other projects, for example: https://github.com/iryont/lua-struct/blob/master/src/struct.lua

davies:
> Since the Lua interpreter embedded in Redis is of version 5.1, we can't use functions like `string.unpack`.
> 
> But we can copy the implementation of unpack from other projects, for example: https://github.com/iryont/lua-struct/blob/master/src/struct.lua

struct.unpack is available in Lua 5.1, right?

PR #328, opened 2021-04-20
Added a option to disable debug agent

### LINKED ISSUE #325: Support to disable listening ports
By default, JuiceFS will listening on `6060`, `6070` and `9567` these 3 ports. `6060` is [pprof](https://golang.org/pkg/net/http/pprof) HTTP server, `6070` is [gops](https://github.com/google/gops) agent, `9567` is Prometheus metrics. We could add an option to disable these ports.

**Why is this needed**:
Open ports and binding to all interfaces may have some security issues.

This issue is from https://github.com/juicedata/juicefs/discussions/320

davies:
This PR also change the default address of metrics to localhost:9567

Closes #325

### DIFF
--- a/cmd/gateway.go
+++ b/cmd/gateway.go
@@ -82,11 +82,6 @@ func gatewayFlags() *cli.Command {
 func gateway(c *cli.Context) error {
 	setLoggerLevel(c)
 	utils.InitLoggers(false)
-	go func() {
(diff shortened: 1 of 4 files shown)


cayolblake:
@davies Apparently `juicefs mount` doesn't know about this parameter so no way to set it from `/etc/fstab`

davies:
Fixed in d749382

PR #349, opened 2021-04-26
store blocks into multiple bucket

davies:
To enable this, a volume should be formatted with `--shards  N` and `--bucket TEMPLATE`, N means the number of buckets and TEMPLATE is used to generated different bucket names, for example, `test%d`.

Then all the blocks will be stored into different buckets based on the hash of object keys.

### DIFF
--- a/pkg/object/sharding.go
+++ b/pkg/object/sharding.go
@@ -43,7 +43,7 @@ func (s *sharded) Create() error {
 
 func (s *sharded) pick(key string) ObjectStorage {
 	h := fnv.New32a()
-	h.Write([]byte(key))
+	_, _ = h.Write([]byte(key))
 	i := h.Sum32() % uint32(len(s.stores))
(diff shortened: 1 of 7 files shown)


pkg/object/sharding.go
```diff
@@ -0,0 +1,203 @@
...
+}
+
+func (s *sharded) ListAll(prefix, marker string) (<-chan Object, error) {
+	heads := &nextObjects{make([]nextKey, 0)}
+	for i := range s.stores {
+		ch, err := ListAll(s.stores[i], prefix, marker)
+		if err != nil {
+			return nil, fmt.Errorf("list %s: %s", s.stores[i], err)
+		}
+		if err == nil {
```
Sandy4999, line 149: this bracket is unnecessary

Sandy4999 (approved):

PR #354, opened 2021-04-27
fix the bug

201341:
-----------------

when Head is wrong, the return value shouldn't be nil

Signed-off-by: shenwenjie <1186093704@qq.com>

### DIFF
--- a/pkg/sync/sync.go
+++ b/pkg/sync/sync.go
@@ -184,7 +184,7 @@ func copyObject(src, dst object.ObjectStorage, obj object.Object) error {
 		in, e := src.Get(key, 0, -1)
 		if e != nil {
 			if _, err := src.Head(key); err != nil {
-				return nil
+				return err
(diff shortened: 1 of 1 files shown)


pkg/sync/sync.go
```diff
@@ -184,7 +184,7 @@ func copyObject(src, dst object.ObjectStorage, obj object.Object) error {
 		in, e := src.Get(key, 0, -1)
 		if e != nil {
 			if _, err := src.Head(key); err != nil {
-				return nil
+				return err
```
davies, line 187: This is a feature rather than a bug. For example, a file is found and deleted from source later, it will be ignored rather than reported as failed.

Have you run into problem because of this behavior?

201341:
Hi, it confused about the data. Found and copied values is the same. And the data isn't copied into dstination, which caused the data loss?

PR #396, opened 2021-05-11
Heartbeat should have consistent date format

polyrabbit:
When setting `heartbeat`, I see it's formatted using **UTC**. Whereas when comparing it, a **local** timezone is used. Which in my case (CST), causes all sessions being deleted.

### DIFF
--- a/pkg/meta/sql.go
+++ b/pkg/meta/sql.go
@@ -1354,7 +1354,7 @@ func (m *dbMeta) cleanStaleSession(sid uint64) {
 func (m *dbMeta) cleanStaleSessions() {
 	// TODO: once per minute
 	var s session
-	rows, err := m.engine.Where("Heartbeat < ?", time.Now().Add(time.Minute*-5)).Rows(&s)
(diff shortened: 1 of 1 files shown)


davies (approved):

polyrabbit:
> @polyrabbit Could you also fix the delfile.Expire ? Which has the same problem.

After testing with sqlite3, I see it indeed has the same problem. But I'm wondering - do other DBs also have the same problem?

`sqlite3` is special because it doesn't understand datetime type, so `xorm` handles it differently:

```go
if dialect.URI().DBType == schemas.SQLITE {
	engine.DatabaseTZ = time.UTC
} else {
	engine.DatabaseTZ = time.Local
}
```

So my question is - it's easy to add a `.UTC()` suffix to every where condition, but will it break other DBs? I don't have mysql on handle for the testing.

In the long term, I suppose it's better to store all datetime types using `timestamp` to avoid all the problems caused by time zone. What do you think? @davies

davies:
@polyrabbit Yes, the timezone issue could be tough, we'd better to use timestamp instead.

PR #408, opened 2021-05-13
Use `join` to fetch node attr and name in one sql query

polyrabbit:


davies:
@polyrabbit Looks great, this is exactly the thing we want to do.

For the case `plus` is zero, then readdir don't need the attributes, so we can keep the first half of original implementation.

polyrabbit:
> @polyrabbit Looks great, this is exactly the thing we want to do.
> 
> For the case `plus` is zero, then readdir don't need the attributes, so we can keep the first half of original implementation.

I didn't see any usage of the `plus` parameter in the original implementation. 

But if `attr` field is not required when `plus` is zero, how about just omitting it when creating an `Entry`? eg:
```go
		&Entry{
			Inode: n.node.Inode,
			Name:  []byte(n.edge.Name),
		}
```

davies:
@polyrabbit We have a mistake in this, the Redis one is correct.

davies:
@polyrabbit we don't need the join if attributes are not needed.

polyrabbit:
> @polyrabbit we don't need the join if attributes are not needed.

@davies Join is removed when `plus` is zero, see commit: https://github.com/juicedata/juicefs/pull/408/commits/cef8ee5718fccd779d42b18773e88569b0709051

pkg/meta/sql.go
```diff
@@ -434,7 +438,9 @@ func (m *dbMeta) Lookup(ctx Context, parent Ino, name string, inode *Ino, attr *
 			// TODO: in SQL
 			if e := m.resolveCase(ctx, parent, name); e != nil {
 				*inode = e.Inode
-				*attr = *e.Attr
+				if attr != nil && e.Attr != nil {
```
davies, line 441: if attr is not nil, but e.Attr is, we need fill the attr using GetAttr()

polyrabbit: OK, that needs another fix. I'll revert this change for now.

pkg/meta/sql.go
```diff
@@ -1336,44 +1342,21 @@ func (m *dbMeta) Readdir(ctx Context, inode Ino, plus uint8, entries *[]*Entry)
...
-	for nodes.Next() {
-		err = nodes.Scan(&n)
-		if err != nil {
-			return errno(err)
-		}
-		attr := new(Attr)
-		m.parseAttr(&n, attr)
+	for _, n := range nodesWithParent {
+		var attr Attr
+		m.parseAttr(&n.node, &attr) // .Type is fetched into node, so attr will always have .Typ filled regardless of plus value
```
davies, line 1355: Without the Join, how do it know edge.Type == node.Type ?

davies, line 1353: parseAttr() will always have `attr.Full = true`, but when `plus==0`, we should have a `attr.Full = false`, and only Typ in attr.

polyrabbit: No, it doesn't know, but isn't it guaranteed in the `mknod` method?
Without `join`, we can only get `.Typ` from the `edge` table, is it reliable?

polyrabbit: OK, will fix it. I have no idea how `attr.Full` is used.

davies: it should be 
```
entry := &Entry{
			Inode: n.node.Inode,
			Name:  []byte(n.edge.Name),
			Attr:  &Attr{},
		}
if plus == 1 {
  entry.Attr.Type = n.edge.Type
} else {
  m.parseAttr(&n.node, entry.Attr)
}
*entries = append(*entries, entry)
```

polyrabbit: Updated

pkg/meta/sql.go
```diff
@@ -1336,45 +1340,26 @@ func (m *dbMeta) Readdir(ctx Context, inode Ino, plus uint8, entries *[]*Entry)
...
-		m.parseAttr(&n, attr)
-		*entries = append(*entries, &Entry{
-			Inode: Ino(n.Inode),
-			Name:  names[n.Inode],
-			Attr:  attr,
-		})
+		if plus != 0 {
+			m.parseAttr(&n.node, entry.Attr)
+		} else {
+			entry.Attr.Typ = n.node.Type // xorm fetches .Type is into node, actually it's from the edge table
```
davies, line 1360: Why not use `n.edge.Type` here?

pkg/meta/sql.go
```diff
@@ -67,6 +66,11 @@ type node struct {
 	Parent Ino
 }
 
+type nodeParent struct {
+	node `xorm:"extends"`
+	edge `xorm:"extends"`
```
davies, line 71: we should put `edge` before `node`, otherwise xorm will be confused.

pkg/meta/sql.go
polyrabbit: Because xorm fills type value into `node` not `edge`😂. I left a comment here as a reminder.

davies: It's because we have node before edge in nodeParent. Reverse that, it just works as expected.

polyrabbit: Are you sure? I put node before edge because when joining two tables, their order matters according to xorm's doc. I'll try some new ways tomorrow.

davies: Yes, I tried, it works perfectly.

polyrabbit: I reversed `node` and `edge` fields, **ALSO** reversed are their orders in the `inner join` sql, so it won't break conventions in xorm.

pkg/meta/sql.go
```diff
@@ -1336,45 +1340,26 @@ func (m *dbMeta) Readdir(ctx Context, inode Ino, plus uint8, entries *[]*Entry)
...
 		return errno(err)
 	}
-	defer nodes.Close()
-	for nodes.Next() {
-		err = nodes.Scan(&n)
-		if err != nil {
-			return errno(err)
+	for _, n := range nodes {
+		entry := &Entry{
+			Inode: n.Inode,
```
davies, line 1353: It's ambiguous, `n.edge.Inode` is better

davies (approved):
LGTM

pkg/meta/sql.go
polyrabbit: OK, I found it too error-prone to embed both `node` and `edge` in one struct. People may confuse where to get attributes like `inode`/`type`.

So I re-defined `namedNode` by:
```go
type namedNode struct {
	node `xorm:"extends"`
	Name string
}
```

Now `namedNode` contains everything about a node - attrs and name, without any ambiguousness. What do you think @davies

PR #439, opened 2021-05-24
uid file and grouping file unified configuration

### LINKED ISSUE #423: Global uid mapping for Java SDK
Currently, we use local uid mapping for Java SDK, which requires all the machine have consistent mapping.

We can have a configurable global mapping, for example, a file in JuiceFS, to easy the burden to maintain uid mapping across the cluster.

tangyoupeng:
Added `juicefs.users` to specify the uids for all users in JuiceFS, which should be an URI to a text file in the following format:

```
username1:1
username2:2
```

Added `juicefs.groups` to specify the id of groups also the grouping of users, which is an URI to a file in the following format

```
group1:1:user1,user2
group2:2:user3,user4
```

Closes #423


tangyoupeng:
#423

tangyoupeng:
#423

davies (approved):

PR #475, opened 2021-05-31
Support sticky bit in Java SDK

### LINKED ISSUE #472: Support Sticky bit
Sticky bit is used to protect file from deleting or renamed by others rather than the owner or super-user in a public read/writable directory.

see https://en.wikipedia.org/wiki/Sticky_bit for details

davies:
When default permission is enabled on FUSE, the kernel will check for sticky bit, and JuiceFS don't need to do anything rather than storing the sticky bit.

For Java SDK and S3 gateway, JuiceFS need to check permission, so we added the permission check for all meta engines, so they will be active for Java SDK and S3 gateway.

Closes #472

### DIFF
--- a/pkg/meta/redis.go
+++ b/pkg/meta/redis.go
@@ -1412,7 +1412,7 @@ func (r *redisMeta) Rename(ctx Context, parentSrc Ino, nameSrc string, parentDst
 					r.Unlock()
 				}
 			}
-			if ctx.Uid() != 0 && sattr.Mode&01000 != 0 && ctx.Uid() != sattr.Uid && ctx.Uid() != tattr.Uid {
+			if ctx.Uid() != 0 && dattr.Mode&01000 != 0 && ctx.Uid() != dattr.Uid && ctx.Uid() != tattr.Uid {
 				return syscall.EACCES
 			}
 		} else {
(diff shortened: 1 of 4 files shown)


pkg/meta/redis.go
```diff
@@ -1389,6 +1412,9 @@ func (r *redisMeta) Rename(ctx Context, parentSrc Ino, nameSrc string, parentDst
 					r.Unlock()
 				}
 			}
+			if ctx.Uid() != 0 && sattr.Mode&01000 != 0 && ctx.Uid() != sattr.Uid && ctx.Uid() != tattr.Uid {
```
Sandy4999, line 1415: for target's parent, should check dattr not sattr?

pkg/meta/sql.go
```diff
@@ -1197,6 +1213,12 @@ func (m *dbMeta) Rename(ctx Context, parentSrc Ino, nameSrc string, parentDst In
 					m.Unlock()
 				}
 			}
+			if ctx.Uid() != 0 && spn.Mode&01000 != 0 && ctx.Uid() != spn.Uid && ctx.Uid() != dn.Uid {
```
Sandy4999, line 1216: same here, check dpn

Sandy4999 (approved):

PR #482, opened 2021-06-02
add client information for each session

Sandy4999:


Sandy4999:
Better to split to small PRs. And a new `session` subcommand should be more comprehensive than adding options to `status`.

PR #488, opened 2021-06-03
version: record version when formatting

Sandy4999:


Sandy4999:
Use meta engine version instead of release version.

PR #510, opened 2021-06-10
cmd: add `dump` command to dump metadata to a file and `load` to load from a file

### LINKED ISSUE #50: Backup tool for meta

Provide a tool to dump metadata as JSON format, then we can have other tool to assemble them and get data back.

**Why is this needed**:

In worsest case, if we lose Redis, we'd like to have a tool to get most of data back from S3.

Sandy4999:
close #50

Todo:
- [x] calculate slice refs when loading
- [ ] customize json marshal to stream encode/decode; and to make the dumped file more compact


cmd/main.go
```diff
@@ -81,6 +81,7 @@ func main() {
 			profileFlags(),
 			statusFlags(),
 			warmupFlags(),
+			backupFlags(),
```
davies, line 84: We could have dump/load as separate command

Sandy4999: c139306

cmd/dump.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+	defer fp.Close()
+	return m.DumpMeta(fp)
+}
+
+func dumpFlags() *cli.Command {
+	return &cli.Command{
+		Name:      "dump",
+		Usage:     "dump JuiceFS metadata into a standalone file",
+		ArgsUsage: "META-ADDR [FILE (default: juicefs-metadata.dump)]",
```
davies, line 48: no default value for FILE

pkg/meta/types.go
```diff
@@ -0,0 +1,129 @@
...
+	Uid       uint32 `json:"uid"`
+	Gid       uint32 `json:"gid"`
+	Atime     int64  `json:"atime"`
+	Mtime     int64  `json:"mtime"`
+	Ctime     int64  `json:"ctime"`
+	Atimensec uint32 `json:"atimensec"`
+	Mtimensec uint32 `json:"mtimensec"`
+	Ctimensec uint32 `json:"ctimensec"`
+	Nlink     uint32 `json:"nlink"`
+	Length    uint64 `json:"length,omitempty"`
```
davies, line 50: 0 length of a file should not be omitted

pkg/meta/sql.go
```diff
@@ -2281,3 +2283,314 @@ func (m *dbMeta) RemoveXattr(ctx Context, inode Ino, name string) syscall.Errno
 		return err
 	}))
 }
+
+func (m *dbMeta) dumpEntry(name string, inode Ino) (*DumpedEntry, error) {
+	e := &DumpedEntry{Name: name, Inode: inode}
+	return e, m.txn(func(s *xorm.Session) error {
```
davies, line 2289: Should we use repeatable read isolation level?

davies:
@Sandy4999 Could you add a few unit tests for them?  loading a sample file into redis/db, verify some functions, then dump it again, and compare the  results.

cmd/dump.go
Sandy4999: b936bea

pkg/meta/types.go
Sandy4999: b936bea

davies (approved):
pending on unit test

PR #526, opened 2021-06-16
meta/test: use random db name for sqlite tests

### LINKED ISSUE #516: Flaky test: TestStickyBitSQL
Sometimes, the TestStickyBitSQL case failed without any reason, we should look into that.

```
=== RUN   TestStickyBitSQL
[xorm] [info]  2021/06/12 08:43:27.207240 PING DATABASE sqlite3
FAIL	github.com/juicedata/juicefs/pkg/meta	1.467s
```

Sandy4999:
close #516

### DIFF
--- a/pkg/meta/sql_test.go
+++ b/pkg/meta/sql_test.go
@@ -17,13 +17,13 @@
 package meta
 
 import (
-	"os"
 	"testing"
+
+	"github.com/juicedata/juicefs/pkg/utils"
 )
 
 func TestSQLClient(t *testing.T) {
-	os.Remove("test.db")
-	m, err := newSQLMeta("sqlite3", "test.db", &Config{})
+	m, err := newSQLMeta("sqlite3", utils.RandString(8)+".db", &Config{})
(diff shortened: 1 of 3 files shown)


pkg/meta/sql_test.go
```diff
@@ -17,13 +17,13 @@
...
-	"os"
 	"testing"
+
+	"github.com/juicedata/juicefs/pkg/utils"
 )
 
 func TestSQLClient(t *testing.T) {
-	os.Remove("test.db")
-	m, err := newSQLMeta("sqlite3", "test.db", &Config{})
+	m, err := newSQLMeta("sqlite3", utils.RandString(8)+".db", &Config{})
```
davies, line 26: Should we use `io/ioutil.TempFile()`?

suzaku: +1, it seems to be a good use case for `TempFile`.

Sandy4999: e326154

PR #528, opened 2021-06-17
cache attribute and chunks for open files

### LINKED ISSUE #454: Keep page cache in kernel
For file read before, the kernel may have the page cache, we should use the page cache if the is not modified since last open.

davies:
For not-modified files, we should keep the page cache in kernel, also cache the chunks to speed up read.

For any changes to the data, the mtime of file will be updated in microseconds, so we can check the mtime to know whether the file is changed or not. If not, we will keep the cache for chunks and pages in kernel, otherwise invalidate them.

And, we introduce a new option called `OpenCache`, which tell the client to don't check the mtime of a file before that, then JuiceFS will not know any change in the past period. So `OpenCache` should be set based on the change interval of files.

TODO: add command option for OpenCache

Closes #454

### DIFF
--- a/pkg/meta/openfile.go
+++ b/pkg/meta/openfile.go
@@ -19,10 +19,12 @@ type openfiles struct {
 }
 
 func newOpenFiles(expire time.Duration) *openfiles {
-	return &openfiles{
+	of := &openfiles{
 		expire: expire,
 		files:  make(map[Ino]*openFile),
 	}
+	go of.cleanup()
+	return of
 }
 
 func (o *openfiles) cleanup() {
@@ -80,9 +82,6 @@ func (o *openfiles) Close(ino Ino) bool {
 	of, ok := o.files[ino]
 	if ok {
 		of.refs--
-		if of.refs <= 0 && of.lastCheck.Add(o.expire).Before(time.Now()) {
-			delete(o.files, ino)
-		}
 		return of.refs <= 0
 	}
 	return true
--- a/pkg/meta/openfile.go
+++ b/pkg/meta/openfile.go
@@ -7,8 +7,8 @@ import (
 
 type openFile struct {
 	attr      Attr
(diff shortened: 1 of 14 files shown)


pkg/meta/openfile.go
```diff
@@ -0,0 +1,155 @@
...
+	o.Lock()
+	defer o.Unlock()
+	of, ok := o.files[ino]
+	if !ok {
+		of = &openFile{}
+		of.chunks = make(map[uint32][]Slice)
+		o.files[ino] = of
+	} else if attr != nil && attr.Mtime == of.attr.Mtime && attr.Mtimensec == of.attr.Mtimensec {
+		attr.KeepCache = true
+	} else {
```
Sandy4999, line 69: attr.KeepCache = false?

pkg/meta/openfile.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+	return true
+}
+
+func (o *openfiles) Check(ino Ino, attr *Attr) bool {
+	o.Lock()
+	defer o.Unlock()
+	of, ok := o.files[ino]
+	if ok && time.Since(of.lastCheck) < o.expire {
+		*attr = of.attr
```
Sandy4999, line 95: attr must not be nil

Sandy4999 (approved):

pkg/meta/openfile.go
davies: This is the only place to assign KeepCache (initialized as zero).

daimashusheng:
May I ask, if I only use juicefs on linux, is it possible to do same thing without the openfile.go? I just need to mount filesystem without setting ExplicitDataCacheControl of MountOptions and always set FOPEN_KEEP_CACHE when I open a file, because linux kernel can invalidates the page cache by detecting the length or the modification time of the file has changed.

PR #537, opened 2021-06-19
invalidate cache fo open files

davies:
1. Invadate the cached attribute for open files after some fields changed.

2. When the cached mtime is updated by GetAttr(), future Open() will assume the mtime is consistent with page cache, but not, so we should mark the cached attribute so that future Open() will not keep the cache.

3. When a file is compacted by other clients, the chunk was deleted, so we should invalidate the chunk cache after a temporary read failure.

### DIFF
--- a/pkg/meta/redis.go
+++ b/pkg/meta/redis.go
@@ -1996,6 +1996,11 @@ func (r *redisMeta) NewChunk(ctx Context, inode Ino, indx uint32, offset uint32,
 	return errno(err)
 }
 
+func (r *redisMeta) InvalidateChunkCache(ctx Context, inode Ino, indx uint32) syscall.Errno {
+	r.of.InvalidateChunk(inode, indx)
+	return 0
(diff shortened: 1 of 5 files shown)


pkg/meta/redis.go
```diff
@@ -1049,6 +1049,7 @@ func (r *redisMeta) Fallocate(ctx Context, inode Ino, mode uint8, off uint64, si
 }
 
 func (r *redisMeta) SetAttr(ctx Context, inode Ino, set uint16, sugidclearmode uint8, attr *Attr) syscall.Errno {
+	defer func() { r.of.InvalidateChunk(inode, 0xFFFFFFFE) }()
```
xiaogaozi, line 1052: Update `attr` in `of` also?

davies: yes

Sandy4999 (approved):

PR #541, opened 2021-06-21
Docs: Add "Metadata Cache in Client"

xiaogaozi:
Chinese translation will in another PR.


davies (changes requested):
As we discussed offline, the document should provide clear view for end-user, the implement details should be excluded.

PR #542, opened 2021-06-21
meta: support PostgreSQL as the meta engine

### LINKED ISSUE #450: Support PostgresQL
We should add PG driver and test it with PG, make sure JuiceFS work well with it.

davies:
Closes #450

### DIFF
--- a/pkg/meta/sql_test.go
+++ b/pkg/meta/sql_test.go
@@ -45,7 +45,7 @@ func TestSQLClient(t *testing.T) {
 }
 
 func TestMySQLClient(t *testing.T) {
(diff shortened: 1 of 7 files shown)


pkg/meta/sql_test.go
```diff
@@ -74,6 +74,36 @@ func TestMySQLClient(t *testing.T) {
 	testCaseIncensi(t, m)
 }
 
+func TestPostgresQLClient(t *testing.T) {
+	m, err := newSQLMeta("postgres", "postgres://localhost:5432/davies?sslmode=disable", &Config{})
```
Sandy4999, line 78: use a general test database name, e.g. 'juicefs' or 'test'

Sandy4999 (approved):

PR #545, opened 2021-06-21
cmd/bench: add -p/--threads options to run benchmarks in parallel

### LINKED ISSUE #357: Run benchmark in parallel
`juicefs bench` run basic benchmark using single threads, which can be run in parallel to know  how it works under concurrent workload.

Sandy4999:
close #357

### DIFF
--- a/cmd/benchmark.go
+++ b/cmd/benchmark.go
@@ -40,7 +40,7 @@ type Benchmark struct {
 	threads        int
 }
 
-func writeFiles(prefix string, fCount, bSize, bCount int, cost chan<- bool) {
+func writeFiles(prefix string, fCount, bSize, bCount int, progress chan<- bool) {
 	for i := 0; i < fCount; i++ {
 		fname := fmt.Sprintf("%s.%d", prefix, i)
 		fp, err := os.OpenFile(fname, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0644)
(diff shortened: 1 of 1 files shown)


chnliyong:
Is `-p`(parallel) for short more suitable(`-t` may intuitively means test). In addition, `sync` sub command also use `-p` for short (`--threads` for long option), keep the option name consistent.

cmd/benchmark.go
```diff
@@ -330,6 +320,12 @@ func benchmarkFlags() *cli.Command {
 				Value: 100,
 				Usage: "number of small files",
 			},
+			&cli.IntFlag{
+				Name:    "threads",
+				Aliases: []string{"p"},
+				Value:   3,
```
davies, line 326: The default should be 1, to be backwards compatible.

cmd/benchmark.go
```diff
@@ -37,127 +37,125 @@ type Benchmark struct {
...
-	}
-	defer file.Close()
-	for i := 0; i < blockCount; i++ {
-		start := time.Now()
-		if _, err = file.Write(buf); err != nil {
-			return nil, err
-		}
-		if i == blockCount-1 {
-			file.Close()
+func writeFiles(prefix string, fCount, bSize, bCount int, cost chan<- bool) {
```
davies, line 43: cost -> progress

davies (changes requested):

cmd/benchmark.go
Sandy4999: ebf2379

cmd/benchmark.go
Sandy4999: ebf2379

davies (approved):

PR #551, opened 2021-06-22
cmd/dump: dump metadata into STDOUT

davies:
This PR change to FILE argument of dump/load to be optional, then STDOUT/STDIN will be used, then we can use it like this:

```bash
$ ./juicefs dump 127.0.0.1/ | ./juicefs load 127.0.0.1/2
```

### DIFF
--- a/docs/en/command_reference.md
+++ b/docs/en/command_reference.md
@@ -525,8 +525,10 @@ dump metadata into a JSON file
 #### Synopsis
 
 ```
-juicefs dump [command options] META-ADDR FILE
+juicefs dump [command options] META-ADDR [FILE]
 ```
+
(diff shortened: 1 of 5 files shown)


docs/en/command_reference.md
```diff
@@ -534,5 +536,7 @@ load metadata from a previously dumped JSON file
 #### Synopsis
 
 ```
-juicefs load [command options] META-ADDR FILE
+juicefs load META-ADDR [FILE]
```
Sandy4999, line 539: better keep "[command options]" here for consistency

Sandy4999 (approved):

Sandy4999:
the chinese doc should be updated as well

PR #558, opened 2021-06-23
cmd/dump&load: encode/decode json in streams

Sandy4999:


pkg/meta/types.go
```diff
@@ -15,6 +15,30 @@
...
+const (
+	jsonIndent    = "  "
+	jsonWriteSize = 64 << 10
+)
+
+type bufWriter struct {
+	bufio.Writer
+}
+
+func (bw *bufWriter) write(s string) {
```
Sandy4999, line 36: @davies used a wrapper with panic to simplify code, any opposition to this?

davies: have this in `writeJSON` will simpler:
```
write = func(s string) {
        if _, err := bw.WriteString(s); err != nil {
		panic(err)
	}
}
```

PR #573, opened 2021-06-29
docs: add docs for WebDAV

davies:


docs/en/how_to_setup_object_storage.md
```diff
@@ -578,6 +579,23 @@ $ ./juicefs format \
     localhost test
 ```
 
+## WebDAV <span id='webdav'></span>
+
+[WebDAV](https://en.wikipedia.org/wiki/WebDAV) is an extension of the Hypertext Transfer Protocol (HTTP)
+that not allows clients to perform remote Web content authoring operations. For a storage that speaks WebDAV, JuiceFS can use
```
xyb, line 585: "not allows"?

davies: copied from Wikipedia :(

PR #575, opened 2021-06-29
logging into file in background mode

### LINKED ISSUE #574: Redirect stdout/stderr to a file in daemon mode
Currently, stdout and stderr are redirected to /dev/null in daemon mode, so we can't find the logging if it crashes.

davies:
Closes #574

### DIFF
--- a/cmd/mount_unix.go
+++ b/cmd/mount_unix.go
@@ -94,10 +94,9 @@ func mount_flags() []cli.Flag {
 			Usage: "disable syslog",
 		},
 		&cli.StringFlag{
-			Name:    "logfile",
-			Aliases: []string{"l"},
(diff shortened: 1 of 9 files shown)


cmd/mount_unix.go
```diff
@@ -70,8 +71,14 @@ func makeDaemon(name, mp string) error {
 				}
 			}
 		}
+		var err error
+		logfile := c.String("logfile")
+		attrs.Stdout, err = os.OpenFile(logfile, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0755)
```
Sandy4999, line 76: x is not needed for logfile, use mode 0644

Sandy4999 (approved):

PR #607, opened 2021-07-08
Docs: update min required Go version to 1.15

xiaogaozi:


xiaogaozi:
@davies Also update `go.mod`?

davies:
@xiaogaozi no, we don't need go 1.15+ to compile

davies:
Right now, we recommend Go 1.15+, but 1.14 still works.

Issue #580, opened 2021-06-30
Support Transactional Key-Value database as metadata engines


Sandy4999:
**What would you like to be added**:
Support transactional key-value databases as metadata engines.
- [TiKV](https://pkg.go.dev/github.com/tikv/client-go/v2@v2.0.0-alpha)
- [FoundationDB](https://pkg.go.dev/github.com/apple/foundationdb/bindings/go/src/fdb)

**Why is this needed**:
High reliablity and hopefully good performace.

**Progress**:
TiKV as metadata engine has been supported; more tests required.

TODO:

- [x] #614 
- [x] #615
- [x] #616 
- [x] #630 
- [x] #617
- [ ] #618
- [x] #666

Sandy4999:
- All keys of TKV (Transactional Key-Value Database) have a filesystem-wide prefix: `jfs` + `<fs-name>`
- Inodes in TKV should be little-endian encoded 

|               | Redis                                            | SQL                         | TKV                                                      |
| ------------- | ------------------------------------------------ | --------------------------- | -------------------------------------------------------- |
| Node          | i$inode -> attr                                  | inode -> type, mode...      | A + $inode + i -> attr                                   |
| Dir           | d$inode -> {name -> (inode,type)}                | inode, name -> inode, type  | A + $inode + d + name -> (inode, type)                   |
| File          | c$inode_$indx -> [slice]                         | inode, index -> [slice]     | A + $inode + c + $indx -> [slice]                        |
| Symlink       | s$inode -> target                                | inode -> target             | A + $inode + s -> target                                 |
| Xattr         | x$inode -> {name -> value}                       | inode, name -> value        | A + $inode + x + name -> value                           |
| Flock         | lockf$inode -> {$sid_$owner -> ltype}            | inode, sid, owner -> ltype  | A + $inode + f + sid + owner -> ltype                    |
| Plock         | lockp$inode -> {$sid_$owner -> Plocks}           | inode, sid, owner -> Plocks | A + $inode + p + sid + owner -> Plocks                   |
| Sessions      | sessions -> [$sid -> heartbeat]                  | sid -> heartbeat, info      | SH + $sid -> heartbeat                                   |
| SessionInfos  | sessionInfos -> {$sid -> info}                   | N/A                         | SI + $sid  -> info                                       |
| sustained     | session$sid -> [$inode]                          | sid, inode                  | SS + $sid -> [$inode]                                    |
[...]

PR #610, opened 2021-07-08
meta: add support for TiKV as meta engine

Sandy4999:
First PR based on TiKV for #580

By default KV meta is not compiled; to enable it, run `make juicefs.tikv`


Sandy4999:
Client-go for TiKV need Go 1.16 to compile

davies (approved):

PR #622, opened 2021-07-10
speedup listing on file

davies:
file.Readdir() calls lstat() on every file, so it's slow on huge directory.

This PR changed to use file.ReadDir(), which does not call lstat(), then we can return some of the files earlier.

### DIFF
--- a/pkg/object/file.go
+++ b/pkg/object/file.go
@@ -178,15 +178,18 @@ func walk(path string, info os.FileInfo, walkFn filepath.WalkFunc) error {
 		return nil
 	}
 
-	infos, err := readDirSorted(path)
+	entries, err := readDirSorted(path)
 	if err != nil {
 		return walkFn(path, info, err)
 	}
 
-	for _, fi := range infos {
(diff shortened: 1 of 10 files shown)


davies:
This feature require Go 1.16 :(

pkg/object/file.go
```diff
@@ -212,37 +215,47 @@ func Walk(root string, walkFn filepath.WalkFunc) error {
...
-			f, _ := os.Stat(filepath.Join(dirname, fi.Name()))
-			fi = &mInfo{fi.Name(), f}
-			fis[i] = fi
-		}
-		if fi.IsDir() {
-			fis[i] = &mInfo{fi.Name() + dirSuffix, fi}
+			fi, _ := os.Stat(filepath.Join(dirname, e.Name()))
+			name := e.Name()
+			if fi.IsDir() {
+				name = e.Name() + dirSuffix
```
SandyXSD, line 252: name += dirSuffix

SandyXSD (approved):

PR #633, opened 2021-07-13
implement deleteChunk/deleteFile for TKV engine

xuhui-lu:


pkg/meta/tkv.go
```diff
@@ -1674,13 +1672,70 @@ func (m *kvMeta) CopyFileRange(ctx Context, fin Ino, offIn uint64, fout Ino, off
...
+			for _, val := range vals {
+				lens = append(lens, len(val)/sliceBytes)
+			}
+			return nil
+		})
+		for i, len := range lens {
+			if len == 0 {
+				continue
+			}
+			idx := binary.BigEndian.Uint32(keys[i][10:])
```
davies, line 1730: 10 -> len(m.prefix)+10

pkg/meta/tkv.go
```diff
@@ -1674,13 +1672,70 @@ func (m *kvMeta) CopyFileRange(ctx Context, fin Ino, offIn uint64, fout Ino, off
...
+	if err != syscall.Errno(0) {
+		return fmt.Errorf("delete slice from chunk %s fail: %s, retry later", key, err)
+	}
+	return nil
+}
 
-func (m *kvMeta) deleteFile(inode Ino, length uint64) {}
+func (m *kvMeta) deleteFile(inode Ino, length uint64) {
+	var indx uint32
+	for uint64(indx)*ChunkSize < length {
```
davies, line 1711: tx.scanRange() to find all they chunks

pkg/meta/tkv.go
```diff
@@ -1674,13 +1672,70 @@ func (m *kvMeta) CopyFileRange(ctx Context, fin Ino, offIn uint64, fout Ino, off
...
+	key := m.chunkKey(inode, indx)
+	var slices []*slice
+	err := m.txn(func(tx kvTxn) error {
+		buf := tx.get(key)
+		slices = readSliceBuf(buf)
+		tx.set(key, []byte{})
+		for _, s := range slices {
+			r := tx.incrBy(m.sliceKey(s.chunkid, s.size), -1)
+			if r < 0 {
+				m.deleteSlice(s.chunkid, s.size)
```
davies, line 1698: This may block, so we could like to start a goroutine

pkg/meta/tkv.go
```diff
@@ -1910,11 +1910,29 @@ func (m *kvMeta) CopyFileRange(ctx Context, fin Ino, offIn uint64, fout Ino, off
...
-func (m *dbMeta) deleteChunk(inode Ino, indx uint32) error {}
 */
 
+func (m *kvMeta) deleteChunk(inode Ino, indx uint32) error {
+	key := m.chunkKey(inode, indx)
+	var slices []*slice
+	err := m.txn(func(tx kvTxn) error {
+		buf := tx.get(key)
+		slices = readSliceBuf(buf)
+		tx.set(key, []byte{})
```
davies, line 1921: tx.dels(key)

pkg/meta/tkv.go
```diff
@@ -1940,7 +1958,33 @@ func (m *kvMeta) deleteSlice(chunkid uint64, size uint32) {
 	}
 }
 
-func (m *kvMeta) deleteFile(inode Ino, length uint64) {}
+func (m *kvMeta) deleteFile(inode Ino, length uint64) {
+	keys, err := m.scanKeys(m.chunkKey(inode, 0))
```
davies, line 1962: the prefix should be  `m.fmtKey("A", inode, "C")`

pkg/meta/tkv.go
xuhui-lu: hmm, to be aligned with redis implementation, should we leave the key with empty value there? Though i think there is no big difference.

pkg/meta/tkv.go
```diff
@@ -1940,7 +1958,33 @@ func (m *kvMeta) deleteSlice(chunkid uint64, size uint32) {
...
+
+	var lens []int
+	m.txn(func(tx kvTxn) error {
+		vals := tx.gets(keys...)
+		for _, val := range vals {
+			lens = append(lens, len(val)/sliceBytes)
+		}
+		return nil
+	})
+	for i, length := range lens {
```
davies, line 1976: All the returned keys should be a valid chunk, we should iterate on that.

pkg/meta/tkv.go
davies: they will be leaked in TiKV. Redis will cleanup the empty keys, but other KV will not.

pkg/meta/tkv.go
```diff
@@ -1940,7 +1958,26 @@ func (m *kvMeta) deleteSlice(chunkid uint64, size uint32) {
...
 }
 
-func (m *kvMeta) deleteFile(inode Ino, length uint64) {}
+func (m *kvMeta) deleteFile(inode Ino, length uint64) {
+	keys, err := m.scanKeys(m.fmtKey("A", inode, "C"))
+	if err != nil {
+		logger.Warnf("delete chunks of inode %d: %s", inode, err)
+		return
+	}
+	var vals [][]byte
```
davies, line 1967: the `vals` is not used

pkg/meta/tkv.go
```diff
@@ -1940,7 +1958,26 @@ func (m *kvMeta) deleteSlice(chunkid uint64, size uint32) {
...
+	m.txn(func(tx kvTxn) error {
+		vals = tx.gets(keys...)
+		return nil
+	})
+	for i := range vals {
```
davies, line 1972: use `keys` should be the same

davies (approved):

PR #636, opened 2021-07-13
meta/tkv: add dump/load functions

### LINKED ISSUE #616: Implement dump/load for KV meta
- [x] DumpMeta
- [x] LoadMeta

Sandy4999:
close #616

### DIFF
--- a/pkg/meta/tkv_tikv.go
+++ b/pkg/meta/tkv_tikv.go
@@ -172,6 +172,14 @@ func (c *tikvClient) name() string {
 	return "TiKV"
 }
 
+func (c *tikvClient) isEmpty() bool {
+	tx, err := c.client.Begin()
+	if err != nil {
+		panic(err)
+	}
+	return tx.Len() == 0
+}
+
 func (c *tikvClient) txn(f func(kvTxn) error) error {
 	tx, err := c.client.Begin()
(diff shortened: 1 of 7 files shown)


pkg/meta/tkv_tikv.go
```diff
@@ -172,6 +172,19 @@ func (c *tikvClient) name() string {
 	return "TiKV"
 }
 
+func (c *tikvClient) isEmpty() bool {
+	tx, err := c.client.Begin()
+	if err != nil {
+		panic(err)
+	}
+	it, err := tx.Iter(nil, nil)
```
davies, line 180: Check with the prefix, we could use the tx.exists()

Sandy4999: Prefix is optional

Sandy4999: Nvm, we can also modify `nextKey`

davies: We should give a warning if prefix is empty

davies: 0xFF could be used when prefix is empty

Sandy4999: Isn't that reserved for internal keys?

PR #645, opened 2021-07-14
support TiKV as object store

davies:


Sandy4999:
It works well for simple tests, but we have to be very careful when use both TxnClient & RawKVClient at the same time, e.g. `Scan` method of RawKVClient can find deleted keys of TxnClient

PR #680, opened 2021-07-23
feat: add progress bar for fsck

zhijian-pro:
I  add a progress bar use [mpb](https://github.com/vbauerster/mpb). 
I suggest that all about the progress bar use the [mpb](https://github.com/vbauerster/mpb) implement.

### DIFF
--- a/cmd/fsck.go
+++ b/cmd/fsck.go
@@ -17,6 +17,9 @@ package main
 
 import (
 	"fmt"
+	"github.com/sirupsen/logrus"
+	"github.com/vbauerster/mpb/v7"
+	"github.com/vbauerster/mpb/v7/decor"
 	"strings"
 	"time"
 
@@ -66,7 +69,6 @@ func fsck(ctx *cli.Context) error {
 	}
 	logger.Infof("Data use %s", blob)
 
-	logger.Infof("Listing all blocks ...")
 	blob = object.WithPrefix(blob, "chunks/")
 	objs, err := osync.ListAll(blob, "", "")
 	if err != nil {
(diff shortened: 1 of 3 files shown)


davies:
@zhijian-pro Thanks for work on this, mbp looks promising, we should use that in all place once we find a good way to use it.

I tried it locally, and the final logging looks like this:
```
2021/07/23 20:19:10.632682 juicefs[39094] <INFO>: Listing all blocks 5557 / 5557 [=========================] 100 %
2021/07/23 20:19:10.632799 juicefs[39094] <INFO>: Found 5520 blocks (308848977 bytes)
2021/07/23 20:19:10.634512 juicefs[39094] <INFO>: Listing all slices done ╢▌▌▌▌▌▌▌▌▌▌▌▌▌▌▌▌▌▌▌▌▌▌▌▌▌▌▌▌▌▌▌▌╟ 100 %
2021/07/23 20:19:10.634835 juicefs[39094] <INFO>: Used by 6 slices (249246 bytes)

```

Overall, the progress bar should be added to the steps which take longer time, for example:

1. List objects from object store, but we don't know the total number of them, so show the current found objects without a bar should be enough.
2. List all the slices from database (m.ListSlices) we should add a fallback to update progress bar in m.ListSlices).
3. Compare the two sets, this one could be the fastest, maybe not worth a progress bar (nice to have).

zhijian-pro:
@davies   I apologize for adding the progress bar to where it didn't take longer, I didn't understand the project well enough.
Now I have a question：
1.  If meta in redis , get total of the slices  action is also expensive , should use redis  command `keys c*_*`.   the command  `scan` cannot get total at once .  but if cannot get total at once ,progress bar unable be implement.  useing redis  command `keys c*_*` Is it acceptable?
2. If meta not in redis，tikv and sql  can get the total at once.

zhijian-pro:
@davies   maybe not worth a progress bar ？ 😂

davies:
@zhijian-pro I think we should show some progress (maybe not a bar), better than quiet blocking.

zhijian-pro:
Kapture 2021-07-24 at 17 44 34
@davies  How does this work dynamically? If the program is currently ListSlices, there will be a dynamic icon spinning behind it, and if the ListSlices are finished, the icon will disappear.
In addition, there is no performance loss because the total is not obtained.
Different types of metadata are displayed and handled in the same way

But for the ListSlices interface I added a parameter `bar *mpb.Bar ` which can be passed as nil when the status display is not needed  

Is this solution acceptable?

PR #690, opened 2021-07-28
meta: when creating an existed file, open it instead

Sandy4999:


pkg/meta/tkv.go
```diff
@@ -1246,7 +1245,27 @@ func (m *kvMeta) mknod(ctx Context, parent Ino, name string, _type uint8, mode,
...
+				foundType, foundIno = entry.Attr.Typ, entry.Inode
+			}
+		}
+		if foundIno != 0 {
+			if _type == TypeFile {
+				a = tx.get(m.inodeKey(foundIno))
+				if a != nil {
+					m.parseAttr(a, attr)
+				} else {
+					*attr = Attr{Typ: foundType, Parent: parent} // corrupt entry
```
davies, line 1263: attr could be nil

davies (changes requested):
attr could be niml

davies (approved):

PR #702, opened 2021-08-03
cmd: add `stats` command to show runtime stats

### LINKED ISSUE #665: Add a tool to show stats in real time
similar to dstat

Sandy4999:
close #665

### DIFF
--- a/cmd/stats.go
+++ b/cmd/stats.go
@@ -16,15 +16,12 @@
 package main
 
 import (
-	"encoding/json"
 	"fmt"
 	"os"
 	"path"
-	"sort"
 	"strings"
 	"time"
 
-	"github.com/juicedata/juicefs/pkg/meta"
 	"github.com/juicedata/juicefs/pkg/utils"
 	"github.com/mattn/go-isatty"
 	"github.com/urfave/cli/v2"
@@ -37,8 +34,8 @@ const (
 	YELLOW
 	BLUE
 	MAGENTA
-	// CYAN
-	// GRAY
+	CYAN
+	WHITE
 )
 
 const (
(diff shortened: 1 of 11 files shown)


pkg/meta/redis_metrics.go
```diff
@@ -27,9 +27,15 @@ var (
 		Name: "transaction_restart",
 		Help: "The number of times a transaction is restarted.",
 	})
+	opDist = prometheus.NewHistogram(prometheus.HistogramOpts{
```
davies, line 30: Should we just the existing `txDist`?

cmd/stats.go
```diff
@@ -0,0 +1,381 @@
...
+		}
+		if tick%watcher.interval == 0 {
+			watcher.printDiff(start, current, false)
+			start = current
+		} else {
+			watcher.printDiff(last, current, true)
+		}
+		last = current
+		tick++
+		<-ticker.C
```
davies, line 349: nit:  `time.Sleep(time.Second)` works the same.

cmd/stats.go
```diff
@@ -0,0 +1,381 @@
...
+		logger.Fatalf("path %s is not a mount point", mp)
+	}
+
+	watcher := &statsWatcher{
+		tty:      !ctx.Bool("nocolor") && isatty.IsTerminal(os.Stdout.Fd()),
+		interval: ctx.Uint("interval"),
+		path:     path.Join(mp, ".stats"),
+	}
+	watcher.buildSchema(ctx.String("schema"), ctx.Uint("verbosity"))
+	watcher.formatHeader()
```
davies, line 328: we should show the header for every 40 lines

pkg/meta/redis_metrics.go
davies: Then the existing grafana should just works.

cmd/stats.go
Sandy4999: use a standalone ticker will be more accurate in case `readStats` or `printDiff` takes considerable time

pkg/meta/redis_metrics.go
Sandy4999: I think they are two different things: one is meta operation, the other one is engine write transaction, which is more likely to be the trouble maker. Also txRestart is a relational item with txDist, but not that close to opDist.

Sandy4999: Grafana should be compatible since txDist is still there with no modification

cmd/stats.go
davies: Sounds good

pkg/meta/redis_metrics.go
davies: We care more about on all the meta operations rather than some write operation, the transaction is kind of implementation details, that's why we prefer the operation latency rather than transaction details, same in Grafana.

Once we have opDist to cover all the operations, txDist is less useful, so I'd suggest to combine them together (the txDist is almost identical to  opDist for write operations). 

The txRestart is also useful to know the conflict between update opeartions.

davies (approved):

PR #728, opened 2021-08-12
meta: sql: Fix delete Bug

201341:
Signed-off-by: shenwenjie <1186093704@qq.com>

### DIFF
--- a/pkg/meta/sql.go
+++ b/pkg/meta/sql.go
@@ -2205,7 +2205,7 @@ func (m *dbMeta) deleteChunk(inode Ino, indx uint32) error {
 				return err
 			}
 		}
-		n, err := ses.Delete(&c)
+		n, err := ses.Delete(chunk{Inode: c.Inode, Indx: c.Indx})
 		if err == nil && n == 0 {
 			err = fmt.Errorf("chunk %d:%d changed, try restarting transaction", inode, indx)
 		}
@@ -2240,6 +2240,7 @@ func (m *dbMeta) deleteFile(inode Ino, length uint64) {
 	for _, indx := range indexes {
 		err = m.deleteChunk(inode, indx)
 		if err != nil {
+			logger.Debugf("deleteChunk inode %d index %d error: %s", inode, indx, err)
 			return
 		}
 	}


cmd/gc.go
```diff
@@ -217,7 +217,7 @@ func gc(ctx *cli.Context) error {
 		if obj.IsDir() {
 			continue
 		}
-		if obj.Mtime().After(maxMtime) || obj.Mtime().Unix() == 0 {
+		if obj.Mtime().UnixNano() > maxMtime.UnixNano() || obj.Mtime().Unix() == 0 {
```
davies, line 220: Why this?

Time.After() is implemented using Unix timestamp
```
func (t Time) After(u Time) bool {
	if t.wall&u.wall&hasMonotonic != 0 {
		return t.ext > u.ext
	}
	ts := t.sec()
	us := u.sec()
	return ts > us || ts == us && t.nsec() > u.nsec()
}
```

The bug may came from how we create the obj.Time() with wrong timezone.

pkg/meta/sql.go
```diff
@@ -2205,7 +2205,7 @@ func (m *dbMeta) deleteChunk(inode Ino, indx uint32) error {
 				return err
 			}
 		}
-		n, err := ses.Delete(&c)
+		n, err := ses.Delete(chunk{Inode: c.Inode, Indx: c.Indx})
```
davies, line 2208: Good catch!

cmd/gc.go
201341: when the backend storage is minio, The obj.Mtime() may return the UTC timezone, and time.Now() return current timezone

davies: That does not matter, After() works even t and u in different timezone.

And we use Time.After() in unit test [1], and MinIO pass the unit tests, so I think this is not the problem.

[1] https://github.com/juicedata/juicefs/blob/main/pkg/object/object_storage_test.go#L109

davies:
@201341 Can you revert the last commit, so we can merge the bugfix sooner?

201341:
> @201341 Can you revert the last commit, so we can merge the bugfix sooner?

Done.

davies (approved):

PR #730, opened 2021-08-13
Delay upload staging file when  writeback enabled with new parameter delay-upload

allwefantasy:
When user set writebackMode to `upload-on-full`, the system will not async upload staging data to object store
unless staging data fails.

with the support of writebackMode ,we can add more strategies e.g.  `upload-on-delay-1-hour`  in later to make the user have more control on when to upload local disk/mem data to object store to avoid waste of bandwith.

The `upload-on-full` is more suitable for shuffle operation. Since the shuffle data is short life-time, will be delete even in seconds, and the engine will only read the data where you write , so no need to upload the data to object store if the disk is not out.

There are two points we may concern:

1. The juicefs client will clean the cache, so will the strategy of cache evict affect the staging data?

No, for now , the strategy of cache evict will ignore the cache item link to staging data which the size is set zero, and the it will not occupy cache quota.

2. How to clean the staging data?

The app (spark/flink) will response for  cleaning the staging data. Notice that if the app crash down(or shutdown) unexpected, the data  become orphaned, and no app will access them again.  Also the meta storage also have the garbage file infos.


davies:
JuiceFS should always upload the block when the disk is full, so we don't need to specify that.  Rather than that, what's should happen when the disk is full?

Since the new written data is more likely to be read in future, so we'd like to have them in disk, rather than having some older data (which could be read and wasted) occupy the disk space. So it's better to upload the older large block to object store and leave the space for new written data.

In order to do that, we can 1) have a configurable period of time (2 hours) to delay the upload, and the blocks are very possible wasted or has longer life cycle, so we move them into object store to save space for new written data.  2) when the free space is under than some ratio (10%), we move some old and large blocks into object store to save space for new data.

allwefantasy:
> JuiceFS should always upload the block when the disk is full, so we don't need to specify that. Rather than that, what's should happen when the disk is full?
> 
> Since the new written data is more likely to be read in future, so we'd like to have them in disk, rather than having some older data (which could be read and wasted) occupy the disk space. So it's better to upload the older large block to object store and leave the space for new written data.
> 
> In order to do that, we can 1) have a configurable period of time (2 hours) to delay the upload, and the blocks are very possible wasted or has longer life cycle, so we move them into object store to save space for new written data. 2) when the free space is under than some ratio (10%), we move some old and large blocks into object store to save space for new data.

In shuffle operation, it seems no need to upload the local disk data to object store if the disk is not full. Once the disk full（under than some ratio）, the we can move the oldest block into object store.   so `have a configurable period of time (2 hours) to delay the upload` is designed for  we just want to make sure we always upload the data to object store, and this bring into correspondence with the original behavior?

allwefantasy:
According to your suggestion,   we can add a new parameter :

```
delay-upload = 2h
```

and this parameter only available when the `writeback` is enabled.

And we start a new thread to check if  the disk free radio is under some value then scan the cache  , find the cache items who match the following conditions:

1. The block size is bigger then some value.
2. The atime is after now exceed 2h.

davies:
+1， we should start a new thread to scan the stagging blocks.

PR #733, opened 2021-08-13
Show repo diagram in Architecture section

suzaku:


suzaku:
@davies Shall we skip the `docs` directory when generating the diagram?

davies:
This diagram is cool, but is not that useful for new comers, I'd like to not put it in the readme, since it's already really long.

davies:
Maybe we could have a page on "how to read the code" in the docs, which could have this diagram.

PR #736, opened 2021-08-15
Delay upload staging file when writeback enabled 

allwefantasy:
According to the discussion in #730 , we add a new parameter to delay upload staging file.

### DIFF
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -113,16 +113,21 @@ func mount(c *cli.Context) error {
 	prometheus.DefaultRegisterer = prometheus.WrapRegistererWith(metricLabels,
 		prometheus.WrapRegistererWithPrefix("juicefs_", prometheus.DefaultRegisterer))
 
+	if !c.Bool("writeback") && c.IsSet("delay-upload"){
+		logger.Warnf("writeback should be set true before congiure deploy-upload")
+	}
+
 	chunkConf := chunk.Config{
 		BlockSize: format.BlockSize * 1024,
 		Compress:  format.Compression,
 
-		GetTimeout: time.Second * time.Duration(c.Int("get-timeout")),
-		PutTimeout: time.Second * time.Duration(c.Int("put-timeout")),
-		MaxUpload:  c.Int("max-uploads"),
-		Writeback:  c.Bool("writeback"),
-		Prefetch:   c.Int("prefetch"),
-		BufferSize: c.Int("buffer-size") << 20,
+		GetTimeout:  time.Second * time.Duration(c.Int("get-timeout")),
+		PutTimeout:  time.Second * time.Duration(c.Int("put-timeout")),
+		MaxUpload:   c.Int("max-uploads"),
+		Writeback:   c.Bool("writeback"),
+		DelayUpload: c.Duration("delay-upload"),
+		Prefetch:    c.Int("prefetch"),
+		BufferSize:  c.Int("buffer-size") << 20,
 
 		CacheDir:       c.String("cache-dir"),
 		CacheSize:      int64(c.Int("cache-size")),
@@ -131,6 +136,7 @@ func mount(c *cli.Context) error {
 		CacheFullBlock: !c.Bool("cache-partial-only"),
 		AutoCreate:     true,
 	}
+
 	if chunkConf.CacheDir != "memory" {
 		ds := utils.SplitDir(chunkConf.CacheDir)
 		for i := range ds {
@@ -292,6 +298,11 @@ func clientFlags() []cli.Flag {
 		&cli.BoolFlag{
 			Name:  "writeback",
 			Usage: "upload objects in background",
+		},
+		&cli.DurationFlag{
+			Name:  "delay-upload",
+			Value: time.Duration(0),
+			Usage: "upload objects with delay time.Valid time units are \"ns\", \"us\" (or \"µs\"), \"ms\", \"s\", \"m\", \"h\"",
 		},
 		&cli.StringFlag{
 			Name:  "cache-dir",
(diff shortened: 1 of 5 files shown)


pkg/chunk/cached_store.go
```diff
@@ -604,6 +611,7 @@ type Config struct {
 	Compress       string
 	MaxUpload      int
 	Writeback      bool
+	DelayUpload    string
```
davies, line 614: It's better to use number seconds here, move the parse into cmd/

pkg/chunk/cached_store.go
```diff
@@ -614,6 +622,34 @@ type Config struct {
 	Prefetch       int
 }
 
+func (c *Config) ValidateParams() {
+	if !c.Writeback && c.DelayUpload != "" {
+		logger.Fatalf("writeback should be set true before congiure deploy-upload")
```
davies, line 627: The DelayUpload will be ignore if writeback is false, so Warning() should be better than Fatalf()

pkg/chunk/cached_store.go
```diff
@@ -755,6 +797,142 @@ func parseObjOrigSize(key string) int {
...
+				if int64(freed) >= goal {
+					break
+				}
+			}
+		}()
+	}
+}
+
+func (store *cachedStore) cleanupDelayStaging() {
+	innerStores := store.bcache.(*cacheManager).stores
```
davies, line 890: It's better to move these into cacheManager rather than pull in the details into cachedStore.

pkg/chunk/cached_store.go
allwefantasy: `time.Duration` is better?  I will change to time.Duration

pkg/chunk/cached_store.go
allwefantasy: Done!

pkg/chunk/cached_store.go
allwefantasy: You means  we can add getStores in CacheManager?

pkg/chunk/cached_store.go
```diff
@@ -755,6 +770,94 @@ func parseObjOrigSize(key string) int {
...
+						store.pendingMutex.Unlock()
+					}()
+					fileInfo := innerStore.keys[key]
+					isTimeout := time.Now().Unix()-int64(fileInfo.atime) > int64(store.conf.DelayUpload.Seconds())
+					if isTimeout {
+						freed += store.uploadStagingFile(key, stagingPath)
+					}
+
+				}(key, innerStore.stagePath(key))
+				if int64(freed) >= goal {
```
Sandy4999, line 853: it looks like **goal** is the target space used, which is not comparable with **freed**

Sandy4999 (changes requested):

cmd/mount.go
```diff
@@ -293,6 +325,11 @@ func clientFlags() []cli.Flag {
 			Name:  "writeback",
 			Usage: "upload objects in background",
 		},
+		&cli.StringFlag{
```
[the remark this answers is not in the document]
allwefantasy: Done

pkg/chunk/cached_store.go
```diff
@@ -755,6 +770,94 @@ func parseObjOrigSize(key string) int {
...
+			for _, key := range keys {
+				store.currentUpload <- true
+				go func(key string, stagingPath string) {
+					defer func() {
+						<-store.currentUpload
+					}()
+
+					defer func() {
+						store.pendingMutex.Lock()
+						delete(store.pendingKeys, key)
```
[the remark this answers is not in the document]
allwefantasy: Done

pkg/chunk/cached_store.go
```diff
@@ -755,6 +770,88 @@ func parseObjOrigSize(key string) int {
...
+	} else {
+		goal = toFree
+	}
+	return goal
+}
+
+func (store *cachedStore) cleanDelayStaging() {
+	cacheStoreKeys := make(map[*cacheStore][]string)
+
+	for key, _ := range store.pendingKeys {
```
Sandy4999, line 821: lock required here

pkg/chunk/cached_store.go
```diff
@@ -742,6 +749,14 @@ func NewCachedStore(storage object.ObjectStorage, config Config) ChunkStore {
 			return float64(used)
 		}))
 	go store.uploadStaging()
+	if store.conf.CacheDir != "memory" {
+		go func() {
+			for {
+				store.cleanDelayStaging()
+				time.Sleep(time.Second)
```
Sandy4999, line 756: 1s is too short if there are many pending keys, in which case the **scan** step of `cleanDelayStaging` may cost considerable time & resource

pkg/chunk/cached_store.go
allwefantasy: If the delete/insert on pendingKeys will not make the foreach fails , then no need to lock it.
We can check later

allwefantasy:
## How to test

Format a new file system:

```shell
./juicefs format \
	--storage file \
	--bucket /tmp/jfs \
	redis://192.168.31.95:6379/1 \
	test-filesystem

```

mount:

```shell
./juicefs mount  --cache-dir /tmp/m1:/tmp/m2 --cache-size 10m --writeback --delay-upload 2m redis://192.168.31.95:6379/2 /tmp/jack
```

Write small file (<1m) to `/tmp/jack`.

Check /tmp/jfs, there should have no blocks.

Then file with 9m to `/tmp/jack`, after 3m ,

Check /tmp/jfs, there should have blocks.

allwefantasy:
the logical to upload staging files with two steps:
1) scan pendingkeys, atime > delay-upload，upload it.
2) when disk free < cache-freeRatio *1.5, use TwoRandomEvict algorithm to choose left keys which bock  should be uploaded until the disk free > cache-freeRatio *1.5.  In this case, we will ignore delay-upload since we have aready upload all atime > delay-upload.

Sandy4999 (approved):

davies (approved):

PR #747, opened 2021-08-19
cmd/warmup: limit request size under 128 KiB

Sandy4999:


Sandy4999:
An alternative fix is to move the `handleInternalMsg` from Write to Read, then we can wait for all data received before triggering the command

Sandy4999:
fixed by #752

PR #755, opened 2021-08-20
object: Fix the bucket bug

201341:
If the bucket contains dot, the s3 bucketName and ep will be
parsed error

### DIFF
--- a/pkg/object/s3.go
+++ b/pkg/object/s3.go
@@ -355,7 +355,7 @@ func newS3(endpoint, accessKey, secretKey string) (ObjectStorage, error) {
 		}
 	} else {
 		// [BUCKET].[ENDPOINT]
-		hostParts := strings.SplitN(uri.Host, ".", 2)
+		hostParts := strings.Split(uri.Host, ".")
 		if len(hostParts) == 1 {
 			// take endpoint as bucketname
 			bucketName = hostParts[0]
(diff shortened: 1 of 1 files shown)


davies:
The bucket name should not have dot in it. Can you really have a dot in the name?

201341:
> The bucket name should not have dot in it. Can you really have a dot in the name?

For example, I can create the bucket name `test.test` on minio or ceph.

davies:
I'd suggest not to do that, it may introduce problem later.

201341:
> I'd suggest not to do that, it may introduce problem later.

I'm not sure about what problem it will cause, but it is useful when sync to ceph using s3

pkg/object/s3.go
```diff
@@ -374,8 +374,9 @@ func newS3(endpoint, accessKey, secretKey string) (ObjectStorage, error) {
 				region = parseRegion(endpoint)
 			} else {
 				// compatible s3
-				bucketName = hostParts[0]
-				ep = hostParts[1]
+				l := len(hostParts)
+				bucketName = strings.Join(hostParts[:l-4], ".")
```
davies, line 378: This assume that the endpoint has 4 parts, which could be IP (ipv4 or ipv6)  or DNS, it will not always have 4 parts.

davies:
@201341 S3 does not allow "." in the bucket name, so we could have the same assumption for S3 compatible storage.

201341:
> @201341 S3 does not allow "." in the bucket name, so we could have the same assumption for S3 compatible storage.

Ok, Got it! I will close this pr.

davies:
@201341 "." is allowed but not recommended

```
For best compatibility, we recommend that you avoid using dots (.) in bucket names, except for buckets that are used only for static website hosting. If you include dots in a bucket's name, you can't use virtual-host-style addressing over HTTPS, unless you perform your own certificate validation. This is because the security certificates used for virtual hosting of buckets don't work for buckets with dots in their names.
```

After some googling, you can find lots of issues around using dot in S3 bucket name, so we still suggest that don't put dot in the name.

PR #761, opened 2021-08-23
hadoop: fix read stats bug

tangyoupeng:


sdk/java/src/main/java/io/juicefs/JuiceFileSystemImpl.java
```diff
@@ -853,6 +853,7 @@ public synchronized int read(long pos, ByteBuffer b) throws IOException {
       int got;
       if (b.hasArray()) {
         got = read(pos, b.array(), b.position(), b.remaining());
+        statistics.incrementBytesRead(-got);
```
zzcclp, line 856: move this line after `if (got <= 0)`?

PR #784, opened 2021-09-01
remove broken file/dir

davies:
In case of broken metadata, the attributes of inode could be lost, then we should be able to remove the broken file/directory.

### DIFF
--- a/pkg/meta/redis.go
+++ b/pkg/meta/redis.go
@@ -1387,6 +1387,7 @@ func (r *redisMeta) Unlink(ctx Context, parent Ino, name string) syscall.Errno {
 		pattr.Ctime = now.Unix()
 		pattr.Ctimensec = uint32(now.Nanosecond())
 		attr = Attr{}
+		opened = false
 		if rs[1] != nil {
 			r.parseAttr([]byte(rs[1].(string)), &attr)
 			attr.Ctime = now.Unix()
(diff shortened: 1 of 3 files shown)


pkg/meta/redis.go
```diff
@@ -1387,11 +1387,14 @@ func (r *redisMeta) Unlink(ctx Context, parent Ino, name string) syscall.Errno {
...
-		if ctx.Uid() != 0 && pattr.Mode&01000 != 0 && ctx.Uid() != pattr.Uid && ctx.Uid() != attr.Uid {
-			return syscall.EACCES
+		if rs[1] != nil {
+			r.parseAttr([]byte(rs[1].(string)), &attr)
+			attr.Ctime = now.Unix()
+			attr.Ctimensec = uint32(now.Nanosecond())
+			if ctx.Uid() != 0 && pattr.Mode&01000 != 0 && ctx.Uid() != pattr.Uid && ctx.Uid() != attr.Uid {
+				return syscall.EACCES
+			}
+			attr.Nlink--
```
SandyXSD, line 1397: better print a warning log if is nill

SandyXSD:
https://github.com/juicedata/juicefs/blob/main/pkg/meta/redis.go#L1418
maybe check attr here: `opened && attr != Attr{}`, or just `opened && attr.Full == true`

davies:
@SandyXSD When the attribute is missing, user can not open the file, so open should be false.

PR #797, opened 2021-09-07
disable check of object storage using env

zwwhdls:
Inside CSI controller and node pod, we can't setup the configuration for Ceph from secrets, so we can disable the check for it in node and controller pod.

### DIFF
--- a/cmd/format.go
+++ b/cmd/format.go
@@ -204,8 +204,10 @@ func format(c *cli.Context) error {
 		logger.Fatalf("object storage: %s", err)
 	}
 	logger.Infof("Data uses %s", blob)
-	if err := test(blob); err != nil {
-		logger.Fatalf("Storage %s is not configured correctly: %s", blob, err)
+	if os.Getenv("NO_CHECK_OBJECT_STORAGE") != "" {
+		if err := test(blob); err != nil {
+			logger.Fatalf("Storage %s is not configured correctly: %s", blob, err)
+		}
 	}
 
 	err = m.Init(format, c.Bool("force"))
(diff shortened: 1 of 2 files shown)


davies:
@zwwhdls We should also add one for `mount`, to remove a subdir in controller node of CSI

cmd/format.go
```diff
@@ -201,8 +201,12 @@ func format(c *cli.Context) error {
 	}
 
 	blob, err := createStorage(&format)
-	if err != nil && !noCheck {
-		logger.Fatalf("object storage: %s", err)
+	if err != nil {
+		if noCheck {
+			logger.Warnf("object storage: %s", err)
```
davies, line 206: blob is nil, it will crash later.

I think we should disable the check in ceph

pkg/object/ceph.go
```diff
@@ -207,11 +208,13 @@ func newCeph(endpoint, cluster, user string) (ObjectStorage, error) {
 	if err != nil {
 		return nil, fmt.Errorf("Can't create connection to cluster %s for user %s: %s", cluster, user, err)
 	}
-	if err := conn.ReadDefaultConfigFile(); err != nil {
-		logger.Fatalf("Can't read default config file: %s", err)
-	}
-	if err := conn.Connect(); err != nil {
-		return nil, fmt.Errorf("Can't connect to cluster %s: %s", cluster, err)
+	if !os.Getenv("NO_CHECK_OBJECT_STORAGE") == "" {
```
davies, line 211: !=

davies:
@zwwhdls When format a volume, it will test the object store, we should also disable that (or no panic)?

PR #842, opened 2021-09-16
meta: support lookup '.' and '..'

SandyXSD:


pkg/meta/tkv.go
```diff
@@ -885,41 +885,42 @@ func (m *kvMeta) resolveCase(ctx Context, parent Ino, name string) *Entry {
 }
 
 func (m *kvMeta) Lookup(ctx Context, parent Ino, name string, inode *Ino, attr *Attr) syscall.Errno {
+	if inode == nil || attr == nil {
```
davies, line 888: attr could be nil

SandyXSD: I didn't find any caller to pass a nil attr :<

PR #855, opened 2021-09-22
keep existing LZ4 algr when formatted with default algr.

davies:
`lz4` is the previous default compress algrithm, and we changed that
to `none` in v0.12, then you can't update the AK/SK without specifying
`--compresss` for existing volumes.

This PR allows user to update AK/SK without specify `--compress` for
existing volumes.

### DIFF
--- a/cmd/format.go
+++ b/cmd/format.go
@@ -210,6 +210,11 @@ func format(c *cli.Context) error {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


chnliyong (approved):

chnliyong:
Seems we couldn't distinguish `--compress none` and without `--compress`

PR #858, opened 2021-09-25
Align struct fields to make structs smaller

suzaku:


pkg/meta/types.go
```diff
@@ -68,10 +68,10 @@ type DumpedAttr struct {
 
 type DumpedSlice struct {
 	Pos     uint32 `json:"pos"`
```
davies, line 70: How about move pos to the end? We'd like to have chunkid and Size together for readability.

suzaku: Yes, we can place it to be between Len and Chunkid.

suzaku: Or we can just swap these two lines so that they stay together.

davies (approved):

PR #933, opened 2021-10-30
Fix the bug that file credential authentication fails when using GCS

### LINKED ISSUE #934: Insufficient permissions to use juice-csi-driver to access gcs in GKE
Insufficient permissions to use juice-csi-driver to access gcs in GKE.
The same situation occurs when using file credentials to access GCS in environments other than Google Cloud.

zhijian-pro:
Current strategy:
juicefs first looks up  `GOOGLE_APPLICATION_CREDENTIALS` environment variables.
+ If it is found, use its value as the local path of the file credential, and the credential file will be brought with it when accessing gcs. At this time, the permissions of juicefs are the permissions of the service-account that generated the credential file.
+ If the environment variable is not found, juicefs will think that it is in the Google Cloud environment, and its permissions are based on the permissions of the service account of the machine where it is located.
so，If juice runs in all environments except Google Cloud, you must set the `GOOGLE_APPLICATION_CREDENTIALS` environment variable, unless you have placed it in the default location: `$HOME/.config/gcloud/application_default_credentials.json`
BTW, there is no need to set `GOOGLE_APPLICATION_CREDENTIALS` environment variables when using jucefs in GKE.

Closes #934


pkg/object/gs.go
```diff
@@ -75,86 +75,77 @@ func (g *gs) Create() error {
...
-	return resp.Body, nil
+	return reader, nil
 }
 
 func (g *gs) Put(key string, data io.Reader) error {
-	obj := &storage.Object{Name: key}
-	_, err := g.service.Objects.Insert(g.bucket, obj).Media(data).Do()
-	return err
+	writer := g.client.Bucket(g.bucket).Object(key).NewWriter(ctx)
+	defer writer.Close()
```
davies, line 113: return the `error` here

pkg/object/gs.go
```diff
@@ -170,15 +161,12 @@ func newGS(endpoint, accessKey, secretKey string) (ObjectStorage, error) {
...
-	client, err := google.DefaultClient(ctx)
-	if err != nil {
-		log.Fatalf("Failed to create client: %v", err)
-	}
-	service, err := storage.NewService(ctx, option.WithHTTPClient(client))
+
+	client, err := storage.NewClient(ctx)
 	if err != nil {
-		log.Fatalf("Failed to create service: %v", err)
+		logger.Fatalf("Failed to create client: %v", err)
```
davies, line 167: return the error

davies (changes requested):
return error rather than fatal

davies (approved):

davies (changes requested):
we can't ignore the error from Close

zhijian-pro:
> we can't ignore the error from Close

I got it wrong just now, I will correct it immediately

davies (approved):

PR #935, opened 2021-10-31
cmd/info: add '--recursive' flag to get summary of a tree

SandyXSD:


cmd/info.go
```diff
@@ -79,10 +88,11 @@ func info(ctx *cli.Context) error {
 			continue
 		}
 
-		wb := utils.NewBuffer(8 + 8)
+		wb := utils.NewBuffer(8 + 9)
```
davies, line 91: For mountpoint using older version, what will happen?

pkg/vfs/internal.go
```diff
@@ -155,9 +155,10 @@ func handleInternalMsg(ctx Context, cmd uint32, r *utils.Buffer) []byte {
 	case meta.Info:
 		var summary meta.Summary
 		inode := Ino(r.Get64())
+		recursive := r.Get8()
```
davies, line 158: IF the client is older version, there will be no extra byte .

cmd/info.go
SandyXSD: The last byte will be ignored, as older clients will always return the recursive result

pkg/vfs/internal.go
SandyXSD: 459c3bc

PR #936, opened 2021-11-02
添加天翼云对象存储融合版、原生版支持

daybug:
天翼云对象存储各版本：https://www.ctyun.cn/h5/help2/10000101/10012468


davies:
@daybug  It seems that xstore and zos are compatible with S3, Can you use the s3 backend to work with them?  Right now, there are many duplicated code for those S3-compatible storage, we'd like to avoid them in future.

daybug:
> @daybug It seems that xstore and zos are compatible with S3, Can you use the s3 backend to work with them? Right now, there are many duplicated code for those S3-compatible storage, we'd like to avoid them in future.

OK. Understood.

PR #943, opened 2021-11-04
cache: force sync upload if local disk is too full for writeback

### LINKED ISSUE #942: Stop writeback when the disk is almost full
Currently, we don't limit the capacity used for writeback, which may fill all the free space, and other application may fail.

We should stop the writeback when the disk is almost full (the free ration is below half of specified `free ratio`), because writeback should have higher priority than read cache.

SandyXSD:
close #942

### DIFF
--- a/pkg/chunk/disk_cache.go
+++ b/pkg/chunk/disk_cache.go
@@ -57,6 +57,7 @@ type cacheStore struct {
 	pending   chan pendingFile
 	pages     map[string]*Page
 
+	pauseWB  int32 // atomic bool
 	used     int64
 	keys     map[string]cacheItem
 	scanned  bool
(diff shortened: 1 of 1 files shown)


SandyXSD:
An alternative is to add a flag `pauseWriteBack` and adjust it in `checkFreeSpace`, then we don't have to call `statfs` in every `stage`. What do you think? @davies

davies (changes requested):
don't call statfs() for every object.

pkg/chunk/disk_cache.go
```diff
@@ -112,7 +113,12 @@ func (cache *cacheStore) checkFreeSpace() {
 			cache.cleanup()
 			cache.Unlock()
 
-			br, fr := cache.curFreeRatio()
+			br, fr = cache.curFreeRatio()
+			if br < cache.freeRatio/2 || fr < cache.freeRatio/2 {
+				atomic.StoreInt32(&cache.pauseWB, 1)
```
davies, line 118: The address of cache.pauseWB should be aligned.

Since it's OK to see stale value of `pauseWB`, we don't need use atomic here.

davies (changes requested):
atomic is not required

davies (approved):

PR #953, opened 2021-11-07
fix redis-sentinel addresses parsing

nsampre:
with meta url being: `redis://mymaster,10.XX.XX.197,10.XX.XX.193,10.XX.XX.67:26379`

from: 

```
 SentinelAddrs: ([]string) (len=3 cap=3) {
  (string) (len=6) ":26379",
  (string) (len=6) ":26379",
  (string) (len=18) "10.XX.XX.67:26379"
 },
```

to

```
 SentinelAddrs: ([]string) (len=3 cap=3) {
  (string) (len=20) "10.XX.XX.197:26379",
  (string) (len=20) "10.XX.XX.193:26379",
  (string) (len=18) "10.XX.XX.67:26379"
 },
```

Close https://github.com/juicedata/juicefs/pull/953

### DIFF
--- a/pkg/meta/redis.go
+++ b/pkg/meta/redis.go
@@ -117,7 +117,7 @@ func newRedisMeta(driver, addr string, conf *Config) (Meta, error) {
 	var rdb *redis.Client
 	if strings.Contains(opt.Addr, ",") {
 		var fopt redis.FailoverOptions
-		ps := strings.Split(opt.Addr, ",")
+		ps := strings.Split(addr, ",")
 		fopt.MasterName = ps[0]
 		fopt.SentinelAddrs = ps[1:]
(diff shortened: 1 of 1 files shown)


pkg/meta/redis.go
```diff
@@ -117,16 +117,18 @@ func newRedisMeta(driver, addr string, conf *Config) (Meta, error) {
 	var rdb *redis.Client
 	if strings.Contains(opt.Addr, ",") {
 		var fopt redis.FailoverOptions
-		ps := strings.Split(opt.Addr, ",")
+		ps := strings.Split(addr, ",")
```
davies, line 120: the `addr` may contain db (host:port/db), so we should use `opt.Addr` here

nsampre: Fixed

PR #958, opened 2021-11-08
fsck: list broken files

### LINKED ISSUE #949: Find out broken files in `fsck`
When doing `juicefs fsck`, we can scan out lost objects, but users probably want to know the broken files (paths) containing these objects.

SandyXSD:
close #949

### DIFF
--- a/cmd/fsck.go
+++ b/cmd/fsck.go
@@ -148,10 +148,11 @@ func fsck(ctx *cli.Context) error {
 				if _, ok := blocks[key]; !ok {
 					if _, err := blob.Head(key); err != nil {
 						if _, ok := brokens[inode]; !ok {
-							if p, st := m.GetPath(meta.Background, inode); st == 0 {
+							if p, st := meta.GetPath(m, meta.Background, inode); st == 0 {
 								brokens[inode] = p
 							} else {
 								logger.Warnf("getpath of inode %d: %s", inode, st)
+								brokens[inode] = st.Error()
 							}
 						}
 						logger.Errorf("can't find block %s for file %s: %s", key, brokens[inode], err)
@@ -169,9 +170,10 @@ func fsck(ctx *cli.Context) error {
 	logger.Infof("Used by %d slices (%d bytes)", totalSlices, totalBytes)
 	if lost.Current() > 0 {
 		msg := fmt.Sprintf("%d objects are lost (%d bytes), %d broken files:\n", lost.Current(), lostBytes.Current(), len(brokens))
+		msg += fmt.Sprintf("%13s: PATH\n", "INODE")
 		var fileList []string
 		for i, p := range brokens {
-			fileList = append(fileList, fmt.Sprintf("%10d %s", i, p))
+			fileList = append(fileList, fmt.Sprintf("%13d: %s", i, p))
 		}
 		sort.Strings(fileList)
 		msg += strings.Join(fileList, "\n")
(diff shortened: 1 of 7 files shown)


SandyXSD:
TODO: The GetPath may not work if the first link of a file is removed (attr.Parent is not properly maintained)

cmd/fsck.go
```diff
@@ -127,24 +132,32 @@ func fsck(ctx *cli.Context) error {
...
+					sz = int(s.Size) - int(i)*chunkConf.BlockSize
+				}
+				key := fmt.Sprintf("%d_%d_%d", s.Chunkid, i, sz)
+				if _, ok := blocks[key]; !ok {
+					if _, err := blob.Head(key); err != nil {
+						if _, ok := brokens[inode]; !ok {
+							if p, st := m.GetPath(meta.Background, inode); st == 0 {
+								brokens[inode] = p
+							} else {
+								logger.Warnf("getpath of inode %d: %s", inode, st)
```
davies, line 154: use `inode:%d` as the path, otherwise it will be empty string

pkg/meta/tkv.go
```diff
@@ -2357,6 +2357,39 @@ func (r *kvMeta) ListSlices(ctx Context, slices *[]Slice, delete bool, showProgr
...
+		}
+		for key := range vals {
+			names = append(names, key[len(m.entryKey(inode, "")):])
+			break
+		}
+		inode = attr.Parent
+	}
+	names = append(names, "") // add root
+
+	for i, j := 0, len(names)-1; i < j; i, j = i+1, j-1 { // reverse
```
davies, line 2387: use a function for this

davies:
```
return "/" + strings.Join(reverse(names), "/"), 0
```

pkg/meta/interface.go
```diff
@@ -307,7 +307,9 @@ type Meta interface {
 	// Compact all the chunks by merge small slices together
 	CompactAll(ctx Context) syscall.Errno
 	// ListSlices returns all slices used by all files.
-	ListSlices(ctx Context, slices *[]Slice, delete bool, showProgress func()) syscall.Errno
+	ListSlices(ctx Context, slices map[Ino][]Slice, delete bool, showProgress func()) syscall.Errno
+	// GetPath returns full path of an inode; a random one is picked if it has multiple hard links
+	GetPath(ctx Context, inode Ino) (string, syscall.Errno)
```
davies, line 312: nit: GetPath could be implemented use  `Lookup(ctx, inode, "..")` and `Readdir(ctx, inode, 0)` without introducing a new API.

davies (changes requested):
This could be simplified.

pkg/meta/interface.go
davies: But `Lookup(ctx, inode, "..")` does not work on files ...

cmd/fsck.go
SandyXSD: It prints `$inode: $path` for each file, so maybe use the error message to replace path would be more comprehensive. Like:
```
juicefs[24127] <FATAL>: 3 objects are lost (24 bytes), 3 broken files:
        INODE: PATH
          325: /d3/f33
        32181: /fff
        32182: no such file or directory
```

pkg/meta/interface.go
SandyXSD: Just use `GetAttr` for now

davies (approved):

PR #959, opened 2021-11-08
meta/dump: fix incorrect field values of DumpedSlice

SandyXSD:


davies:
@SandyXSD Can we update the unit test to cover these cases?

SandyXSD:
> @SandyXSD Can we update the unit test to cover these cases?

16393eb : just compare the dumped result with the original sample file

PR #961, opened 2021-11-09
get more coverage info with integration test

sanwan:


fstests/Makefile
```diff
@@ -8,7 +8,7 @@ xattrs:
 	getfattr -n user.k /jfs/test_xattrs | grep -q user.k=
 
 fsracer: healthcheck secfs.test/tools/bin/fsracer
-	secfs.test/tools/bin/fsracer $(DURATION) /jfs >fsracer.log
```
davies, line 11: It's noisy, it's better to keep it.

integration/integration_test.go
```diff
@@ -0,0 +1,636 @@
+package integration
```
davies, line 1: license header

PR #975, opened 2021-11-12
mount: add `--backup-meta` to periodically backup metadata to object storage

### LINKED ISSUE #957: Backup metadata to object store automatically
Right metadata is critical to data safety. If the metadata is lost, we can't get the data back, because we don't know how to assemble the objects in object store back.

People may forget to setup the backup for Redis/MySQL/TiKV, or the backup is broken, we'd like to have another layer to cover the worst case. 

The client (`juicefs mount`) can dump the metadata hourly, and upload them into object store, and keep them based on some rotation policy.  In the worse case, we can download these backup and use it to recover most of the data.

SandyXSD:
close #957

### DIFF
--- a/pkg/vfs/backup.go
+++ b/pkg/vfs/backup.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/juicedata/juicefs/pkg/meta"
 	"github.com/juicedata/juicefs/pkg/object"
+	osync "github.com/juicedata/juicefs/pkg/sync"
 )
 
 // Backup metadata periodically in the object storage
@@ -90,7 +91,8 @@ func backup(blob object.ObjectStorage, name string) error {
 }
 
 func cleanupBackups(blob object.ObjectStorage) {
-	objChan, err := blob.ListAll("meta/", "")
+	blob = object.WithPrefix(blob, "meta/")
+	objChan, err := osync.ListAll(blob, "", "")
 	if err != nil {
 		logger.Warnf("listAll prefix meta/: %s", err)
 		return
@@ -122,11 +124,11 @@ func cleanupBackups(blob object.ObjectStorage) {
 	}
 
 	for i := len(objs) - 1; i >= 0; i-- {
-		if len(objs[i]) != 35 { // len("meta/dump-2006-01-02-150405.json.gz")
+		if len(objs[i]) != 30 { // len("dump-2006-01-02-150405.json.gz")
 			logger.Warnf("bad object for metadata backup: %s", objs[i])
 			continue
 		}
-		ts, err := time.Parse("2006-01-02-150405", objs[i][10:27])
+		ts, err := time.Parse("2006-01-02-150405", objs[i][5:22])
 		if err != nil {
 			logger.Warnf("bad object for metadata backup: %s", objs[i])
 			continue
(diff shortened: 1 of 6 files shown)


pkg/vfs/backup.go
```diff
@@ -0,0 +1,145 @@
...
+		time.Sleep(interval / 10)
+		var value []byte
+		if st := m.GetXattr(ctx, 1, key, &value); st != 0 && st != meta.ENOATTR {
+			logger.Warnf("getxattr inode 1 key %s: %s", key, st)
+			continue
+		}
+		var last time.Time
+		var err error
+		if len(value) == 17 { // len("2006-01-02-150405")
+			last, err = time.Parse("2006-01-02-150405", string(value))
```
davies, line 43: we can use timestamp here

pkg/vfs/backup.go
```diff
@@ -0,0 +1,145 @@
...
+		logger.Warnf("listAll prefix meta/: %s", err)
+		return
+	}
+	var objs []string
+	for o := range objChan {
+		objs = append(objs, o.Key())
+	}
+	sort.Strings(objs)
+
+	// Cleanup policy:
```
davies, line 105: Add test cases for these

pkg/vfs/backup.go
```diff
@@ -0,0 +1,145 @@
...
+				logger.Infof("backup metadata skipped because of too many inodes: %d %s", iused, interval)
+				continue
+			}
+			val := now.Format("2006-01-02-150405")
+			if st := m.SetXattr(ctx, 1, key, []byte(val), meta.XattrCreateOrReplace); st != 0 {
+				logger.Warnf("setxattr inode 1 key %s: %s", key, st)
+				continue
+			}
+			go cleanupBackups(blob)
+			logger.Infof("backup metadata started")
```
davies, line 62: debug level

pkg/vfs/backup.go
```diff
@@ -0,0 +1,145 @@
...
+			if err = backup(blob, val); err == nil {
+				logger.Info("backup metadata succeed")
```
davies, line 64: Added the time used by backup.

davies (changes requested):
Unit tests

pkg/vfs/backup.go
```diff
@@ -39,29 +39,28 @@ func Backup(blob object.ObjectStorage, interval time.Duration) {
...
 		if err != nil {
 			logger.Warnf("parse time value %s: %s", value, err)
 			continue
 		}
-		if now := time.Now().UTC(); now.Sub(last) >= interval {
+		if now := time.Now(); now.Sub(last) >= interval {
 			var iused, dummy uint64
 			_ = m.StatFS(ctx, &dummy, &dummy, &iused, &dummy)
-			if iused/(5<<20) > uint64(interval/time.Hour) {
+			if iused/5e6 > uint64(interval/time.Hour) {
```
davies, line 52: when interval is less than 1 hour, it will not backup anything

pkg/vfs/backup.go
SandyXSD: 7e44334

davies (approved):

PR #993, opened 2021-11-16
java coverage

sanwan:


sdk/java/pom.xml
```diff
@@ -9,14 +9,15 @@
...
+		<argLine>-Djdk.net.URLClassPath.disableClassPathURLCheck=true -Djava.library.path=${project.basedir}/../mount/libjfs:${java.library.path}</argLine>
 	</properties>
 	<build>
 		<plugins>
 			<plugin>
 				<artifactId>maven-surefire-plugin</artifactId>
 				<version>2.19.1</version>
 				<configuration>
-					<argLine>-Djdk.net.URLClassPath.disableClassPathURLCheck=true -Djava.library.path=${project.basedir}/../mount/libjfs:${java.library.path}</argLine>
+					<argLine>@{argLine}</argLine>
```
davies, line 20: ?

davies:
@sanwan I can't see which part of the source code is covered by existing tests, how?

sdk/java/pom.xml
sanwan: If surefire and jacoco are used in same pom file , argline paramters of jacoco will be overwritted .More details may refer to https://stackoverflow.com/questions/18107375/getting-skipping-jacoco-execution-due-to-missing-execution-data-file-upon-exec.

davies: I see, it's better to leave a comment here.

davies:
I can see the coverage of some of the Java files, the most important one (JuiceFileSystemImpl.java) is loaded in another classloader, how can we measure the coverage of it?

sdk/java/pom.xml
sanwan: ok

PR #996, opened 2021-11-16
speed up metadata dumping and loading for Tikv 

zhijian-pro:
speed up metadata dumping and loading for Tikv

### DIFF
--- a/pkg/meta/tkv.go
+++ b/pkg/meta/tkv.go
@@ -16,6 +16,7 @@
 package meta
 
 import (
+	"bufio"
 	"bytes"
 	"encoding/binary"
 	"encoding/json"
@@ -28,6 +29,8 @@ import (
 	"sync/atomic"
 	"syscall"
 	"time"
+
+	"github.com/pkg/errors"
 
 	"github.com/juicedata/juicefs/pkg/utils"
 )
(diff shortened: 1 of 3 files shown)


pkg/meta/tkv.go
```diff
@@ -2694,10 +2877,41 @@ func (m *kvMeta) LoadMeta(r io.Reader) error {
 		NextSession: 1,
 	}
 	refs := make(map[string]int64)
+
+	//TODO: try raw kv   cli.BatchPut()
```
davies, line 2881: rawkv is not compatible to transactional kv, we can't use that here.

pkg/meta/tkv.go
```diff
@@ -2598,7 +2756,32 @@ func (m *kvMeta) DumpMeta(w io.Writer) error {
...
+	if _, err = bw.WriteString("\n}\n"); err != nil {
+		return err
+	}
+	if bar.Current() != total {
+		logger.Warnf("Dumped %d / total %d, some entries are not dumped", bar.Current(), total)
+	}
+	bar.SetTotal(0, true)
+	progress.Wait()
+
+	return bw.Flush()
```
davies, line 2784: We should not fail the dump even bw.Flush() failed

zhijian-pro: Doesn't flush failure mean that the result is incomplete?

davies: my bad, bw is the writer, never mind.

davies:
in favor #998

PR #1006, opened 2021-11-17
Speed up dump for sql engine

zhijian-pro:
Speed up for dbmeta

### DIFF
--- a/pkg/meta/sql.go
+++ b/pkg/meta/sql.go
@@ -2599,7 +2599,8 @@ func (m *dbMeta) dumpEntry(inode Ino) (*DumpedEntry, error) {
 			return err
 		}
 		if !ok {
-			return fmt.Errorf("inode %d not found", inode)
+			logger.Warnf("The entry of the inode was not found. inode: %v", inode)
+			return nil
 		}
 		attr := &Attr{}
 		m.parseAttr(n, attr)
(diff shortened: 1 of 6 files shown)


pkg/meta/sql.go
```diff
@@ -2637,35 +2647,170 @@ func (m *dbMeta) dumpEntry(inode Ino) (*DumpedEntry, error) {
...
+	for idx, e := range edges {
+		var entry *DumpedEntry
+		if m.root == 1 {
+			entry, err = m.dumpEntryFast(e.Inode)
+		} else {
+			entry, err = m.dumpEntry(e.Inode)
+		}
 		if err != nil {
-			return nil, err
+			return err
```
davies, line 2730: Should we skip the edge and continue (log it as warning)?

zhijian-pro: It has been modified

davies (approved):

zhijian-pro:
## Performance Testing
### data and env: 
1. A total of 100w empty files are evenly distributed in 200 local folders
2. local mysql

### result:
#### old version
+ mysql dump: 3600s
+ mysql load: 46m

#### new version
+ mysql  dump :25s
+ mysql load: 500s

## Test conclusion   
+ The performance of the dump operation is **improved by 144 times** 
+ The performance of the load operation is **improved by 5 times**

PR #1011, opened 2021-11-18
improve test coverage for pkg/vfs

davies:


pkg/vfs/vfs.go
```diff
@@ -587,6 +563,7 @@ func (v *VFS) Write(ctx Context, ino Ino, buf []byte, off, fh uint64) (err sysca
 		size := int(rb.Get32())
 		if rb.Left() < int(size) {
 			logger.Debugf("message not complete: %d %d > %d", cmd, size, rb.Left())
+			err = syscall.EINVAL
```
SandyXSD, line 566: should be no error for this

PR #1021, opened 2021-11-19
Fix docs image directories structure to reflect i18n

chnliyong:
Move the images in `docs/images` to corresponding locale image directories(`en/images` and `zh_cn/images`)


chnliyong:
Close as there's a PR #1020

PR #1030, opened 2021-11-22
fix log no line break

zhijian-pro:
fix log no line break


zhijian-pro:
No reason found yet,resolve later when it reappears during use.

PR #1035, opened 2021-11-22
metadata: should not skip error if open a file not exist

chiyutianyi:
The original implements will always return ok if open a file that does
not exist.

Another unresolved problem is that if an "attr" is given with "Full=true", even
if they do not match, the "GetAttr()" call will still be skipped. But
this does not seem to be a serious problem.

Signed-off-by: 炽天 <hanxin.hx@alibaba-inc.com>

### DIFF
--- a/pkg/meta/redis.go
+++ b/pkg/meta/redis.go
@@ -1934,6 +1934,10 @@ func (r *redisMeta) Open(ctx Context, inode Ino, flags uint32, attr *Attr) sysca
 	if r.conf.ReadOnly && flags&(syscall.O_WRONLY|syscall.O_RDWR|syscall.O_TRUNC|syscall.O_APPEND) != 0 {
 		return syscall.EROFS
 	}
+	if attr != nil {
+		// reset attr.Full to avoid mismatch between inode and attr
+		attr.Full = false
+	}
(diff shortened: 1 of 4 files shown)


chiyutianyi:
By the way, I found that there are many duplicate implementations in redis.go, sql.go, tkv.go, such as Access(), Open(), Mkdir(), etc., should there be a default implementation?

davies (approved):

davies:
@chiyutianyi It's better to reset attr.Full to false before OpenCheck(), please add that also?

davies:
> By the way, I found that there are many duplicate implementations in redis.go, sql.go, tkv.go, such as Access(), Open(), Mkdir(), etc., should there be a default implementation?

Yes, we will look into that.

chiyutianyi:
> @chiyutianyi It's better to reset attr.Full to false before OpenCheck(), please add that also?

OK, I will do that. 

However, the reason why I did not fix the problem before is that, the current code does not cause actual problems in " vfs.go". It is just that during the execution of the test cases, it may produce inaccurate situations. 

In addition to the inner openfiles cache, I also like an external cache wrapper to make "GetAttr()" interact with "Open()". If we do this, resetting attr.Full may not be the best choice.

PR #1040, opened 2021-11-22
meta: reduce duplicated code

davies:


pkg/meta/redis_test.go
```diff
@@ -91,7 +91,7 @@ func testMetaClient(t *testing.T, m Meta) {
 	}
 	switch r := m.(type) {
 	case *redisMeta:
-		if r.sid != int64(ses[0].Sid) {
+		if r.sid != uint64(ses[0].Sid) {
```
SandyXSD, line 94: Sid is uint64, no need to convert

pkg/meta/load_dump_test.go
```diff
@@ -113,6 +117,10 @@ func TestLoadDump(t *testing.T) {
 		if err = m.DumpMeta(fp); err != nil {
 			t.Fatalf("dump meta: %s", err)
 		}
+		cmd := exec.Command("diff", "redis.dump", "sqlite3.dump")
```
SandyXSD, line 120: compare sqlite3.dump with sampleFile directly might be better since it's a standalone test now

pkg/meta/redis.go
```diff
@@ -1840,7 +1632,7 @@ func (r *redisMeta) cleanStaleSession(sid int64, sync bool) {
 		logger.Warnf("SMembers %s: %s", key, err)
 		return
 	}
-	ssid := strconv.FormatInt(sid, 10)
+	ssid := strconv.FormatInt(int64(sid), 10)
```
chiyutianyi, line 1635: Why don't you use "strconv.FormatUint" here ?

pkg/meta/redis.go
```diff
@@ -2235,7 +1984,7 @@ func (r *redisMeta) cleanupSlices() {
 							chunkid, _ := strconv.Atoi(ps[0][1:])
```
chiyutianyi, line 1984: Although "chunkid" has no change here, it is better to use "strconv.ParseUint(ps[0][1:], 10, 64)" instead of strconv.Atoi(ps[0][1:]) here.

pkg/meta/sql.go
```diff
@@ -750,66 +577,22 @@ func (m *dbMeta) Lookup(ctx Context, parent Ino, name string, inode *Ino, attr *
...
-		m.of.Update(inode, attr)
-	} else if inode == 1 {
-		err = nil
-		attr.Typ = TypeDirectory
-		attr.Mode = 0777
-		attr.Nlink = 2
-		attr.Length = 4 << 10
-	} else if err == nil {
-		err = syscall.ENOENT
+	if err == nil {
```
chiyutianyi, line 590: I think it will be better like: 
```
	if err != nil {
		return errno(err)
	}
	if ok {
		m.parseAttr(&n, attr)
		return syscall.F_OK
	}
	return syscall.ENOENT
```

pkg/meta/base.go
```diff
@@ -0,0 +1,473 @@
...
+func (m *baseMeta) InvalidateChunkCache(ctx Context, inode Ino, indx uint32) syscall.Errno {
+	m.of.InvalidateChunk(inode, indx)
+	return 0
+}
+
+func (m *baseMeta) NewChunk(ctx Context, inode Ino, indx uint32, offset uint32, chunkid *uint64) syscall.Errno {
+	m.freeMu.Lock()
+	defer m.freeMu.Unlock()
+	if m.freeChunks.next >= m.freeChunks.maxid {
+		v, err := m.incrCounter("nextChunk", 1000)
```
chiyutianyi, line 424: When you are doing refactoring, why not change these magic values to constants?

chiyutianyi (commented):
I am glad to see this patch so soon, it will make it easier for readers to see the differences of each database, and will be more conducive to add a new metadata storage implementation.

pkg/meta/tkv.go
```diff
@@ -1229,32 +969,8 @@ func (m *kvMeta) Fallocate(ctx Context, inode Ino, mode uint8, off uint64, size
...
-
-func (m *kvMeta) Symlink(ctx Context, parent Ino, name string, path string, inode *Ino, attr *Attr) syscall.Errno {
-	defer timeit(time.Now())
-	return m.mknod(ctx, parent, name, TypeSymlink, 0644, 022, 0, path, inode, attr)
-}
-
-func (m *kvMeta) Mknod(ctx Context, parent Ino, name string, _type uint8, mode, cumask uint16, rdev uint32, inode *Ino, attr *Attr) syscall.Errno {
-	defer timeit(time.Now())
-	return m.mknod(ctx, parent, name, _type, mode, cumask, rdev, "", inode, attr)
+func (m *kvMeta) readlink(inode Ino) ([]byte, error) {
```
SandyXSD, line 972: Return ENOENT if value is empty, same as in sql.go.
Then in base maybe we don't need to check `target == nil`

pkg/meta/base.go
```diff
@@ -0,0 +1,473 @@
...
+	}
+	return err
+}
+
+func (m *baseMeta) InvalidateChunkCache(ctx Context, inode Ino, indx uint32) syscall.Errno {
+	m.of.InvalidateChunk(inode, indx)
+	return 0
+}
+
+func (m *baseMeta) NewChunk(ctx Context, inode Ino, indx uint32, offset uint32, chunkid *uint64) syscall.Errno {
```
SandyXSD, line 420: Let's delete unused arguments this time

pkg/meta/base.go
```diff
@@ -0,0 +1,473 @@
...
+
+func (m *baseMeta) Open(ctx Context, inode Ino, flags uint32, attr *Attr) syscall.Errno {
+	if m.conf.ReadOnly && flags&(syscall.O_WRONLY|syscall.O_RDWR|syscall.O_TRUNC|syscall.O_APPEND) != 0 {
+		return syscall.EROFS
+	}
+	if m.conf.OpenCache > 0 && m.of.OpenCheck(inode, attr) {
+		return 0
+	}
+	var err syscall.Errno
+	if attr != nil && !attr.Full {
```
SandyXSD, line 406: Reset attr.Full = fale, or maybe just use `if attr != nil` here

PR #1047, opened 2021-11-23
pkg/meta: add more test for meta

zhijian-pro:
pkg/meta: add more test for meta


zhijian-pro:
@sanwan  Why does codecov not take effect ？

davies (changes requested):
we should update the test case for TKV as well.

davies (approved):

PR #1056, opened 2021-11-24
add namespace and label to existing metrics

davies:


SandyXSD:
Should modify https://github.com/juicedata/juicefs/blob/main/cmd/stats.go#L137-L138 as well

PR #1058, opened 2021-11-25
cmd: add more test for cmd

zhijian-pro:
cmd: add more test for cmd

### DIFF
--- a/cmd/dump_test.go
+++ b/cmd/dump_test.go
@@ -25,16 +25,19 @@ import (
 
 func TestDumpAndLoad(t *testing.T) {
 	metaUrl := "redis://127.0.0.1:6379/10"
+	opt, err := redis.ParseURL(metaUrl)
+	if err != nil {
+		t.Fatalf("ParseURL: %v", err)
+	}
+	rdb := redis.NewClient(opt)
+	rdb.FlushDB(context.Background())
+
 	t.Run("Test Load", func(t *testing.T) {
 		loadArgs := []string{"", "load", metaUrl, "./../pkg/meta/metadata.sample"}
-		opt, err := redis.ParseURL(metaUrl)
+		err = Main(loadArgs)
 		if err != nil {
-			t.Fatalf("ParseURL: %v", err)
+			t.Fatalf("load failed: %v", err)
 		}
-		rdb := redis.NewClient(opt)
-		rdb.FlushDB(context.Background())
-
-		Main(loadArgs)
 		if rdb.DBSize(context.Background()).Val() == 0 {
 			t.Fatalf("load error: %v", err)
 		}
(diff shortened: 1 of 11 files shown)


cmd/dump_test.go
```diff
@@ -0,0 +1,53 @@
...
+	})
+	t.Run("Test dump", func(t *testing.T) {
+		dumpArgs := []string{"", "dump", metaUrl, "/tmp/dump_test.json"}
+
+		Main(dumpArgs)
+		_, err := os.Stat("/tmp/dump_test.json")
+		if err != nil {
+			t.Fatalf("dump error: %v", err)
+		}
+	})
```
davies, line 51: Add a case for subdir

cmd/main.go
```diff
@@ -56,7 +56,7 @@ func globalFlags() []cli.Flag {
 	}
 }
 
-func main() {
+func Main(args []string) {
```
davies, line 59: return error

cmd/main.go
```diff
@@ -97,12 +97,16 @@ func main() {
 		}
 	}
 
-	err := app.Run(reorderOptions(app, os.Args))
+	err := app.Run(reorderOptions(app, args))
 	if err != nil {
 		log.Fatal(err)
```
davies, line 102: move the `log.Fatal(err)` into main()

cmd/info_test.go
```diff
@@ -0,0 +1,45 @@
...
+	for i := 0; i < 10; i++ {
+		filename := fmt.Sprintf("%s/dir1/f%d.txt", mountpoint, i)
+		err := ioutil.WriteFile(filename, []byte("test"), 0644)
+		if err != nil {
+			t.Fatalf("mount failed: %v", err)
+		}
+	}
+
+	infoArgs := []string{"", "info", metaUrl, fmt.Sprintf("%s/dir1", mountpoint)}
+	Main(infoArgs)
```
davies, line 43: check the output.

We replace os.Stdout to bytes writer before calling Main() to capture the output.

zhijian-pro: ok

cmd/umount_test.go
```diff
@@ -0,0 +1,53 @@
...
+		t.Fatalf("mount failed: %v", err)
+	}
+
+	err := UmountTmp(mountpoint)
+	if err != nil {
+		t.Fatalf("umount failed: %v", err)
+	}
+
+	t.Log("---------------")
+	println(prometheus.DefaultRegisterer)
```
davies, line 43: ?

davies: We should check that the /tmp/testDir is not a mountpoint

cmd/rmr_test.go
```diff
@@ -0,0 +1,62 @@
...
+		t.Fatalf("rmr failed : %v", err)
+	}
+
+	for _, path := range paths {
+		dir, err := os.ReadDir(mountpoint + path)
+		if len(dir) != 0 {
+			t.Fatalf("test rmr error: %v", err)
+		}
+	}
+	err := UmountTmp(mountpoint)
```
davies, line 58: call it with defer

cmd/info.go
```diff
@@ -49,6 +50,14 @@ func infoFlags() *cli.Command {
 	}
 }
 
+var writerInfo = func(content string, w io.Writer) error {
```
davies, line 53: We should not messy the source code just because of tests.

davies (approved):

PR #1064, opened 2021-11-26
Remove duplicate code

201341:


pkg/meta/redis.go
```diff
@@ -412,53 +412,6 @@ func (r *redisMeta) parseEntry(buf []byte) (uint8, Ino) {
 	return buf[0], Ino(binary.BigEndian.Uint64(buf[1:]))
 }
 
-func (r *redisMeta) parseAttr(buf []byte, attr *Attr) {
```
davies, line 415: move this into baseMeta, then you don't need update the caller.

201341: ok

PR #1065, opened 2021-11-27
Add unit tests for prefetch.go

suzaku:


pkg/chunk/prefetch_test.go
```diff
@@ -0,0 +1,49 @@
+package chunk
+
+import (
+	"testing"
+	"time"
+
+	"go.uber.org/atomic"
```
davies, line 7: Should we use the sync package instead?

suzaku: Fixed

PR #1066, opened 2021-11-28
add a temporary .trash directory to keep removed files for a certain time

### LINKED ISSUE #1057: Proposal: TrashBin
Data safety is always critical for a storage system. We should add TrashBin in JuiceFS and keep user-removed files for a certain period. Details below:

### How to Configure

Add a new option `--trash-days`  in `format` command to specify the days after which removed files will be permanently deleted.

- the default value is 1, which means files removed before 1 day will be automatically cleaned up
- use `--trash-days 0` to disable this feature; i.e, empty the trash now and make all removed files afterwards be cleaned up immediately
- run the `format` command again to change the configured value of days

### How to Use

The `.trash` directory should be located under root `/`, and is created automatically after volume formatted.

#### Remove

- Empty directory is removed as usual
- Non-directory file is moved to the `.trash`, under a sub-directory named by the date and hour that remove occurs
- File is rename as `{parent inode}-{file inode}-{original name}`, e.g, `1-10-test.txt`
- Data usage is not reduced until file is deleted from the `.trash`
- Move files directly into the `.trash` is not permitted

#### Browse

- All users can browse file list in `.trash`, and see files removed by others
- Removed file keeps its mode, so a user can still read it as long as he/she can read it previously, even if it's not remove by this user
- If mounted with `--subdir`, the whole `.trash` is still browsable, including files removed outside of the subdir

#### Restore/Delete

- Root user can restore a file by simply move it to another location outside of the `.trash`, without data copy
- Root user can permanently delete a file, and release its data storage in a short time
- Non-root users are not allowed to move or delete files in the `.trash`; however, they can still copy readable files to another location
  - Not decided yet, maybe should allow users to manipulate files owned by him/her?

#### Other

- All other operations that may change status of a file are not permitted, and should return `EPERM`

SandyXSD:
close #1057


SandyXSD:
TODO:
- [ ] Maybe create a real inode for trash directory, and use meta APIs to simplify code
  - a bit of performance regression
  - lookup/getattr may return different attribute from that in readdir
- [ ] Need a more persuasive file name for those moved into the trash
- [ ] Fix Lua script for Redis lookup/resolve
- [ ] Add unit tests

pkg/meta/base.go
```diff
@@ -526,27 +647,116 @@ func (m *baseMeta) deleteSlice(chunkid uint64, size uint32) {
...
+	now := time.Now()
+	var entries []*Entry
+	if st := m.en.doReaddir(ctx, TrashInode, 0, &entries); st != 0 {
+		logger.Warnf("readdir trash %d: %s", TrashInode, st)
+		return
+	}
+	sort.Slice(entries, func(i, j int) bool { return entries[i].Inode < entries[j].Inode })
+	var count int
+	defer func() {
+		logger.Infof("cleanup trash: deleted %d files in %v", count, time.Since(now))
```
davies, line 724: no logging if nothing was deleted

pkg/meta/base.go
```diff
@@ -526,27 +647,116 @@ func (m *baseMeta) deleteSlice(chunkid uint64, size uint32) {
...
 	}
-	*entries = []*Entry{
-		{
-			Inode: inode,
-			Name:  []byte("."),
-			Attr:  &Attr{Typ: TypeDirectory},
-		},
+	m.Unlock()
+
+	st := m.en.doLookup(Background, TrashInode, name, trash, nil)
```
davies, line 667: pass the context here so we can interrupt it

pkg/meta/base.go
```diff
@@ -526,27 +647,116 @@ func (m *baseMeta) deleteSlice(chunkid uint64, size uint32) {
...
+		}
+		if ts.Before(edge) {
+			var subEntries []*Entry
+			if st := m.en.doReaddir(ctx, e.Inode, 0, &subEntries); st != 0 {
+				logger.Warnf("readdir subTrash %d: %s", e.Inode, st)
+				continue
+			}
+			rmdir := true
+			for _, se := range subEntries {
+				if st := m.en.doUnlink(ctx, e.Inode, string(se.Name)); st == 0 {
```
davies, line 742: e could be a directory, should we call doRmdir() ?

pkg/meta/base.go
```diff
@@ -526,27 +647,116 @@ func (m *baseMeta) deleteSlice(chunkid uint64, size uint32) {
...
+	if st == syscall.ENOENT {
+		st = m.en.doMknod(Background, TrashInode, name, TypeDirectory, 0555, 0, 0, "", trash, nil)
+	}
+
+	m.Lock()
+	if st != 0 && st != syscall.EEXIST {
+		logger.Warnf("create subTrash %s: %s", name, st)
+		return st
+	}
+	m.subTrash.inode = *trash
```
davies, line 677: It's better to double check that *trash is valid

pkg/meta/base.go
```diff
@@ -526,27 +647,116 @@ func (m *baseMeta) deleteSlice(chunkid uint64, size uint32) {
 	}
 }
 
-func (m *baseMeta) Readdir(ctx Context, inode Ino, plus uint8, entries *[]*Entry) syscall.Errno {
-	inode = m.checkRoot(inode)
-	var attr Attr
-	if err := m.GetAttr(ctx, inode, &attr); err != 0 {
-		return err
+func (m *baseMeta) toTrash(parent Ino) bool {
+	return m.fmt.TrashDays > 0 && !isTrash(parent)
```
davies, line 651: Where do we check that only root can delete a trash?

pkg/meta/redis_test.go
```diff
@@ -1094,6 +1095,56 @@ func testCloseSession(t *testing.T, m Meta) {
...
+	if len(entries) != 2 {
+		t.Fatalf("entries: %d", len(entries))
+	}
+	entries = entries[:0]
+	if st := m.Readdir(ctx, TrashInode+1, 0, &entries); st != 0 {
+		t.Fatalf("readdir: %s", st)
+	}
+	if len(entries) != 6 {
+		t.Fatalf("entries: %d", len(entries))
+	}
```
davies, line 1145: Add a few case for cleaning trash files.

davies: can we call doCleanupTrash(force=True) to empty the trash?

pkg/meta/base.go
SandyXSD: Only for Lookup? The reason here is that we want the auto-created subTrash directory to be owned by root.

davies: Both of them. If we can't interrupt them, then we don't need to release the lock m.

pkg/meta/base.go
SandyXSD: Trash and SubTrash directory are created with uid/gid = 0 and mode = 0555

SandyXSD: 8f2f1b2

PR #1067, opened 2021-11-29
cmd: add test for cmd

zhijian-pro:
cmd: add test for cmd


zhijian-pro:
The  command  of  `bench , profile ,  stat, status,  gateway`  No unit tests have been added.
Now  the test coverage  of cmd  package  is  44.18%

PR #1089, opened 2021-12-02
meta: force backup to dump the whole metadata

SandyXSD:


pkg/meta/tkv.go
```diff
@@ -2238,7 +2240,7 @@ func (m *kvMeta) DumpMeta(w io.Writer) (err error) {
 		bar.SetTotal(total, false)
 		bar.IncrInt64(currentIncr)
 	}
-	if err = m.dumpDir(m.root, tree, bw, 1, showProgress); err != nil {
+	if err = m.dumpDir(root, tree, bw, 1, root == 1, showProgress); err != nil {
```
davies, line 2243: we can check m.snap != nil instead, do we still need to pass `useSnap`?

PR #1091, opened 2021-12-02
add elasticsearch to github action

sanwan:


sanwan:
It costs about 80 minutes for elasticsearch test.

PR #1095, opened 2021-12-03
fix lua script handling big inode number

SandyXSD:


SandyXSD:
- Lua `redis.call('Get', xxx)` a non-existed key returns Go nil, not an empty string
- Lua `error(xxx)` method returns an error message containing "Error running script"

PR #1108, opened 2021-12-08
Add integration test for S3 gateway

zhijian-pro:
1. support multi-bucket mode of s3 gateway
2. support list object  use --recursive
3. add s3gateway test

### DIFF
--- a/integration/s3gateway_test.sh
+++ b/integration/s3gateway_test.sh
@@ -488,16 +488,6 @@ function test_multipart_upload_0byte() {
     fi
 
     if [ $rv -eq 0 ]; then
-        ret_etag=$(echo "$out" | jq -r .ETag | sed -e 's/^"//' -e 's/"$//')
-        # match etag
-        if [ "$etag" != "$ret_etag" ]; then
-            rv=1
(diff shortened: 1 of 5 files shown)


pkg/gateway/gateway.go
```diff
@@ -171,18 +172,23 @@ func (n *jfsObjects) DeleteBucket(ctx context.Context, bucket string, forceDelet
 	if !n.isValidBucketName(bucket) {
 		return minio.BucketNameInvalid{Bucket: bucket}
 	}
-	if n.conf.Format.Name != "" {
+	if !n.multiBucket {
 		return minio.BucketNotEmpty{Bucket: bucket}
 	}
-	eno := n.fs.Delete(mctx, n.path(bucket))
+	var eno syscall.Errno
+	if forceDelete {
```
davies, line 179: We should call `Rmr` for forceDelete

integration/s3gateway_test.sh
```diff
@@ -0,0 +1,1848 @@
+#!/bin/bash
```
davies, line 1: License header

zhijian-pro:
This test ignores errors related to `etag` and `storageClass`

Issue #1129, opened 2021-12-15
gateway create or delete dir wrong  


sjp00556:
https://github.com/juicedata/juicefs/blob/e85b22f4d12c7f044a06a66f348066da0d1372f5/pkg/gateway/gateway.go#L155

this should be `sep + path.Join(p...)`, we don't need retain trailing SlashSeparator of the last element

`curl -X PUT "endpoint/bucket/path/"`  expect to create dir path, but now it will create path/path

davies:
@sjp00556 Is there a document for this behavior?

sjp00556:
when object is end with / , n.path shouldn't retain the last /. we treat dir and file samely in fs.FileSystem, dir no need end with /.   i have already tested put/get/head/copy/deletet/multipart api.  it went well

i fix it here: https://github.com/juicedata/juicefs/pull/1130

sjp00556:
here is delete example to understand this bug:
`curl -X DELETE sjpdev:8889/starfstest/path/`  we expect to delete` /path`,  but it will delete `/path/path `

![image](https://user-images.githubusercontent.com/6017354/146168905-f93e8765-4b84-4e8a-ba7a-a3842c61816d.png)

davies:
@sjp00556 We sent #1149 to fix the bug in fs.Mkdir() and fs.Delete(), thanks for reporting it.

For Put with trailing slash, JuiceFS can't be compatible with S3, so we follow the behavior in MinIO: 1) create the directory when the body is empty 2) fail the request when body is not empty.

PR #1130, opened 2021-12-15
bugfix: fix gateway path when object endwith SlashSeparator

sjp00556:
link to: https://github.com/juicedata/juicefs/issues/1129

### DIFF
--- a/pkg/gateway/gateway.go
+++ b/pkg/gateway/gateway.go
@@ -152,7 +152,7 @@ func (n *jfsObjects) path(p ...string) string {
 	if len(p) > 0 && p[0] == n.conf.Format.Name {
 		p = p[1:]
 	}
-	return sep + minio.PathJoin(p...)
+	return sep + path.Join(p...)
 }
 
 func (n *jfsObjects) tpath(p ...string) string {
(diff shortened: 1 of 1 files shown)


SandyXSD:
@sjp00556 Thank you for contribution~ Please check and sign the CLA first.

pkg/gateway/gateway.go
```diff
@@ -520,7 +520,7 @@ func (n *jfsObjects) GetObjectInfo(ctx context.Context, bucket, object string, o
 		return
 	}
 	if strings.HasSuffix(object, sep) && !fi.IsDir() {
-		err = jfsToObjectErr(ctx, os.ErrNotExist, bucket, object)
```
davies, line 523: Will this change any behavior?

pkg/gateway/gateway.go
```diff
@@ -152,7 +152,7 @@ func (n *jfsObjects) path(p ...string) string {
 	if len(p) > 0 && p[0] == n.conf.Format.Name {
 		p = p[1:]
 	}
-	return sep + minio.PathJoin(p...)
```
davies, line 155: After we fixed the bug in fs.Mkdir(), is this still needed?

pkg/gateway/gateway.go
sjp00556: os.ErrNotExist will give 500 http code InternalError, syscall.ENOENT give 404 code. i think 404 is better

pkg/gateway/gateway.go
sjp00556: this change can fix path problem like this: "/path/b////" => "/path/b", but i think it doesn't metter

pkg/gateway/gateway.go
davies: I see, can you send another PR for this one?

sjp00556: ok, i sent a new PR: https://github.com/juicedata/juicefs/pull/1158

PR #1133, opened 2021-12-15
Keep SGID for directory

### LINKED ISSUE #1132: Directory keep SGID
**What happened**:
```bash
mkdir dir1
chmod g+s  dir1
chown admin:yarn dir1
```
We lost SGID on the directory.

SGID should be kept when chown directory.

```bash
mkdir dir1
chmod g+s  dir1
chown admin:yarn dir1
```

**Environment**:
- JuiceFS version (use `./juicefs --version`) or Hadoop Java SDK version:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Object storage (cloud provider and region):
- Redis info (version, cloud provider managed or self maintained):
- Network connectivity (JuiceFS to Redis, JuiceFS to object storage):
- Others:

davies:
The man page of chown does not mention the behavior for directory, so we keep the behavior close to ext3/ext4 in Linux.

In macOS (darwin), we pick the behavior in APFS.

close #1132

### DIFF
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -341,7 +341,7 @@ func (m *baseMeta) marshal(attr *Attr) []byte {
 	return w.Bytes()
 }
 
-func clearSGID(ctx Context, cur *Attr, set *Attr) {
+func clearSUGID(ctx Context, cur *Attr, set *Attr) {
 	switch runtime.GOOS {
 	case "darwin":
 		if ctx.Uid() != 0 {
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -17,6 +17,7 @@ package meta
 
 import (
 	"fmt"
+	"runtime"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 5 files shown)


pkg/meta/sql.go
```diff
@@ -598,6 +598,30 @@ func (m *dbMeta) doGetAttr(ctx Context, inode Ino, attr *Attr) syscall.Errno {
 	return errno(err)
 }
 
+func clearSGIDSQL(ctx Context, cur *node, set *Attr) {
```
SandyXSD, line 601: maybe make cur an interface{} ?

pkg/meta/base.go
```diff
@@ -340,6 +341,30 @@ func (m *baseMeta) marshal(attr *Attr) []byte {
 	return w.Bytes()
 }
 
+func clearSGID(ctx Context, cur *Attr, set *Attr) {
```
SandyXSD, line 344: the name is a bit misleading since it may clear SUID as well

pkg/meta/sql.go
davies: not worth it

SandyXSD (approved):

PR #1137, opened 2021-12-15
cmd: add `config` command to change config of a volume

### LINKED ISSUE #1063: Find a better way to change volume quota
Currently `format` is used to update the volume quota, which is not proper

SandyXSD:
close #1063 

For now the `format` command is not touched, so it can still be used to change the config as well.  However, the new `config` is a recommended way, and `format` will be deprecated (only keep the real **FORMAT** function) after one or two minor versions.

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -16,12 +16,30 @@
 package main
 
 import (
+	"bufio"
 	"fmt"
+	"os"
 	"strings"
 
 	"github.com/juicedata/juicefs/pkg/meta"
 	"github.com/urfave/cli/v2"
 )
+
+func userConfirmed(prompt string) bool {
+	fmt.Println(prompt)
+	fmt.Print("Still proceed? [y/N]: ")
+	scanner := bufio.NewScanner(os.Stdin)
+	for scanner.Scan() {
+		if text := strings.ToLower(scanner.Text()); text == "y" || text == "yes" {
+			return true
+		} else if text == "" || text == "n" || text == "no" {
+			return false
+		} else {
+			fmt.Print("Please input y(yes) or n(no): ")
+		}
+	}
+	return false
+}
 
 func config(ctx *cli.Context) error {
 	setLoggerLevel(ctx)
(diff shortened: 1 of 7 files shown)


cmd/config.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	m := meta.NewClient(ctx.Args().Get(0), &meta.Config{Retries: 10, Strict: true})
+
+	format, err := m.Load()
+	if err != nil {
+		return err
+	}
+	var msg strings.Builder
+	for _, flag := range ctx.LocalFlagNames() {
+		switch flag {
```
davies, line 65: ignore the default value

SandyXSD: `LocalFlagNames` returns only flags explicitly specified in the command

cmd/config.go
```diff
@@ -0,0 +1,103 @@
...
+		case "capacity":
+			msg.WriteString(fmt.Sprintf("%10s: %d -> %d\n", flag, format.Capacity, ctx.Uint64(flag)))
```
davies, line 67: Show the message when it's different?

cmd/config.go
```diff
@@ -0,0 +1,103 @@
...
+		Name:  "capacity",
+		Value: 0,
+		Usage: "the limit for space in GiB",
+	},
+	&cli.Uint64Flag{
+		Name:  "inodes",
+		Value: 0,
+		Usage: "the limit for number of inodes",
+	},
+	&cli.StringFlag{
```
davies, line 37: We should have `--bucket` here to update the endpoint

cmd/config.go
```diff
@@ -0,0 +1,103 @@
...
+		Name:  "access-key",
+		Usage: "Access key for object storage (env ACCESS_KEY)",
```
davies, line 39: Should we take the ACCESS_KEY from env?

User may not want to expose the secret in command line history

cmd/config.go
SandyXSD: again the message shows up only if user input the option explicitly, so imo checking whether they're different is trivial

davies (approved):

davies (approved):

PR #1143, opened 2021-12-16
fix fmtKey to sliceKey

Shoothzj:
fix the bug founded by wechat `crud_boy`


davies (approved):

davies:
@Shoothzj It's fortunate that this bug is harmless, the wrong key is not used by others, it will not delete anything.

PR #1148, opened 2021-12-17
progressbar quiet in ci

zhijian-pro:


SandyXSD (approved):

pkg/utils/utils.go
```diff
@@ -75,8 +79,12 @@ func NewDynProgressBar(title string, quiet bool) (*mpb.Progress, *mpb.Bar) {
 
 // NewProgressCounter init a progress counter
 func NewProgressCounter(title string) (*mpb.Progress, *mpb.Bar) {
+	var quiet bool
```
SandyXSD, line 82: quiet := os.Getenv("PROGRESSBAR_DISPLAY") == "false"
or compare it directly in the `if` statment

.travis.yml
```diff
@@ -6,7 +6,7 @@ services:
   - postgresql
   - docker
 env:
-  - MINIO_TEST_BUCKET=127.0.0.1:9000/testbucket  MINIO_ACCESS_KEY=testUser  MINIO_SECRET_KEY=testUserPassword
+  - MINIO_TEST_BUCKET=127.0.0.1:9000/testbucket  MINIO_ACCESS_KEY=testUser  MINIO_SECRET_KEY=testUserPassword PROGRESSBAR_DISPLAY=false
```
SandyXSD, line 9: maybe `DISPLAY_PROGRESSBAR` would be better?

SandyXSD (changes requested):

PR #1149, opened 2021-12-17
fix mkdir/rename/delete/rmr with trailing slash

### LINKED ISSUE #1129: gateway create or delete dir wrong

davies:
Slash is a separator in JuiceFS, the trailing slash means the last component is a directory, so we can't allow upload an object with trailing slash.

close #1129

### DIFF
--- a/pkg/gateway/gateway.go
+++ b/pkg/gateway/gateway.go
@@ -618,9 +618,17 @@ func (n *jfsObjects) PutObject(ctx context.Context, bucket string, object string
 	}
 
 	p := n.path(bucket, object)
-	if strings.HasSuffix(object, sep) && r.Size() == 0 {
+	if strings.HasSuffix(object, sep) {
 		if err = n.mkdirAll(ctx, p, os.FileMode(0755)); err != nil {
 			err = jfsToObjectErr(ctx, err, bucket, object)
+			return
+		}
+		if r.Size() > 0 {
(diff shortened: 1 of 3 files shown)


pkg/gateway/gateway.go
```diff
@@ -618,11 +618,19 @@ func (n *jfsObjects) PutObject(ctx context.Context, bucket string, object string
 	}
 
 	p := n.path(bucket, object)
-	if strings.HasSuffix(object, sep) && r.Size() == 0 {
+	if strings.HasSuffix(object, sep) {
 		if err = n.mkdirAll(ctx, p, os.FileMode(0755)); err != nil {
 			err = jfsToObjectErr(ctx, err, bucket, object)
 			return
 		}
+		if r.Size() > 0 {
```
sjp00556, line 626: i think if r.Size()>0, we needn't create dir. this condition should before mkdirAll

davies: We see this behavior in MinIO, so we'd like to keep this side-affected behavior.

PR #1157, opened 2021-12-19
meta: add Reset method to clean up all metadata

SandyXSD:


pkg/meta/tkv_prefix.go
```diff
@@ -109,6 +109,10 @@ func (c *prefixClient) txn(f func(kvTxn) error) error {
 	})
 }
 
+func (c *prefixClient) reset(prefix []byte) error { // prefix ignored
+	return c.tkvClient.reset(c.prefix)
```
davies, line 113: prefix MUST be nil

pkg/meta/tkv_mem.go
```diff
@@ -256,3 +256,11 @@ func (c *memKV) txn(f func(kvTxn) error) error {
 	}
 	return nil
 }
+
+func (c *memKV) reset(prefix []byte) error { // prefix ignored
```
davies, line 260: prefix must be nil

davies (approved):

PR #1169, opened 2021-12-21
change default cache-size to 100GB

### LINKED ISSUE #1076: Increase default cache-size to 100G
Since JuiceFS will NOT fill all the space of cache-dir, so it's safe to increase the default cache-size to 100G or even higher

davies:
The Java SDK still use 100MB memory as block cache.

close #1076


SandyXSD:
Change the value in `docs/en/administration/cache_management.md` as well.

SandyXSD:
close #1076

SandyXSD (approved):

PR #1171, opened 2021-12-21
set the region of the object storage compatible with the s3 protocol from the env

zhijian-pro:
First, set the region of the object storage compatible with the s3 protocol from the environment variable

### DIFF
--- a/pkg/object/s3.go
+++ b/pkg/object/s3.go
@@ -358,7 +358,6 @@ func newS3(endpoint, accessKey, secretKey string) (ObjectStorage, error) {
 		} else {
 			// compatible s3
 			ep = uri.Host
-			region = awsDefaultRegion
 		}
 	} else {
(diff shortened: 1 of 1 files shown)


yctn:
@zhijian-pro i tested it and then indeed it works.

pkg/object/s3.go
```diff
@@ -383,11 +382,18 @@ func newS3(endpoint, accessKey, secretKey string) (ObjectStorage, error) {
...
 				bucketName = hostParts[0]
 				ep = hostParts[1]
-				region = awsDefaultRegion
 			}
 		}
 	}
 
+	if region == "" {
+		var exist bool
+		region, exist = os.LookupEnv("S3_REGION")
```
davies, line 391: We can use AWS_REGION or AWS_DEFAULT_REGION or 'us-east-1' as awsDefaultRegion

PR #1174, opened 2021-12-21
recover LTP syscalls rename03

sanwan:


sanwan:
https://github.com/juicedata/juicefs/pull/1170 has been merged,rename03 should be recovered.

PR #1177, opened 2021-12-21
Docs & SDK: update default cache size to 100GB

xiaogaozi:


davies:
No, the default cache size is 100MB in memory.

Issue #1181, opened 2021-12-21
Grafana dashboard template have a lot of mistake need to be updated


nature1995:
**What happened**:
Grafana dashboard template have a lot of mistake, such as label and unit

**What you expected to happen**:
Grafana dashboard need correct metrics, promQL and panel parameters

**How to reproduce it (as minimally and precisely as possible)**:
I will upload my correct dashboad json file for your reference

**Anything else we need to know?**

**Environment**:
- JuiceFS version (use `juicefs --version`) or Hadoop Java SDK version:
- Cloud provider or hardware configuration running JuiceFS:
- OS (e.g `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Object storage (cloud provider and region, or self maintained):
- Metadata engine info (version, cloud provider managed or self maintained):
- Network connectivity (JuiceFS to metadata engine, JuiceFS to object storage):
- Others:

PR #1183, opened 2021-12-21
docs: Fix grafana dashboard template mistake(#1181)

nature1995:
fix issue: https://github.com/juicedata/juicefs/issues/1181

### DIFF
--- a/docs/en/grafana_template_k8s.json
+++ b/docs/en/grafana_template_k8s.json
@@ -366,11 +366,11 @@
       "targets": [
         {
           "exemplar": true,
-          "expr": "sum(rate(juicefs_fuse_ops_durations_histogram_seconds_count{vol_name=\"$name\"}[1m]) < 5000000000) by (instance)",
+          "expr": "sum(rate(juicefs_fuse_ops_durations_histogram_seconds_count{vol_name=\"$name\"}[1m]) < 5000000000) by (node)",
           "format": "time_series",
           "interval": "",
           "intervalFactor": 1,
-          "legendFormat": "Write {{instance}}",
+          "legendFormat": "Write {{node}}",
           "refId": "A"
         }
       ],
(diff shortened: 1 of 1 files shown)


docs/en/grafana_template.json
```diff
@@ -15,19 +15,19 @@
...
   "links": [],
   "panels": [
     {
       "aliasColors": {},
       "bars": false,
       "dashLength": 10,
       "dashes": false,
-      "datasource": "$Source",
+      "datasource": "$datasource",
+      "description": "平均使用空间",
```
davies, line 28: Please use English

davies:
@nature1995 Thanks for sending out the PR, we will try it and get back to you.

`node` is good for K8S clusters, maybe we should create different template for K8S cluster and others.

nature1995:
> @nature1995 Thanks for sending out the PR, we will try it and get back to you.
> 
> `node` is good for K8S clusters, maybe we should create different template for K8S cluster and others.

I think we can add another file `grafana_template_k8s.json`, just replace all `instances` with `node`.

nature1995:
> @nature1995 Thanks for sending out the PR, we will try it and get back to you.
> 
> `node` is good for K8S clusters, maybe we should create different template for K8S cluster and others.

nature1995:
> @nature1995 Thanks for sending out the PR, we will try it and get back to you.
> 
> `node` is good for K8S clusters, maybe we should create different template for K8S cluster and others.

docs/en/grafana_template.json
nature1995: I have been deleted the chinese part, add the k8s copy as well

chnliyong:
We import the template on `Grafana Version 8.3.0 (commit: 914fcedb72, branch: HEAD)` and it doesn't work. 
image
If we change the **Min step** from `$interval` you added to a specified value like `1m`, it works. 
The doc [here](https://grafana.com/docs/grafana/latest/datasources/prometheus/#query-editor-in-dashboards) seems not support user defined variable:
> Use ‘Minimum’ or ‘Maximum’ step mode to set the lower or upper bounds respectively on the interval between data points. For example, set “minimum 1h” to hint that measurements are not frequent (taken hourly). Use the ‘Exact’ step mode to set a precise interval between data points. $__interval and $__rate_interval are supported.

nature1995:
> We import the template on `Grafana Version 8.3.0 (commit: 914fcedb72, branch: HEAD)` and it doesn't work. image If we change the **Min step** from `$interval` you added to a specified value like `1m`, it works. The doc [here](https://grafana.com/docs/grafana/latest/datasources/prometheus/#query-editor-in-dashboards) seems not support user defined variable:
> 
> > Use ‘Minimum’ or ‘Maximum’ step mode to set the lower or upper bounds respectively on the interval between data points. For example, set “minimum 1h” to hint that measurements are not frequent (taken hourly). Use the ‘Exact’ step mode to set a precise interval between data points. $__interval and $__rate_interval are supported.

Do you select the sampling optioin which on the top, I think you need choice one from the list ? 
I have removed the interval option, please check the newest commit. If you have the same issue, please let me know.

PR #1204, opened 2021-12-27
add sftp test and fix ipv4 addr

zhijian-pro:


zhijian-pro:
The problem happened in the unit test and not here

PR #1207, opened 2021-12-27
s3: support anonymous access

201341:

### DIFF
--- a/pkg/object/s3.go
+++ b/pkg/object/s3.go
@@ -396,6 +396,8 @@ func newS3(endpoint, accessKey, secretKey string) (ObjectStorage, error) {
 	}
 	if accessKey != "" {
 		awsConfig.Credentials = credentials.NewStaticCredentials(accessKey, secretKey, "")
+	} else if accessKey == "" && secretKey == "" {
+		awsConfig.Credentials = credentials.AnonymousCredentials
 	}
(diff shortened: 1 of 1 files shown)


pkg/object/s3.go
```diff
@@ -396,6 +396,8 @@ func newS3(endpoint, accessKey, secretKey string) (ObjectStorage, error) {
 	}
 	if accessKey != "" {
 		awsConfig.Credentials = credentials.NewStaticCredentials(accessKey, secretKey, "")
+	} else if accessKey == "" && secretKey == "" {
+		awsConfig.Credentials = credentials.AnonymousCredentials
```
davies, line 400: When the accessKey is empty, we usually use IAM role or other temporary token to access S3.

What's the current behavior to access public buckets? And, what's the use case that a public bucket is used with JuiceFS?

201341: For public buckets. the juicefs sync can't downloads the file from s3. If the IAM is disabled and no temporary token , I can't use juicefs sync.

davies: I see. The current patch will break IAM, we need another way to support anonymous access

201341: ok, I will close the pr, and pull an issue.

PR #1208, opened 2021-12-27
cmd/sync: add `check-all` and `check-new` options to compare checksums

SandyXSD:


pkg/sync/sync.go
```diff
@@ -475,19 +557,25 @@ func producer(tasks chan<- object.Object, src, dst object.ObjectStorage, config
...
+		if dstobj == nil || obj.Key() < dstobj.Key() {
 			tasks <- obj
-		} else { // obj.Key() == dstobj.Key()
-			if config.DeleteSrc && obj.Size() == dstobj.Size() {
+		} else { // obj.key == dstobj.key
+			if config.ForceUpdate ||
+				(config.Update && obj.Mtime().After(dstobj.Mtime())) ||
+				(!config.Update && obj.Size() != dstobj.Size()) {
+				tasks <- obj
+			} else if config.Update { // obj.mtime <= dstobj.mtime
```
davies, line 567: We should verify the checksum when obj.mtime == dstobj.mtime

davies: source and destination may have different time resolution, should we round it to seconds?

pkg/sync/sync.go
```diff
@@ -176,96 +179,151 @@ var bufPool = sync.Pool{
...
+	logger.Debugf("Copied permissions (%s:%s:%s) for %s in %s", fi.Owner(), fi.Group(), fi.Mode(), key, time.Since(start))
+}
 
+func doCheckSum(src, dst object.ObjectStorage, key string, equal *bool) error {
 	buf := bufPool.Get().(*[]byte)
 	defer bufPool.Put(buf)
-	if _, e = io.CopyBuffer(f, in, *buf); e != nil {
-		return e
+
+	srcIn, err := src.Get(key, 0, -1)
```
davies, line 225: limit the bandwidth and concurrent connections

pkg/sync/sync.go
```diff
@@ -176,96 +179,151 @@ var bufPool = sync.Pool{
...
+	if err != nil {
+		return fmt.Errorf("src get: %s", err)
+	}
+	defer srcIn.Close()
+	srcH := md5.New()
+	if _, err = io.CopyBuffer(srcH, srcIn, *buf); err != nil {
```
davies, line 231: we read from two streams in the same time, should we just compare the bytes?

Or, we can compare the checksum block by block to find the different earlier.

davies: For larger objects, we can also compare them in parallel.

pkg/sync/sync.go
```diff
@@ -309,74 +366,99 @@ func copyInParallel(src, dst object.ObjectStorage, obj object.Object) error {
...
+					deleteObj(src, key, false)
+				} else if config.Perms { // FIXME: && needCopyPerms(obj, dstobj)
+					copyPerms(dst, obj)
+					copied.Increment()
+				} else {
+					skipped.Increment()
+				}
+				break
+			}
+			// checkSum failed or not equal, copy the object
```
davies, line 432: For failed check, should we mark it as failed?

pkg/sync/sync.go
```diff
@@ -309,74 +366,99 @@ func copyInParallel(src, dst object.ObjectStorage, obj object.Object) error {
...
+			if config.Dry {
+				logger.Infof("Will compare checksum for %s", key)
+				break
+			}
+			obj = obj.(*withSize).Object
+			if equal, err := checkSum(src, dst, key, obj.Size()); err == nil && equal {
+				if config.DeleteSrc {
+					deleteObj(src, key, false)
+				} else if config.Perms { // FIXME: && needCopyPerms(obj, dstobj)
+					copyPerms(dst, obj)
```
davies, line 425: check the permission (`dst.Head(key)`)before copy

pkg/sync/sync.go
SandyXSD: Copying the object might fix the problem if error occurs in the destination side.

SandyXSD: Maybe it is safer to just mark it failed and let users to double check it.

pkg/sync/sync.go
SandyXSD: Parallelism is not implemented because Get() returns a reader, not readerat

pkg/sync/sync.go
```diff
@@ -171,101 +173,173 @@ func ListAll(store object.ObjectStorage, start, end string) (<-chan object.Objec
 
 var bufPool = sync.Pool{
 	New: func() interface{} {
-		buf := make([]byte, 32<<10)
+		buf := make([]byte, defaultPartSize)
```
SandyXSD, line 176: @davies not sure if this change is reasonable

davies: Don't, defaultPartSize is too large as an buffer

pkg/sync/sync.go
davies: Get(key, offset, limit)

davies (approved):

PR #1225, opened 2021-12-29
add s3 gateway deploy in k8s

zwwhdls:
add s3gateway deploy yaml/chart and update doc


yuhr123 (approved):
There is no problem with the document part.

xiaogaozi (approved):

PR #1228, opened 2021-12-29
s3 support anonymous access

### LINKED ISSUE #1213: Support anonymous access
As https://github.com/juicedata/juicefs/pull/1207 disscused, juicefs need an better way to support anonymous access.

zhijian-pro:
After all configurations are loaded , try the `list` first. If the error is `ErrNoValidProvidersFoundInChain`, then enable anonymous access.

Close #1213


pkg/object/s3.go
```diff
@@ -413,6 +413,10 @@ func newS3(endpoint, accessKey, secretKey string) (ObjectStorage, error) {
 	if err != nil {
 		return nil, fmt.Errorf("Fail to create aws session: %s", err)
 	}
+	_, err = (&s3client{bucketName, s3.New(ses), ses}).List("", "", 1)
```
davies, line 416: Test it only when accessKey is ''

davies: It's better to avoid the check and find out that no valid providers

PR #1247, opened 2022-01-04
fix errcheck

zhijian-pro:
This pr is based on #1245   and has processed all errorcheck except defer


davies:
#1245 is merged, please fix the conflect.

PR #1263, opened 2022-01-09
feat: Object storage support go plugin

myml:
## What
This commit implements the go plugin in the object store.
Juicefs now supports many standard protocol storage systems, such as file, s3, webdav, plugin allows jusefs to support more private internal storage systems and does not pollute the juicefs repository.

## Example
format a juicefs file system using the keyvalue storage plugin
```bash
juicefs format --storage goplugin --bucket goplugin:///lib/juicefs-bolt-storage.so sqlite3://myjfs.db myjfs
```
The source code for `juicefs-bolt-storage` is in this [repository](https://github.com/myml/juicefs-bolt-storage) which is implemented with go's kv library blob

Another plugin, and what prompted me to write this pull request: [juicefs-aliyundrive-storage](https://github.com/myml/juicefs-aliyundrive-storage) enables juicefs to support aliyun drive.

## Future

Go plugin has quite a few problems, Get can be replaced with other plug-in library.

### DIFF
--- a/pkg/object/goplugin_linux.go
+++ b/pkg/object/goplugin_linux.go
@@ -1,0 +1,174 @@
+//go:build !nogoplugin
+// +build !nogoplugin
+
+/*
+ * JuiceFS, Copyright (C) 2020 Juicedata, Inc.
+ *
+ * This program is free software: you can use, redistribute, and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3
+ * or later ("AGPL"), as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package object
+
+import (
+	"errors"
+	"io"
+	"net/url"
+	"plugin"
+)
+
+type PluginInterface interface {
+	// Description of the object storage.
+	String() string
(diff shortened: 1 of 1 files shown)


davies:
@myml Good ideal! 

The plugin may need to know the account or other identity to access the storage service, I'm thinking of use plugin like this:

For a storage A, we can implement as a go plugin and is built as libA.so, then we can use it like this:

```
juicefs format --storage A  --bucket XXXX --access-key XXX --secret-key XXX
```

juicefs will try to load libX.so from a list of library path if it's unknown.

Currently, we have difficulty to publish the libRADOS support with JuiceFS, then we can try to implement it as a plugin.

myml:
Nice!  That is what I thought at first, but I was worried about the downside of the go plugin. go plugin is not the best option by default. Maybe it would be better to do this after using more mature plugins.
What do you think?

myml:
The current bucket,access and secret will be passed to the go plugin, More parameters can be passed using query.
example 
```bash
juicefs format --storage goplugin --bucket juicefs-xxx-storage.so?dir=xxx --secret-key xxx sqlite3://myjfs.db myjfs
```
go plugin get param
```go
func New(bucket, accessKey, secretKey string) (interface{}, error) {
	return NewAliyun(bucket, accessKey, secretKey)
}
func NewAliyun(u, user, passwd string) (*Aliyun, error) {
	cfgURL, err := url.Parse(u)
	aliyunDir := cfgURL.Query().Get("dir")
        ...
}
```

davies:
@myml I think we should pick one rather than support two of them.

Currently, we are focus on 1.0 release, will look into this for next release.

davies:
Can you create a issue to track it?

myml:
ok. https://github.com/juicedata/juicefs/issues/1271

PR #1281, opened 2022-01-11
add metrics port

### LINKED ISSUE #1274: The K8s deployment file of S3 gateway should exports P8s metric port

Export Prometheus metric port of S3 gateway in the [K8s deployment file](https://github.com/juicedata/juicefs/blob/main/deploy/juicefs-s3-gateway.yaml)

**Why is this needed**:

After export the Prometheus metric port, user could use [`ServiceMonitor`](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/getting-started.md) CRD provided by Prometheus Operator to collect metrics easily.

zwwhdls:
fix https://github.com/juicedata/juicefs/issues/1274

### DIFF
--- a/deploy/juicefs-s3-gateway.yaml
+++ b/deploy/juicefs-s3-gateway.yaml
@@ -29,7 +29,7 @@ spec:
           command:
             - sh
             - -c
-            - juicefs gateway $(METAURL) $(NODE_IP):9000 --metrics=0.0.0.0:9567
+            - juicefs gateway $(METAURL) $(NODE_IP):9000 --metrics=$(NODE_IP):9567
           env:
             - name: NODE_IP
               valueFrom:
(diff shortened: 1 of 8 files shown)


deploy/juicefs-s3-gateway.yaml
```diff
@@ -29,7 +29,7 @@ spec:
           command:
             - sh
             - -c
-            - juicefs gateway $(METAURL) $(NODE_IP):9000
```
davies, line 32: Why not use NODE_IP?

zwwhdls: Using pod ip, NODE_IP is env from podIP

```
          env:
            - name: NODE_IP
              valueFrom:
                fieldRef:
                  fieldPath: status.podIP
```

davies: I meant `--metrics=${NODE_IP}:9567`

zwwhdls: Both ${NODE_IP} and 0.0.0.0 is ok, because it's separate net namespace in container

PR #1289, opened 2022-01-13
format: Only try to create bucket when it really doesn't exist

xiaogaozi:
When the bucket does not exist, the object store will return the corresponding error code `NoSuchBucket` (e.g. [S3](https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#ErrorCodeList), [OSS](https://help.aliyun.com/document_detail/185804.htm?#title-p5t-cwc-kcv), [COS](https://cloud.tencent.com/document/product/436/7730#.E9.94.99.E8.AF.AF.E7.A0.81.E5.88.97.E8.A1.A8)). This PR fix creates a bucket only when it really does not exist.

Incorrect conditional judgments can cause the log to output something that is confusing to the user, e.g. `please create bucket XXX manually, then format again`.

### DIFF
--- a/cmd/format.go
+++ b/cmd/format.go
@@ -91,16 +91,16 @@ func randSeq(n int) string {
 
 func doTesting(store object.ObjectStorage, key string, data []byte) error {
 	if err := store.Put(key, bytes.NewReader(data)); err != nil {
-		if strings.Contains(err.Error(), "Access Denied") {
-			return fmt.Errorf("Failed to put: %s", err)
+		if strings.Contains(err.Error(), "NoSuchBucket") {
+			if err2 := store.Create(); err2 != nil {
+				return fmt.Errorf("Failed to create %s: %s, previous error: %s\nPlease create bucket %s manually, then format again.",
+					store, err2, err, store)
+			}
+			if err := store.Put(key, bytes.NewReader(data)); err != nil {
+				return fmt.Errorf("Failed to put: %s", err)
+			}
 		}
-		if err2 := store.Create(); err2 != nil {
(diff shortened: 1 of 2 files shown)


cmd/format.go
```diff
@@ -91,16 +91,16 @@ func randSeq(n int) string {
...
-		if strings.Contains(err.Error(), "Access Denied") {
-			return fmt.Errorf("Failed to put: %s", err)
-		}
-		if err2 := store.Create(); err2 != nil {
-			return fmt.Errorf("Failed to create %s: %s,  previous error: %s\nplease create bucket %s manually, then format again",
-				store, err2, err, store)
-		}
-		if err := store.Put(key, bytes.NewReader(data)); err != nil {
-			return fmt.Errorf("Failed to put: %s", err)
+		if strings.Contains(err.Error(), "NoSuchBucket") {
```
SandyXSD, line 94: there is no guarantee that all object stores are using this error message

xiaogaozi: Yes, that is a reasonable assumption. But most object stores follow the S3 API, so this conditional logic should be able to handle most cases.

xiaogaozi: If some object stores don't follow this convention it doesn't have a big impact on this logic.

xiaogaozi: I've updated the description of this PR to explain why this fix is necessary.

davies: After this change, we will not try to create the bucket if different message returned, for example ceph, we will not create the pool.

xiaogaozi: How much of an impact would it have if the bucket could not be created automatically?

davies (approved):
Please update the title and description to reflect the latest changes.

PR #1303, opened 2022-01-17
Upgrade Qingstor  SDK and support private region

zhijian-pro:


pkg/object/qingstor.go
```diff
@@ -255,23 +255,27 @@ func newQingStor(endpoint, accessKey, secretKey string) (ObjectStorage, error) {
...
-	zone := hostParts[1]
-
+	var bucketName, zone, host string
+	if !strings.HasSuffix(uri.Host, "qingstor.com") {
+		// support private cloud
+		hostParts := strings.SplitN(uri.Host, ".", 2)
+		bucketName, zone, host = hostParts[0], "", hostParts[1]
+	} else {
+		hostParts := strings.SplitN(uri.Host, ".", 3)
+		bucketName, zone, host = hostParts[0], hostParts[1], hostParts[2]
```
davies, line 265: Should the zone be part of host? Please add an example here.

zhijian-pro: This is old logic and has not changed.

endpoint="http://zhjijian-test.sh1a.qingstor.com"

image

image

The zone is indeed part of the host.

davies: I meant the `host` should be `sh1a.qingstor.com`, not `qingstor.com`, right?

zhijian-pro: ### host does not include region
image

PR #1312, opened 2022-01-19
utils/logger: colorize outputs to terminal

SandyXSD:


pkg/utils/logger.go
```diff
@@ -67,11 +72,19 @@ func (l *logHandle) Log(args ...interface{}) {
...
+	if l.tty {
+		l.prefixes = make([]string, 7)
+		ps := []string{"PANIC", "FATAL", "ERROR", " WARN", " INFO", "DEBUG", "TRACE"}
+		for i := 0; i < 3; i++ {
+			l.prefixes[i] = fmt.Sprintf("\033[1;%dm%s\033[0m", 31, ps[i])
+		}
+		for i := 3; i < 6; i++ {
+			l.prefixes[i] = fmt.Sprintf("\033[1;%dm%s\033[0m", 30+i, ps[i])
+		}
+		l.prefixes[6] = fmt.Sprintf("\033[1;%dm%s\033[0m", 35, ps[6])
```
davies, line 85: Should we provide a flag to disable this?

SandyXSD: Better do it; but it may be conflicted with flags of subcommands, like:
```
./juicefs stats -h

   --nocolor          disable colors (default: false)
```

davies: we can use it as a global option.

SandyXSD: Workable

davies (approved):

PR #1318, opened 2022-01-19
utils/logger: fix logger format

SandyXSD:


SandyXSD:
Test output:
Screen Shot 2022-01-19 at 6 41 26 PM

PR #1333, opened 2022-01-25
meta: hide password in meta-url

SandyXSD:


pkg/meta/interface.go
```diff
@@ -357,11 +358,23 @@ func Register(name string, register Creator) {
...
+	uri2 := removePassword(uri)
+	logger.Infof("Meta address: %s", uri2)
+	if hidePasswd && uri2 != uri {
+		for i, a := range os.Args {
+			if a == uri || changed && a == uri[8:] {
+				os.Args[i] = uri2
+				break
+			}
+		}
+		gspt.SetProcTitle(strings.Join(os.Args, " "))
```
davies, line 376: We should move this into cmd/

davies (changes requested):
It's better to  change os.Args and proctitle in cmd/ package.

davies (approved):

PR #1335, opened 2022-01-25
improve test for gc command

sanwan:
Now files of modified time in one hour will not  gc. It's not conventient to test .  An enviroment variable  can be used to decrease the waiting time.


SandyXSD:
IMHO it should be better to just add a new flag if we really need this.

cmd/gc_test.go
```diff
@@ -43,10 +71,21 @@ func TestGc(t *testing.T) {
 		}
 	}
 
-	fsckArgs := []string{"", "gc", metaUrl}
-	err := Main(fsckArgs)
+	strEnvSkippedTime := os.Getenv("JFS_GC_SKIPPEDTIME")
+	t.Logf("JFS_GC_SKIPPEDTIME is %s\n",strEnvSkippedTime)
+
+	WriteLeakedData(dataDir)
+	time.Sleep(time.Duration(65)*time.Second)
```
davies, line 78: This will slow down tests

cmd/gc_test.go
```diff
@@ -43,10 +71,21 @@ func TestGc(t *testing.T) {
 		}
 	}
 
-	fsckArgs := []string{"", "gc", metaUrl}
-	err := Main(fsckArgs)
+	strEnvSkippedTime := os.Getenv("JFS_GC_SKIPPEDTIME")
+	t.Logf("JFS_GC_SKIPPEDTIME is %s\n",strEnvSkippedTime)
```
davies, line 75: Logf will add new line in the end, no `\n` needed here.

cmd/gc_test.go
```diff
@@ -43,10 +71,21 @@ func TestGc(t *testing.T) {
...
+
+	WriteLeakedData(dataDir)
+	time.Sleep(time.Duration(65)*time.Second)
+
+	gcArgs := []string{"", "gc", "--delete",metaUrl}
```
davies, line 80: go fmt

cmd/gc_test.go
```diff
@@ -43,10 +71,21 @@ func TestGc(t *testing.T) {
...
+	err := Main(gcArgs)
 	if err != nil {
-		t.Fatalf("fsck failed: %v", err)
+		t.Fatalf("gc failed: %s\n", err)
 	}
 
+	gcArgs = []string{"", "gc", metaUrl}
+	err = Main(gcArgs)
+	if err != nil {
+		t.Fatalf("gc failed: %s\n", err)
```
davies, line 89: Should we check that the leaked data is deleted?

.travis.yml
```diff
@@ -31,8 +31,8 @@ before_script:
   - sudo mysql -e "create database dev;"
   - psql -c 'create database test;' -U postgres
   - curl --proto '=https' --tlsv1.2 -sSf https://tiup-mirrors.pingcap.com/install.sh | sh && source ~/.bash_profile && nohup tiup playground --mode tikv-slim >> output.log 2>&1 &
-  - docker run -d -p 9000:9000 -p 9001:9001 -e "MINIO_ROOT_USER=testUser" -e "MINIO_ROOT_PASSWORD=testUserPassword" quay.io/minio/minio server /data --console-address ":9001"
-  - wget https://dl.min.io/client/mc/release/linux-amd64/mc && chmod +x mc && ./mc config host add local http://127.0.0.1:9000 testUser testUserPassword && ./mc mb local/testbucket
+  - docker run -d -p 9000:9000 -p 9001:9001 -e "MINIO_ROOT_USER=testUser" -e "MINIO_ROOT_PASSWORD=testUserPassword" quay.io/minio/minio:RELEASE.2022-01-25T19-56-04Z server /data --console-address ":9001"
+  - go install github.com/minio/mc@RELEASE.2022-01-07T06-01-38Z && mc config host add local http://127.0.0.1:9000 testUser testUserPassword && mc mb local/testbucket
```
davies, line 35: wget is faster than `go install`

cmd/gc_test.go
sanwan: ok

.travis.yml
sanwan: This is from https://github.com/juicedata/juicefs/pull/1349 . The reason for using go install here  is not easy to use wget to download specific mc version.(I guess)

cmd/gc_test.go
sanwan: ok

cmd/gc_test.go
sanwan: I will reduce the time.

PR #1359, opened 2022-01-29
meta: exit process if no setting found

SandyXSD:


SandyXSD:
Force quit the process is way too dangerous

PR #1379, opened 2022-02-07
Meta: format use `META_PASSWORD` environment variables

zhijian-pro:
The format command uses the environment `META_PASSWORD` variable, which is consistent with the `mount` command


yuhr123 (approved):

SandyXSD:
https://github.com/juicedata/juicefs/blob/main/pkg/meta/interface.go#L354
`META_PASSWORD` is already checked here.

zhijian-pro:
Only need to supplement the document, the original code has already implemented this function.   T_T‘

PR #1392, opened 2022-02-09
CI:if tikv service is not ready,show failed log

sanwan:


.travis.yml
```diff
@@ -38,7 +38,8 @@ before_script:
   - sudo make -C fstests setup
   - chmod +x travis-setup-hdfs.sh
   - ./travis-setup-hdfs.sh
-  - for i in {2222,2379,3306,5432,8020,9000} ; do nc -z localhost $i;if [ $? != 0 ];then sleep 5; else continue; fi; nc -z localhost $i; if [ $? != 0 ];then echo "service not ready, port:"$i; exit 1;fi; done
+  - for i in {2222,3306,5432,8020,9000} ; do nc -z localhost $i;if [ $? != 0 ];then sleep 5; else continue; fi; nc -z localhost $i; if [ $? != 0 ];then echo "service not ready, port:"$i; exit 1;fi; done
+  - nc -z localhost 2379;if [ $? != 0 ];then echo "tikv is not ready";cat output.log;exit 1; fi
```
zhijian-pro, line 42: if tikv not ready, it will exit after printing port, line 42 will not be executed

zhijian-pro (changes requested):

.travis.yml
davies, line 42: Should we wait longer and double check it?

sanwan: In travis debug mode,I try to start tikv and tikv start fast with tikv-slim mode.Before we check here ,more than one minute has passed. It is not double check.In for loop,there is no check 2379 port.

davies: lgtm

PR #1397, opened 2022-02-10
meta: limit the number when scanning sessions/files

SandyXSD:


pkg/meta/base.go
```diff
@@ -203,7 +203,7 @@ func (m *baseMeta) refreshSession() {
 }
 
 func (m *baseMeta) CleanStaleSessions() {
-	sids, err := m.en.doFindStaleSessions(time.Now().Add(time.Minute * -5).Unix())
+	sids, err := m.en.doFindStaleSessions(time.Now().Add(time.Minute*-5).Unix(), 100)
```
davies, line 206: Right now, we can only do it once in a minute, we should increase the limit, for example 1000.

pkg/meta/base.go
```diff
@@ -274,7 +274,7 @@ func (m *baseMeta) cleanupDeletedFiles() {
 		if ok, err := m.en.setIfSmall("lastCleanupFiles", time.Now().Unix(), 60); err != nil {
 			logger.Warnf("checking counter lastCleanupFiles: %s", err)
 		} else if ok {
-			files, err := m.en.doFindDeletedFiles(time.Now().Add(-time.Hour).Unix())
+			files, err := m.en.doFindDeletedFiles(time.Now().Add(-time.Hour).Unix(), 1000)
```
davies, line 277: 100000

pkg/meta/tkv.go
```diff
@@ -486,8 +486,8 @@ func (m *kvMeta) doCleanStaleSession(sid uint64) {
 	}
 }
 
-func (m *kvMeta) doFindStaleSessions(ts int64) ([]uint64, error) {
-	vals, err := m.scanValues(m.fmtKey("SH"), func(k, v []byte) bool {
+func (m *kvMeta) doFindStaleSessions(ts int64, limit int) ([]uint64, error) {
+	vals, err := m.scanValues(m.fmtKey("SH"), limit, func(k, v []byte) bool {
```
davies, line 490: When the first 1000 sessions are not stale, we will get nothing. So we should limit the stale session, not scanned sessions.

SandyXSD: It limits the number of results, i.e. after the filter.

pkg/meta/base.go
SandyXSD: OK

pkg/meta/base.go
SandyXSD: 10k might be enough. 60s * 100 deletes/s = 6k

davies (approved):

PR #1414, opened 2022-02-11
CI:add travis cache

sanwan:


davies:
@sanwan Can we also add cache for other downloaded files, for example, docker images, tikv, minio?

sanwan:
> 

I will try.

PR #1420, opened 2022-02-14
cmd/format: prevend reusing object storage when formating a volume

SandyXSD:


cmd/format.go
```diff
@@ -216,6 +217,19 @@ func format(c *cli.Context) error {
...
+				if o == nil {
+					logger.Fatalf("List storage %s failed", blob)
+				} else if o.IsDir() {
+					continue
+				} else if !strings.HasPrefix(o.Key(), "testing/") {
+					logger.Fatalf("Storage %s is not empty; please clean it up or pick another volume name", blob)
+				}
+			}
+		} else {
+			logger.Fatalf("Storage %s is not configured correctly: %s", blob, err)
```
davies, line 231: Some object stores do not implement List(), or very slow, we should show a warning, rather than Fatal

cmd/format.go
```diff
@@ -216,6 +217,19 @@ func format(c *cli.Context) error {
 		if err := test(blob); err != nil {
 			logger.Fatalf("Storage %s is not configured correctly: %s", blob, err)
 		}
+		if objs, err := osync.ListAll(blob, "", ""); err == nil {
+			for o := range objs {
+				if o == nil {
+					logger.Fatalf("List storage %s failed", blob)
```
davies, line 223: Warning

PR #1421, opened 2022-02-14
Docs: add doc about storage class of object storage

### LINKED ISSUE #1043: Support s3 storageclass
For s3-like object storage backend, storageclass is used to control redundancy level of a object (https://docs.aws.amazon.com/AmazonS3/latest/userguide/storage-class-intro.html) and it has been implemented by s3 client library. It would be great if we could have this property on juicefs format to let user have the choice of what level of redundancy they want and reduce the cost of s3.

**Why is this needed**:
I want to use “Reduced Redundancy” level of storageclass for non-important files.

xiaogaozi:
Fix #1043


yuhr123 (approved):

docs/en/reference/how_to_setup_object_storage.md
```diff
@@ -57,6 +57,14 @@ JuiceFS also provides flexible support for this object storage service that dist
 
 Creating a file system using an internal Endpoint ensures better performance and lower latency, and for clients that cannot be accessed through an internal address, you can specify a public Endpoint to mount with the `--bucket` option.
 
+## Storage Class
+
+Object storage usually supports multiple storage classes, such as standard storage, infrequent access storage, and archive storage. When creating an object storage bucket you can choose the appropriate storage class according to your actual needs, or automatically convert the storage class of existing objects through lifecycle management. Regardless of the storage class used, it is supported as the underlying data storage for JuiceFS.
```
davies, line 62: We can't use the archive (or deep archive) storage class, the objects with that class is not accessible inmediately (requring a restore process).

xiaogaozi: I'll update the document later.

davies: Why not do that in this one?

xiaogaozi: Yes, I mean update this PR later.

PR #1432, opened 2022-02-16
docs: Update for_local.md storage option to s3

AIXjing:
As the practice example uses AWS, it would be better to set --storage option to s3 directly.


yuhr123 (approved):

yuhr123:
Hey @AIXjing , thank you for your contribution! 🎉

PR #1436, opened 2022-02-16
progress bar add write interface

zhijian-pro:
Based on [mbp](github.com/vbauerster/mpb) v7.0.3 version code modification, add write interface to it, to achieve the effect that the progress bar and the log will not overlap each other

### DIFF
--- a/pkg/utils/logger.go
+++ b/pkg/utils/logger.go
@@ -131,7 +131,8 @@ func SetOutFile(name string) {
 	}
 }
 
-func SetOutput(w io.Writer, istty bool) {
+func SetOutput(w io.Writer) {
+	istty := isatty.IsTerminal(os.Stderr.Fd())
 	for _, logger := range loggers {
 		logger.SetOutput(w)
 		logger.tty = istty
--- a/pkg/utils/logger.go
+++ b/pkg/utils/logger.go
@@ -132,9 +132,7 @@ func SetOutFile(name string) {
 }
 
 func SetOutput(w io.Writer) {
-	colorful := isatty.IsTerminal(os.Stderr.Fd())
 	for _, logger := range loggers {
 		logger.SetOutput(w)
-		logger.colorful = colorful
 	}
 }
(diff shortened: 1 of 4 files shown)


zhijian-pro:
**GC** 
image
**unit test**
image

davies (approved):

pkg/utils/progress.go
```diff
@@ -150,6 +153,7 @@ func (p *Progress) Done() {
 		}
 	}
 	p.Progress.Wait()
+	SetOutput(os.Stderr, true)
```
SandyXSD, line 156: It's possible to use progress bar in a mounted client, e.g: backup will call DumpMeta

SandyXSD (changes requested):
need fix

pkg/utils/logger.go
```diff
@@ -129,3 +130,10 @@ func SetOutFile(name string) {
 		logger.tty = false
 	}
 }
+
+func SetOutput(w io.Writer, istty bool) {
+	for _, logger := range loggers {
+		logger.SetOutput(w)
+		logger.tty = istty
```
davies, line 137: tty is used by DisableLogColor(), we need to add `colorful` for that rather than using `tty=false`

zhijian-pro:
> 
stderr stdout is redirected to a file during mount
image

image
Test backup meta when mount is ok
image

pkg/utils/logger.go
```diff
@@ -126,6 +127,14 @@ func SetOutFile(name string) {
 	}
 	for _, logger := range loggers {
 		logger.SetOutput(file)
-		logger.tty = false
+		logger.colorful = false
+	}
+}
+
+func SetOutput(w io.Writer) {
+	colorful := isatty.IsTerminal(os.Stderr.Fd())
```
davies, line 135: We should not change it.

davies (approved):

SandyXSD (approved):

Issue #1441, opened 2022-02-17
Add WebDAV protocol to JuiceFS


davies:
WebDAV is a protocol to access files using HTTP, similar to S3, but much simpler.  It's better to support that.

PR #1444, opened 2022-02-17
support webdav protocol

zhijian-pro:
support webdav protocol #1441 #1378

### DIFF
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -29,6 +29,8 @@ import (
 	"syscall"
 	"time"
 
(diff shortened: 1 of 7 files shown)


cmd/mount.go
```diff
@@ -294,6 +296,13 @@ func mount(c *cli.Context) error {
 	if !c.Bool("no-usage-report") {
 		go usage.ReportUsage(m, version.Version())
 	}
+	if c.String("http") != "" {
```
davies, line 299: We can add `webdav` command

PR #1466, opened 2022-02-21
docs: ascii punctuations should be used in english documentations

rayw000:
Some punctuations scattered in `docs/en` should be ASCII chars but unicode currently.


yuhr123 (approved):
Hi @rayw000 Thank you for the contribution!🍻

PR #1468, opened 2022-02-22
s3: set ContentType when putting object

201341:

### DIFF
--- a/pkg/object/ibmcos.go
+++ b/pkg/object/ibmcos.go
@@ -31,6 +31,7 @@ import (
 	"github.com/IBM/ibm-cos-sdk-go/aws/credentials/ibmiam"
 	"github.com/IBM/ibm-cos-sdk-go/aws/session"
 	"github.com/IBM/ibm-cos-sdk-go/service/s3"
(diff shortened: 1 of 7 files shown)


davies:
This is useful for `juicefs sync`, can you add it for all object storage, or most of them?

201341:
> This is useful for `juicefs sync`, can you add it for all object storage, or most of them?

ok。

pkg/object/ibmcos.go
```diff
@@ -79,10 +81,15 @@ func (s *ibmcos) Put(key string, in io.Reader) error {
 		}
 		body = bytes.NewReader(data)
 	}
+	mimeType := mime.TypeByExtension(path.Ext(key))
+	if !strings.ContainsRune(mimeType, '/') {
```
davies, line 85: we can have a function there these lines:  guessMimeType

201341: which file should I put this function on? utils.go or object_storage.go?

davies: utils.go is good

PR #1469, opened 2022-02-22
meta: add a format Version to reject incompatible clients

SandyXSD:


pkg/version/version.go
```diff
@@ -16,16 +16,45 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"golang.org/x/mod/semver"
+)
 
 var (
-	version      = "1.0-dev"
+	version      = "1.0.0-beta"
```
davies, line 27: dev is invalid? maybe "alpha"

SandyXSD: should be valid

pkg/version/version.go
```diff
@@ -14,18 +14,94 @@
...
+}
+
 func Version() string {
-	return fmt.Sprintf("%v (%v %v)", version, revisionDate, revision)
+	pr := ver.preRelease
+	if pr != "" {
+		pr = "-" + pr
+	}
+	return fmt.Sprintf("%d.%d.%d%s+%s.%s",
+		ver.major, ver.minor, ver.patch, pr, ver.revisionDate, ver.revision)
```
davies, line 48: ver.revisionDate could be invalid, just ignore it

pkg/meta/tkv.go
```diff
@@ -337,6 +337,7 @@ func (m *kvMeta) Init(format Format, force bool) error {
 			old.Capacity = format.Capacity
 			old.Inodes = format.Inodes
 			old.TrashDays = format.TrashDays
+			old.ClientVersions = format.ClientVersions
```
davies, line 340: AllowedClientVersion

davies: Or ClientVersionRequirement

pkg/meta/config.go
```diff
@@ -55,3 +62,34 @@ func (f *Format) RemoveSecret() {
 		f.EncryptKey = "removed"
 	}
 }
+
+func (f *Format) CheckVersion() bool {
+	if f.MetaVersion > 1 {
+		return false
```
davies, line 68: Add logging

pkg/meta/config.go
```diff
@@ -55,3 +62,34 @@ func (f *Format) RemoveSecret() {
...
+	}
+	if ok && len(ps) > 1 {
+		if r, err := version.Compare(ps[1]); err == nil {
+			ok = r <= 0
+		} else {
+			logger.Errorf("Compare versions: %s", err)
+			return false
+		}
+	}
+	return ok
```
davies, line 94: Add meanlingful logging if it not matched

davies: Or, return the error message

pkg/version/version.go
SandyXSD: will be set as "unknown"

pkg/meta/tkv.go
SandyXSD: split to two members: `MinClientVersion` and `MaxClientVersion`

pkg/meta/config.go
SandyXSD: `CheckVersion` returns an error

pkg/meta/config.go
SandyXSD: same as above

pkg/meta/tkv.go
SandyXSD: also can only be modified using `config` subcommand

SandyXSD: also can only be modified using `config` subcommand

cmd/config.go
```diff
@@ -139,6 +151,12 @@ func config(ctx *cli.Context) error {
 				return fmt.Errorf("Aborted.")
 			}
 		}
+		if clientVer && format.CheckVersion() != nil {
+			warn("Clients with the same version of this will be rejected after modification.")
```
davies, line 155: Should we check the version of all active sessions?

davies (approved):

cmd/config.go
SandyXSD: IMO that's too much for `config` command

PR #1483, opened 2022-02-24
fix readme.md part2

AIXjing:
Went through the remained part of readme.md and made a few changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ The metadata of file system will be stored via JuiceFS through Redis, which is a
 
 ![JuiceFS Storage Format](docs/en/images/juicefs-storage-format-new.png)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -104,12 +103,12 @@ Files=235, Tests=8813, 233 wallclock secs ( 2.77 usr  0.38 sys +  2.57 cusr  3.9
 Result: PASS
 ```
 
-Besides the things covered by pjdfstest, JuiceFS provides:
+Aside from the POSIX features covered by pjdfstest, JuiceFS also provides:
 
-- Close-to-open consistency. Once a file is closed, the following open and read are guaranteed see the data written before close. Within same mount point, read can see all data written before it immediately.
-- Rename and all other metadata operations are atomic guaranteed by Redis transaction.
+- Close-to-open consistency. After the file is closed,  the data in the file can still be accessed for read and write in the future.  When using the same mount point, all the data that has been written can be immitiately seen and read.
```
davies, line 108: “After the file is closed, the data in the file can still be accessed for read and write in the future” is different from original meaning

davies (approved):

PR #1491, opened 2022-02-28
cmd: print usage if not enough args

SandyXSD:


cmd/warmup.go
```diff
@@ -102,6 +102,7 @@ func sendCommand(cf *os.File, batch []string, count int, threads uint, backgroun
 }
 
 func warmup(ctx *cli.Context) error {
+	setup(ctx, 0)
```
davies, line 105: 1

SandyXSD: If warmup with `-f` option, the positional argument is not necessary.

davies: lgtm

PR #1515, opened 2022-03-04
cmd/dump: add warning message about the sensitive information

SandyXSD:


pkg/meta/redis.go
```diff
@@ -2859,6 +2859,7 @@ func (m *redisMeta) DumpMeta(w io.Writer, root Ino) (err error) {
 		Sustained: sessions,
 		DelFiles:  dels,
 	}
+	dm.Setting.RemoveSecret()
```
davies, line 2862: Should we update the log message and docs to reflect this?

PR #1520, opened 2022-03-07
Beta2 testagain

sanwan:


sanwan:
select user,host,plugin from user;

update mysql.user set plugin = 'mysql_native_password';

ALTER USER 'root'@'localhost' IDENTIFIED WITH mysql_native_password BY 'root';

PR #1530, opened 2022-03-07
Add genpkey command

acsfer:
As `genrsa` has been [deprecated](https://www.openssl.org/docs/manmaster/man1/openssl-genrsa.html) on OpenSSL 3.0

Also, describe the static encryption method will be stored on the Metadata service _(depending which one is used, can be other than Redis)_.


davies:
@acsfer Thanks！

PR #1531, opened 2022-03-07
Not only Redis is supported nor mandatory

acsfer:
As title states


zhijian-pro:
We do this on purpose. `Redis` is an application that most developers are familiar with. We hope that new users can quickly understand the basic architecture of juicefs through redis + object storage.

acsfer:
Ok, feel free to close then 👍

Issue #894, opened 2021-10-18
switch from dgrijalva/jwt-goto to golang-jwt/jwt 


davies:
There is a security issue from dgrijalva/jwt-go, which is not maintained, we should switch to community maintained golang-jwt/jwt

PR #1534, opened 2022-03-08
fix jwt replace

zhijian-pro:


zhijian-pro:
Set the replace target version to `v3.2.1+incompatible`,
Executing `go mod tidy` will not cause errors like `used for two different module paths`
This way of replacing once is in line with expectations

zhijian-pro:
#894

PR #1543, opened 2022-03-09
access log in mount mode support writing to TCP port

### LINKED ISSUE #241: Redirect access log to a file or TCP port
In order to collect access log in K8S, we could export it to a TCP port

https://docs.fluentbit.io/manual/pipeline/inputs/tcp

https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-input-tcp.html

zhijian-pro:
Close #241


zhijian-pro:
@sanwan  Please check our `wintest`, the probability of failure in the last two days is too high, and it is not due to code changes.
image
image

davies:
The access log could be huge, is usually used to diagnose in realtime rather than analyzing, so we decide to not do it for now.

PR #1548, opened 2022-03-10
cmd/format: create a uuid object in the target bucket

SandyXSD:


cmd/format.go
```diff
@@ -396,6 +396,9 @@ func format(c *cli.Context) error {
 			} else {
 				logger.Warnf("List storage %s failed: %s", blob, err)
 			}
+			if err = blob.Put("juicefs_uuid", strings.NewReader(format.UUID)); err != nil {
```
davies, line 399: Make sure it's deleted with destroy command.

SandyXSD: it will, as destroy cleans up the whole blob, including testing, chunk, meta and this new object.

PR #1552, opened 2022-03-10
meta: check errors in doCleanStaleSession, and distinguish between close/clean

SandyXSD:
1. don't ignore errors, instead log them
2. continue cleaning when error occurs, but mark it failed
3. keep the session entry if the procedure is marked as failed

### DIFF
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -48,7 +48,7 @@ type engine interface {
 	doNewSession(sinfo []byte) error
 	doRefreshSession()
 	doFindStaleSessions(ts int64, limit int) ([]uint64, error) // limit < 0 means all
-	doCleanStaleSession(sid uint64)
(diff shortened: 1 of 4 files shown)


pkg/meta/base.go
```diff
@@ -222,7 +222,7 @@ func (m *baseMeta) CleanStaleSessions() {
 		return
 	}
 	for _, sid := range sids {
-		m.en.doCleanStaleSession(sid)
+		logger.Infof("clean up stale session %d: %s", sid, m.en.doCleanStaleSession(sid))
```
davies, line 225: only log the error?

SandyXSD: We need it to find out the reason why a session is removed. Besides, this log won't be many because client calls CloseSession to gracefully exit.

PR #1555, opened 2022-03-10
fix security issues

zhijian-pro:


zhijian-pro:
fix https://github.com/juicedata/juicefs/security/code-scanning

pkg/utils/utils.go
```diff
@@ -99,3 +99,7 @@ func GuessMimeType(key string) string {
 	}
 	return mimeType
 }
+
+func EncodeUserInput(input string) string {
```
davies, line 103: RemoveNewLine().

It seems that this is only used by pkg/fs, we should put it there.

pkg/fs/http.go
```diff
@@ -127,7 +128,7 @@ func (hfs *webdavFS) Rename(ctx context.Context, oldName, newName string) error
 }
 
 func (hfs *webdavFS) Stat(ctx context.Context, name string) (os.FileInfo, error) {
-	fi, err := hfs.fs.Stat(hfs.ctx, name)
+	fi, err := hfs.fs.Stat(hfs.ctx, utils.EncodeUserInput(name))
```
SandyXSD, line 131: Is it necessary? If yes, should we also apply it on other methods?

SandyXSD: @zhijian-pro no alert for this line

zhijian-pro: image
Here is the source of several other alerts.

SandyXSD (approved):

PR #1586, opened 2022-03-15
sdk support push to graphite

zhijian-pro:


sdk/java/libjfs/bridge.go
```diff
@@ -0,0 +1,334 @@
...
+	vec, err := expfmt.ExtractSamples(&expfmt.DecodeOptions{
+		Timestamp: now,
+	}, mfs...)
+	if err != nil {
+		return err
+	}
+
+	buf := bufio.NewWriter(w)
+	for _, s := range vec {
+		if prefix != "" {
```
zhijian-pro, line 219: The original logic here is to still write a point when the prefix is empty. I think this is a bug, so I modified it here. When prefix is empty, dots should not be written
```suggestion
		if prefix != "" {
```

xiaogaozi (approved):

sdk/java/libjfs/main.go
```diff
@@ -306,6 +308,70 @@ func createStorage(format meta.Format) (object.ObjectStorage, error) {
...
+			UseTags:       true,
+			Timeout:       2 * time.Second,
+			ErrorHandling: ContinueOnError,
+			Logger:        logger,
+			CommonLabels:  commonLabels,
+		})
+		if err != nil {
+			logger.Warnf("NewBridge error:%s", err)
+		}
+		bridges = append(bridges, bridge)
```
davies, line 325: bridge could be nil

sdk/java/libjfs/main.go
```diff
@@ -306,6 +308,70 @@ func createStorage(format meta.Format) (object.ObjectStorage, error) {
 	return object.WithPrefix(blob, format.Name+"/"), nil
 }
 
+func pushMetrics(graphite, pushGatewayAddr, pushAuth string, graphiteInterVal, pushGatewayInterVal int, registry *prometheus.Registry, commonLabels map[string]string) {
```
davies, line 311: it's better split it into two function, one for push gateway, one for graphite

sdk/java/libjfs/main.go
```diff
@@ -306,6 +308,70 @@ func createStorage(format meta.Format) (object.ObjectStorage, error) {
...
+		if pushAuth != "" {
+			if strings.Contains(pushAuth, ":") {
+				parts := strings.Split(pushAuth, ":")
+				pusher.BasicAuth(parts[0], parts[1])
+			}
+		}
+		pushers = append(pushers, pusher)
+	}
+
+	once.Do(func() {
```
davies, line 342: We could have one goroutine  for each pusher

davies: Or one for all pushers and bridges

sdk/java/libjfs/bridge.go
zhijian-pro: This is indeed a bug, upstream has been merged.https://github.com/prometheus/client_golang/pull/1005

zhijian-pro:
**graphite:**
image

**pushgateway:**
image

PR #1591, opened 2022-03-17
meta: add Heartbeat configuration

SandyXSD:
Old sessions can only be cleaned up by old clients.

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -151,7 +151,7 @@ func clientFlags() []cli.Flag {
 		&cli.DurationFlag{
 			Name:  "heartbeat",
 			Value: 12 * time.Second,
-			Usage: "interval to refresh this client; it's recommended that all clients use the same heartbeat value",
+			Usage: "interval to send heartbeat; it's recommended that all clients use the same heartbeat value",
 		},
 		&cli.BoolFlag{
 			Name:  "read-only",
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -148,7 +148,11 @@ func clientFlags() []cli.Flag {
 			Value: time.Hour,
 			Usage: "interval to automatically backup metadata in the object storage (0 means disable backup)",
 		},
-
+		&cli.DurationFlag{
+			Name:  "heartbeat",
+			Value: time.Minute,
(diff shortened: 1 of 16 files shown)


pkg/meta/base.go
```diff
@@ -216,7 +216,7 @@ func (m *baseMeta) refreshSession() {
 }
 
 func (m *baseMeta) CleanStaleSessions() {
-	sids, err := m.en.doFindStaleSessions(time.Now().Add(time.Minute*-5).Unix(), 1000)
```
davies, line 219: There will be a problem when a client use 60s as heartbeat, but another one use 10s, then the second one will remove the first one as stale session.

cmd/flags.go
```diff
@@ -148,7 +148,11 @@ func clientFlags() []cli.Flag {
 			Value: time.Hour,
 			Usage: "interval to automatically backup metadata in the object storage (0 means disable backup)",
 		},
-
+		&cli.DurationFlag{
+			Name:  "heartbeat",
+			Value: time.Minute,
```
davies, line 153: Should we change it to 10 seconds? There will be 10 qps for 1000 clients, kind of reasonable.

pkg/meta/base.go
SandyXSD: Yes, so users must use the same heartbeat value for all sessions.
A maybe better alternative is to set the expire time rather than the current time stamp, but it's not compatible with old clients.

cmd/flags.go
SandyXSD: sure

sdk/java/src/main/java/io/juicefs/JuiceFileSystemImpl.java
```diff
@@ -305,6 +305,7 @@ public void initialize(URI uri, Configuration conf) throws IOException {
     obj.put("cacheSize", Integer.valueOf(getConf(conf, "cache-size", "100")));
     obj.put("openCache", Float.valueOf(getConf(conf, "open-cache", "0.0")));
     obj.put("backupMeta", Integer.valueOf(getConf(conf, "backup-meta", "3600")));
+    obj.put("heartbeat", Integer.valueOf(getConf(conf, "heartbeat", "60")));
```
davies, line 308: 12

PR #1610, opened 2022-03-18
delte local cache even max-deletes=0

lipengchengcs:
When we set `max-delete=0`,  Juicefs will not start any thread to delete the block, even if the block in local. However, `max-delete` should control the number of delete requests to the object storage. Deleting the data in the local cache will not affect it.


davies:
@lipengchengcs Thanks, #1611 have took over this one.

PR #1611, opened 2022-03-18
chunk: move MaxDeletes from meta to chunk package

SandyXSD:
Take over  #1610

### DIFF
--- a/pkg/chunk/cached_store.go
+++ b/pkg/chunk/cached_store.go
@@ -294,7 +294,6 @@ func (c *rChunk) Remove() error {
 				err = e
 			}
 		}
-		c.store.bcache.remove(key)
 	}
 	return err
 }
--- a/pkg/chunk/cached_store.go
+++ b/pkg/chunk/cached_store.go
@@ -292,7 +292,7 @@ func (c *rChunk) Remove() error {
 	}
 
 	if c.store.conf.MaxDeletes == 0 {
-		return nil
+		return errors.New("skip deleting objects because MaxDeletes is 0")
 	}
 	var err error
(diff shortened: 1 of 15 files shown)


pkg/chunk/cached_store.go
```diff
@@ -289,10 +288,20 @@ func (c *rChunk) Remove() error {
 		delete(c.store.pendingKeys, key)
 		c.store.pendingMutex.Unlock()
 		c.store.bcache.remove(key)
+
+	}
+
+	if c.store.conf.MaxDeletes == 0 {
+		return nil
```
davies, line 295: nil means the chunk is removed from object store, then slice will be remove from meta engine, and chunk will be leaked.

We should return EBUSY here, then no error message in deleteSlice()

pkg/chunk/cached_store.go
```diff
@@ -289,10 +288,20 @@ func (c *rChunk) Remove() error {
...
+	}
+	var err error
+	c.store.currentDelete <- struct{}{}
```
davies, line 298: There could be too many goroutines pending for deletions.

davies: btw, Same problem as before this change

SandyXSD: should fix in another PR

pkg/chunk/cached_store.go
SandyXSD: [67a5258](https://github.com/juicedata/juicefs/pull/1611/commits/67a52581a264edb067b4596c97e22909c35e5648)
Not using EBUSY here since that's not accurate enough.

PR #1619, opened 2022-03-21
sync command  support --links

### LINKED ISSUE #1588: Support `--links ` for sync command
From: https://github.com/juicedata/juicesync/issues/121

zhijian-pro:
Close #1588

### DIFF
--- a/pkg/object/file.go
+++ b/pkg/object/file.go
@@ -83,10 +83,10 @@ func (d *filestore) Head(key string) (Object, error) {
 		size = 0
 	}
 	return &obj{
-		key:   key,
-		size:  size,
-		mtime: fi.ModTime(),
-		isDir: fi.IsDir(),
+		key,
+		size,
+		fi.ModTime(),
+		fi.IsDir(),
 	}, nil
 }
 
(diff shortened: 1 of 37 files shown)


pkg/object/s3.go
```diff
@@ -86,10 +86,10 @@ func (s *s3client) Head(key string) (Object, error) {
 		return nil, err
 	}
 	return &obj{
-		key,
```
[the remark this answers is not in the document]
zhijian-pro: ok

pkg/object/file.go
```diff
@@ -326,6 +324,7 @@ func (d *filestore) ListAll(prefix, marker string) (<-chan Object, error) {
 				owner,
 				group,
 				info.Mode(),
+				info.Mode()&os.ModeSymlink != 0,
```
davies, line 327: In order to follow symlink, we should return the info of it's target (may also ignore the broken symlink?)

pkg/object/file.go
```diff
@@ -281,24 +315,19 @@ func (d *filestore) ListAll(prefix, marker string) (<-chan Object, error) {
...
-					return nil
-				}
 				if os.IsNotExist(err) {
 					logger.Warnf("skip not exist file or directory: %s", path)
 					return nil
 				}
 				listed <- nil
 				logger.Errorf("list %s: %s", path, err)
-				return err
+				return nil
```
sjp00556, line 330: why not return err?

sjp00556: if return nil , sync producer will quit loop,  so here need return err to quit recursive walk

PR #1623, opened 2022-03-22
object: fallback to List only if ListAll is not supported

### LINKED ISSUE #1614: listAll prefix meta/: not supported
### Discussed in https://github.com/juicedata/juicefs/discussions/1577

<sup>Originally posted by **solracsf** March 14, 2022</sup>
I'm getting these with **Scaleway** and **JuiceFS 1.0.0 beta2**; does this mean that the provider doesn't support these methods or is something wrong here?

```
<ERROR>: Can't list s3://jfsdata-shard1/: RequestError: send request failed
caused by: Get "https://s3.fr-par.scw.cloud/jfsdata-shard1?marker=&max-keys=10000&prefix=": net/http: timeout awaiting response headers [sharding.go:82]
<ERROR>: Can't list shard16://s3://jfsdata-shard0/jfs/(encrypted)meta/: not supported [sync.go:116]
<WARNING>: listAll prefix meta/: not supported [backup.go:100]
```

Thanks.</div>

SandyXSD:
fix #1614

### DIFF
--- a/pkg/object/object_storage.go
+++ b/pkg/object/object_storage.go
@@ -18,7 +18,6 @@ package object
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"sync"
@@ -83,7 +82,7 @@ type FileSystem interface {
 	Chown(path string, owner, group string) error
 }
 
-var notSupported = errors.New("not supported")
+var notSupported = utils.ENOTSUP
 
 type DefaultObjectStorage struct{}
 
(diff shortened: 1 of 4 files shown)


pkg/object/object_storage.go
```diff
@@ -83,42 +82,40 @@ type FileSystem interface {
 	Chown(path string, owner, group string) error
 }
 
-var notSupported = errors.New("not supported")
```
davies, line 86: var notSupported = utils.ENOTSUP

PR #1625, opened 2022-03-22
chunk: limit number of upload goroutines

SandyXSD:


pkg/chunk/cached_store.go
```diff
@@ -863,81 +851,88 @@ func parseObjOrigSize(key string) int {
 
 func (store *cachedStore) uploadStagingFile(key string, stagingPath string) {
```
SandyXSD, line 852: Within this function, only delete the `go func()`

PR #1644, opened 2022-03-24
sync: reduce memory allocation when write into files

201341:

### DIFF
--- a/pkg/object/object_storage.go
+++ b/pkg/object/object_storage.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"sync"
 	"time"
@@ -40,6 +41,10 @@ type File interface {
 	Owner() string
 	Group() string
 	Mode() os.FileMode
+}
+
+type onlyWriter struct {
+	io.Writer
 }
 
 type file struct {
--- a/pkg/object/file.go
+++ b/pkg/object/file.go
@@ -135,7 +135,7 @@ func (d *filestore) Put(key string, in io.Reader) error {
 	}()
 	buf := bufPool.Get().(*[]byte)
 	defer bufPool.Put(buf)
-	_, err = io.CopyBuffer(f, in, *buf)
+	_, err = io.CopyBuffer(onlyWriter{f}, in, *buf)
 	if err != nil {
 		_ = f.Close()
 		return err


SandyXSD:
@201341 Could you provide some detailed explanation?

201341:
> @201341 Could you provide some detailed explanation?

For file2file, io.CopyBuffer -> readfrom->io.Copy, it doesn't use the bufpool.

SandyXSD:
*os.File has the method `ReadFrom`, why do you want to convert its type to WriteCloser?

201341:
> *os.File has the method `ReadFrom`, why do you want to convert its type to WriteCloser?

In order to use the buffer, not to alloc the memory, we need to convert to WriteCloser.

davies:
@201341 io.CopyBuffer() will use copy_file_range (zero copy or in kernel) in linux, and fallback to io.Copy() in other OS, where do you saw the allocation?

201341:
> @201341 io.CopyBuffer() will use copy_file_range (zero copy or in kernel) in linux, and fallback to io.Copy() in other OS, where do you saw the allocation?

From pprof, `io.CopyBuffer->io.Copy()`  takes up most of the memory, And I see the code `io.Copy`  allocate the new memory!

davies:
What's the OS and version you are using?

201341:
> What's the OS and version you are using?

The version is Centos7(kernel 3.x), And it seems not support the `copy_file_range`.

SandyXSD:
This change makes sense since `ReadFrom` of File requires the `src` to be a *File as well to use `copy_file_range`, while in JuiceFS usually the src is bytes.Reader (except for sync/backup).

```
src, ok := r.(*File)
if !ok {
    return 0, false, nil
}
```

davies:
I see, the pool is not used when the `in` is not a `File`, or copy_file_range is not available.

We should check the difference when src and dst are disk and copy_file_range is available.

SandyXSD:
@201341 It looks like this commit won't change the behavior at all, cuz `io.WriteCloser(f)` still contains the `ReadFrom` method. Could you double check to confirm if it fixes the issue of `io.Copy`?

201341:
@SandyXSD Yes, it doesn't change the behavior. I have changed the code. Now it can use the buffer, But I don't know how to check the copy_file_range

SandyXSD:
Cool, and we need a method to check kernel version. After a quick look at the issue you linked, it looks that the [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1803162) is only NFS related?

pkg/object/object_storage.go
```diff
@@ -42,6 +43,10 @@ type File interface {
 	Mode() os.FileMode
 }
 
+type OnlyWriter struct {
```
davies, line 46: lower case

SandyXSD (approved):

pkg/object/object_storage.go
201341: fix it.

davies (approved):

davies:
@201341 Thanks! This patch disable the copy_file_range when both source and destination are same file system, we will find a way to enable it when it's safe.

penglaixy99:
From golang official webpage:   https://pkg.go.dev/io#CopyBuffer
CopyBuffer is identical to Copy except that it stages through the provided buffer (if one is required) rather than allocating a temporary one. If buf is nil, one is allocated; otherwise if it has zero length, CopyBuffer panics.

If either src implements WriterTo or dst implements ReaderFrom, buf will not be used to perform the copy.

PR #1646, opened 2022-03-24
chunk/store: limit max retry for async upload as well

SandyXSD:


SandyXSD:
Use a simpler modification: #1673

PR #1657, opened 2022-03-27
format: add `--hash-prefix` to add a hashed prefix for objects

SandyXSD:


cmd/format.go
```diff
@@ -142,6 +142,10 @@ Details: https://juicefs.com/docs/community/quick_start_guide`,
 				Value: 1,
 				Usage: "number of days after which removed files will be permanently deleted",
 			},
+			&cli.BoolFlag{
+				Name:  "hash-object-prefix",
```
davies, line 146: hash-prefix is better

PR #1670, opened 2022-03-29
cmd/sync: try `copy_file_range` only when it's probably available

SandyXSD:


pkg/object/file.go
```diff
@@ -150,9 +152,14 @@ func (d *filestore) Put(key string, in io.Reader) error {
 			_ = os.Remove(tmp)
 		}
 	}()
-	buf := bufPool.Get().(*[]byte)
-	defer bufPool.Put(buf)
-	_, err = io.CopyBuffer(onlyWriter{f}, in, *buf)
+
```
qingyunha, line 155: Can we just use `io.CopyBuffer(f, in, *buf)` here? golang has already test copy_file_range support.  https://github.com/golang/go/blob/2e773a3894fba7af744090d7d42968f4993018e2/src/internal/poll/copy_file_range_linux.go#L13

davies: copy_file_range only works if source and target are in the same file system, this is not true i most cases.

qingyunha: What's the problem with levaing it for golang?  it's for performance?

davies: it will fallback to io.Copy() and has lots of memory allocations.

qingyunha: io.Copy and io.CopyBuffer flow the same path to test copy_file_range
https://github.com/golang/go/blob/master/src/io/io.go#L411

qingyunha: https://github.com/golang/go/blob/2e773a3894fba7af744090d7d42968f4993018e2/src/os/readfrom_linux.go#L14

qingyunha: `io.Copy` and `io.CopyBuffer` flow the same path to test copy_file_range.
https://github.com/golang/go/blob/2e773a3894fba7af744090d7d42968f4993018e2/src/io/io.go#L411
https://github.com/golang/go/blob/2e773a3894fba7af744090d7d42968f4993018e2/src/os/readfrom_linux.go#L14

SandyXSD: The priority we want is `copy_file_range` > `io.CopyBuffer` > `io.Copy`. If we use `CopyBuffer` directly, and when `copy_file_range` is not supported, it will fall back to `io.Copy`, see: https://github.com/golang/go/blob/2e773a3894fba7af744090d7d42968f4993018e2/src/os/file.go#L155.
Thus, we have to actively remove the `ReadFrom` method of *os.File if `copy_file_range` is not supported.

qingyunha: ok, i see

PR #1675, opened 2022-03-30
Fix read disk_cache not existed remove cache key

lidaohang:


pkg/chunk/disk_cache.go
```diff
@@ -257,6 +257,10 @@ func (cache *cacheStore) load(key string) (ReadCloser, error) {
 			// update atime
 			cache.keys[key] = cacheItem{it.size, uint32(time.Now().Unix())}
 		}
+	} else {
+                cache.Unlock()
+		cache.remove(key)
```
davies, line 262: Since cache is locked, it's better to call `delete(cache.keys, key)`

PR #1687, opened 2022-03-30
meta: forbid empty name for dentry

SandyXSD:


pkg/meta/base.go
```diff
@@ -591,6 +591,9 @@ func (m *baseMeta) Mknod(ctx Context, parent Ino, name string, _type uint8, mode
 	if m.conf.ReadOnly {
 		return syscall.EROFS
 	}
+	if name == "" {
+		return syscall.ENOENT
```
davies, line 595: EINVAL

pkg/meta/sql.go
```diff
@@ -1589,6 +1589,10 @@ func (m *dbMeta) doReaddir(ctx Context, inode Ino, plus uint8, entries *[]*Entry
 		return errno(err)
 	}
 	for _, n := range nodes {
+		if n.Name == "" {
+			logger.Errorf("Corrupt entry with empty name: inode %d parent %d", n.Inode, inode)
```
davies, line 1593: should we generate a random name for it?

SandyXSD: Even if we do so, this entry can't be accessed, because lookup/unlink will return ENOENT.

pkg/meta/base.go
SandyXSD: It's what returned by glibc creat/rename functions. Also os.Create/os.Rename return ENOENT as well.

PR #1690, opened 2022-03-31
init version compatible tesing workflow

zhoucheng361:


sanwan:
The problem of failed travis(https://app.travis-ci.com/github/juicedata/juicefs/builds/248705624) may need to resovled.

PR #1696, opened 2022-03-31
support redis cluster

davies:
Currently, we don't support Redis cluster because it does not support transaction across slots.

By putting all the keys in single slot, we can use Redis cluster as the meta engine. we can also has many volumes in single cluster, by have them in different slots.

Amazon MemoryDB is a managed Redis service with cluster mode by default, we can use it as the meta engine with this PR.


pkg/meta/redis.go
```diff
@@ -679,10 +751,15 @@ func (r *redisMeta) txn(ctx Context, txf func(tx *redis.Tx) error, keys ...strin
...
+		switch c := r.rdb.(type) {
+		case *redis.Client:
+			err = c.Watch(ctx, txf, keys...)
+		case *redis.ClusterClient:
+			err = c.Watch(ctx, txf, keys...)
+		}
 		if r.shouldRetry(err, retryOnFailture) {
 			txRestart.Add(1)
-			logger.Debugf("Transaction failed, restart it (tried %d): %s", i+1, err)
+			logger.Errorf("Transaction failed, restart it (tried %d): %s", i+1, err)
```
SandyXSD, line 762: Change back or use `Warnf`

pkg/meta/redis.go
```diff
@@ -2387,79 +2436,87 @@ func (r *redisMeta) CompactAll(ctx Context, bar *utils.Bar) syscall.Errno {
...
 				}
 			}
 		}
-		if cursor == 0 {
-			break
+		return nil
+	})
+}
+
+func (r *redisMeta) scan(ctx context.Context, pattern string, f func([]string) error) syscall.Errno {
```
SandyXSD, line 2490: Returns an error makes the life easier.

pkg/meta/redis.go
```diff
@@ -731,12 +808,12 @@ func (r *redisMeta) Truncate(ctx Context, inode Ino, flags uint8, length uint64,
...
 			var keys []string
 			for {
-				keys, cursor, err = tx.Scan(ctx, cursor, fmt.Sprintf("c%d_*", inode), 10000).Result()
+				keys, cursor, err = tx.Scan(ctx, cursor, fmt.Sprintf("%sc%d_*", r.prefix, inode), 10000).Result()
 				if err != nil {
 					return err
 				}
 				for _, key := range keys {
-					indx, err := strconv.Atoi(strings.Split(key, "_")[1])
+					indx, err := strconv.Atoi(strings.Split(key[len(r.prefix)+1:], "_")[1])
```
SandyXSD, line 816: `key[len(r.prefix):]`, no need to `+1`

pkg/meta/redis.go
```diff
@@ -120,45 +133,64 @@ func newRedisMeta(driver, addr string, conf *Config) (Meta, error) {
...
+		fopt.Username = opt.Username
+		fopt.Password = opt.Password
 		fopt.TLSConfig = opt.TLSConfig
-		fopt.MaxRetries = conf.Retries
-		fopt.MinRetryBackoff = query.duration("min-retry-backoff", time.Millisecond*20)
-		fopt.MaxRetryBackoff = query.duration("max-retry-backoff", time.Second*10)
-		fopt.ReadTimeout = query.duration("read-timeout", time.Second*30)
-		fopt.WriteTimeout = query.duration("write-timeout", time.Second*5)
+		fopt.MaxRetries = opt.MaxRetries
+		fopt.MinRetryBackoff = opt.MinRetryBackoff
```
SandyXSD, line 142: Don't we need to adjust the default value of these options?

pkg/meta/redis.go
```diff
@@ -120,45 +133,64 @@ func newRedisMeta(driver, addr string, conf *Config) (Meta, error) {
...
+			copt.MaxRedirects = 1
+			copt.Username = opt.Username
+			copt.Password = opt.Password
+			copt.TLSConfig = opt.TLSConfig
+			copt.MaxRetries = opt.MaxRetries
+			copt.MinRetryBackoff = opt.MinRetryBackoff
+			copt.MaxRetryBackoff = opt.MaxRetryBackoff
+			copt.ReadTimeout = opt.ReadTimeout
+			copt.WriteTimeout = opt.WriteTimeout
+			if conf.ReadOnly {
```
SandyXSD, line 173: It's probably not required for cluster mode.

davies: the replication is still asynchronous

pkg/meta/redis_test.go
```diff
@@ -33,17 +33,22 @@ import (
...
 	m, err := newRedisMeta("redis", "127.0.0.1:6379/10", &conf)
 	if err != nil || m.Name() != "redis" {
 		t.Fatalf("create meta: %s", err)
 	}
 	testMeta(t, m)
 }
 
+func TestRedisCluster(t *testing.T) {
+	var conf = Config{}
+	m, err := newRedisMeta("redis", "127.0.0.1:7001,127.0.0.1:7002,127.0.0.1:7002/2", &conf)
```
SandyXSD, line 45: port 7003 for the last one

pkg/meta/redis.go
```diff
@@ -114,6 +114,18 @@ func newRedisMeta(driver, addr string, conf *Config) (Meta, error) {
 		opt.Password = os.Getenv("META_PASSWORD")
 	}
 	opt.MaxRetries = conf.Retries
+	if opt.MinRetryBackoff == 0 {
+		opt.MinRetryBackoff = time.Millisecond * 20
```
SandyXSD, line 118: Currently the default values of two backoffs are different in sentinel mode & non-sentinel mode

SandyXSD: Also struct `queryMap` can be deleted if it's no longer used.

pkg/meta/redis.go
```diff
@@ -120,45 +145,64 @@ func newRedisMeta(driver, addr string, conf *Config) (Meta, error) {
...
-		if fopt.Password == "" {
-			fopt.Password = os.Getenv("META_PASSWORD")
-		}
 		fopt.SentinelPassword = os.Getenv("SENTINEL_PASSWORD")
 		fopt.DB = opt.DB
+		fopt.Username = opt.Username
+		fopt.Password = opt.Password
 		fopt.TLSConfig = opt.TLSConfig
-		fopt.MaxRetries = conf.Retries
-		fopt.MinRetryBackoff = query.duration("min-retry-backoff", time.Millisecond*20)
```
SandyXSD, line 135: Just recalled that we need this because we can't get them by `redis.ParseURL`

SandyXSD: https://github.com/go-redis/redis/blob/v8.4.0/options.go#L218-L220

pkg/meta/redis.go
```diff
@@ -102,8 +103,32 @@ func newRedisMeta(driver, addr string, conf *Config) (Meta, error) {
...
+		// redis.ParseURL does not support multiple hosts
+		opt.Addr, _, _ = net.SplitHostPort(opt.Addr)
+	}
+	if opt.Password == "" {
+		opt.Password = os.Getenv("REDIS_PASSWORD")
+	}
+	if opt.Password == "" {
+		opt.Password = os.Getenv("META_PASSWORD")
+	}
+	opt.MaxRetries = conf.Retries
```
SandyXSD, line 116: Should we set it only when opt.MaxRetries == 0?

SandyXSD (approved):

pkg/meta/redis.go
SandyXSD: Also `queryMap` can be deleted if it's no longer used.

PR #1724, opened 2022-04-06
fix parse rsa from pem

### LINKED ISSUE #1708: Flaky test: TestRSA
=== RUN   TestRSA
    encrypt_test.go:56: parse without passphrase should return IncorrectPasswordError
--- FAIL: TestRSA (1.14s)

zhijian-pro:
Close #1708


cmd/format.go
```diff
@@ -210,9 +210,12 @@ func createStorage(format meta.Format) (object.ObjectStorage, error) {
 
 	if format.EncryptKey != "" {
 		passphrase := os.Getenv("JFS_RSA_PASSPHRASE")
+		if passphrase == "" {
```
davies, line 213: passphrase could be '' if it's not encrypted.

PR #1726, opened 2022-04-06
cmd/format: encrypt secret key only if it's really updated

SandyXSD:


cmd/config.go
```diff
@@ -71,6 +71,10 @@ $ juicefs config redis://localhost --min-client-version 1.0.0 --max-client-versi
 				Name:  "secret-key",
 				Usage: "secret key for object storage",
 			},
+			&cli.BoolFlag{
+				Name:  "encrypt",
```
davies, line 75: encrypt-secret

cmd/config.go
```diff
@@ -126,6 +130,7 @@ func config(ctx *cli.Context) error {
 	}
 
 	var quota, storage, trash, clientVer bool
+	var encrypt bool
```
davies, line 133: encrypted

PR #1738, opened 2022-04-07
complete version compatible tests and config it as PR check workflow

zhoucheng361:


davies:
--encrypt should be --encrypt-secret

zhoucheng361:
> 

fixed.

.github/workflows/version_compatible.yml
```diff
@@ -1,13 +1,22 @@
 name: "version-compatible-test"
 
 on:
-  schedule:
-    - cron:  '2 0 * * *'
+  push:
+    branches: [ main ]
```
davies, line 5: also release-*

zhoucheng361: > also release-*
fixed

Issue #1567, opened 2022-03-13
[MariaDB] Error 1366: Incorrect string value


solracsf:
**What happened**:

While rsync from local disk to juicefs mount, it suddenly (after several hours) stopped with 

`Failed to sync with 11 errors: last error was: open /mnt/juicefs/folder/file.xls: input/output error`

On the jfs log, i can find these:

```
juicefs[187516] <ERROR>: error: Error 1366: Incorrect string value: '\xE9sa sa...' for column `jfsdata`.`jfs_edge`.`name` at row 1
goroutine 43510381 [running]:
runtime/debug.Stack()
        /usr/local/go/src/runtime/debug/stack.go:24 +0x65
github.com/juicedata/juicefs/pkg/meta.errno({0x2df8860, 0xc0251d34a0})
        /go/src/github.com/juicedata/juicefs/pkg/meta/utils.go:76 +0xc5
github.com/juicedata/juicefs/pkg/meta.(*dbMeta).doMknod(0xc0000e0c40, {0x7fca7e165300, 0xc00ed28040}, 0x399f6f, {0xc025268160, 0x1f}, 0x1, 0x1b4, 0x0, 0x0, ...)
        /go/src/github.com/juicedata/juicefs/pkg/meta/sql.go:1043 +0x29e
github.com/juicedata/juicefs/pkg/meta.(*baseMeta).Mknod(0xc0000e0c40, {0x7fca7e165300, 0xc00ed28040}, 0x399f6f, {0xc025268160, 0x1f}, 0xc0, 0x7b66, 0x7fca, 0x0, ...)
        /go/src/github.com/juicedata/juicefs/pkg/meta/base.go:594 +0x275
github.com/juicedata/juicefs/pkg/meta.(*baseMeta).Create(0xc0000e0c40, {0x7fca7e165300, 0xc00ed28040}, 0x26b5620, {0xc025268160, 0x2847500}, 0x8040, 0xed2, 0x8241, 0xc025267828, ...)
        /go/src/github.com/juicedata/juicefs/pkg/meta/base.go:601 +0x109
github.com/juicedata/juicefs/pkg/vfs.(*VFS).Create(0xc000140640, {0x2e90348, 0xc00ed28040}, 0x399f6f, {0xc025268160, 0x1f}, 0x81b4, 0x22a4, 0xc0)
        /go/src/github.com/juicedata/juicefs/pkg/vfs/vfs.go:357 +0x256
[...]

PR #1762, opened 2022-04-12
sql: changing name column type to binary type

### LINKED ISSUE #1684: PGsql ERROR: invalid byte sequence for encoding "UTF8"
On the server (postgres 14.2) log, multiple entries:

```
ERROR:  invalid byte sequence for encoding "UTF8": 0xe9 0x73 0x61
CONTEXT:  unnamed portal parameter $2
STATEMENT:  SELECT * FROM "jfs_edge" INNER JOIN "jfs_node" ON jfs_edge.inode=jfs_node.inode
            WHERE "jfs_edge"."parent"=$1 AND "jfs_edge"."name"=$2 LIMIT 1
```
Enconding params:
```
# sudo -u postgres psql
psql (14.2 (Ubuntu 14.2-1.pgdg20.04+1))
Type "help" for help.

postgres=# SHOW server_encoding;
 server_encoding
-----------------
 UTF8
(1 row)

postgres=# SHOW client_encoding;
 client_encoding
-----------------
 UTF8
(1 row)
```

Similar to https://github.com/juicedata/juicefs/issues/1567 (same dataset)

zhijian-pro:
This pr is mainly to solve the problem that juicefs cannot create files whose filenames are not UTF8 encoded when using the sql type of metadata engine
To solve this problem, the metadata database field type needs to be updated, so there is a certain compatibility problem with the old version. The following is a smooth upgrade scheme for upgrading the old version to this version:

1. If there are old clients in mount state, they can be upgraded without stopping them
2. Enter the metadata database and manually execute sql
```sql
alter table jfs_edge
    modify name varbinary(2048) not null;
alter table jfs_symlink
    modify target blob not null;
```

3. At this time, the old client in the mount state can already write files with non-utf8 file names, but it should be noted that after executing sql, the old version of the client cannot use the metadata engine to remount, That is to say, the client in the non-mounted state is incompatible with the metadata engine after executing sql, and a new version of the client is required to mount. 

Close #1684  #1567

### DIFF
--- a/pkg/meta/dump.go
+++ b/pkg/meta/dump.go
@@ -132,9 +132,9 @@ func parseHex(c byte) (byte, error) {
 	}
 }
 
-func unescape(s string) string {
+func unescape(s string) []byte {
 	if !strings.ContainsRune(s, '%') {
-		return s
+		return []byte(s)
 	}
 
 	p := []byte(s)
@@ -152,7 +152,7 @@ func unescape(s string) string {
 		p[n] = c
 		n++
 	}
-	return string(p[:n])
+	return p[:n]
 }
 
 func (de *DumpedEntry) writeJSON(bw *bufio.Writer, depth int) error {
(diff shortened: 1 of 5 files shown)


zhijian-pro:
Close #1684  #1567

PR #1777, opened 2022-04-13
implement listall api for redis 

### LINKED ISSUE #1764: Support List for Redis and Swift
When Redis and Swift is used as data store, the List() API is required for gc, fsck and destroy.

zhijian-pro:
Close #1764

### DIFF
--- a/pkg/object/redis.go
+++ b/pkg/object/redis.go
@@ -98,9 +98,13 @@ func (t *redisStore) ListAll(prefix, marker string) (<-chan Object, error) {
 	now := time.Now()
 	for _, key := range keyList {
 		data, err := t.rdb.Get(c, key).Bytes()
-		if err != nil && err != redis.Nil {
+		if err != nil {
+			if err == redis.Nil {
+				continue
+			}
 			return nil, err
 		}
+
 		// FIXME: mtime
 		objs <- &obj{key, int64(len(data)), now, strings.HasSuffix(key, "/")}
 	}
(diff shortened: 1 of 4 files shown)


pkg/object/redis.go
```diff
@@ -69,6 +71,52 @@ func (r *redisStore) Delete(key string) error {
...
+	if marker == "" {
+		marker = prefix
+	}
+	if limit > 1000 {
+		limit = 1000
+	}
+	var keyList []string
+	var cursor uint64
+	for {
+		keys, c, err := t.rdb.Scan(context.TODO(), cursor, marker+"*", limit).Result()
```
davies, line 84: marker means the first key, not prefix

davies: the results should be sorted by key, so we should get implement the ListAll(), get all the keys, sort them, then return them

pkg/object/swift.go
```diff
@@ -69,6 +68,31 @@ func (s *swiftOSS) Delete(key string) error {
 	return s.conn.ObjectDelete(s.container, key)
 }
 
+func (s *swiftOSS) List(prefix, marker string, limit int64) ([]Object, error) {
```
davies, line 71: please split these into another PR

pkg/object/redis.go
```diff
@@ -69,6 +72,51 @@ func (r *redisStore) Delete(key string) error {
...
+			break
+		}
+		cursor = c
+	}
+	sort.Strings(keyList)
+	now := time.Now()
+	for _, key := range keyList {
+		data, err := t.rdb.Get(c, key).Bytes()
+		if err != nil {
+			return nil, err
```
davies, line 102: a key could be removed, ignore the `redis.Nil`

davies: ignore the key, don't return a object with empty data

PR #1789, opened 2022-04-14
Add log to print SA email when create a GCS client

liufuyang:
So to help debug with issues when connecting with GCS
https://juicefs.slack.com/archives/C03400WJ2C8/p1649932508711669

And this allows me to test with different ways of google auth login:

So if I do `gcloud auth application-default login` but start juices directly, I got:
```
2022/04/14 14:22:31.439873 juicefs[10746] <INFO>: GCS client credentail client_email: "<nil>", type: "<nil>" [gs.go:188]
```
And this will only make allow me to read the buckets but not able to write (permission error)

And if I do `gcloud auth application-default login` but start juices with `export GOOGLE_APPLICATION_CREDENTIALS=/content/.config/application_default_credentials.json`
then it is like:
```
2022/04/14 14:25:16.601912 juicefs[10778] <INFO>: GCS client credentail client_email: "<nil>", type: "authorized_user" [gs.go:188]
```
This will allow me to write.

So adding this log I feel is quite helpful and also if some service account is used it also can print out the SA's email.

### DIFF
--- a/pkg/object/gs.go
+++ b/pkg/object/gs.go
@@ -182,16 +182,17 @@ func newGS(endpoint, accessKey, secretKey string) (ObjectStorage, error) {
 		return nil, err
 	}
 
+	// The block below only logs the GCS client's credential service account's email and type, added for providing extra useful logging information to end users.
 	credential, err := google.FindDefaultCredentials(ctx)
 	if err == nil {
 		credentialContent := map[string]interface{}{}
 		if err := json.Unmarshal(credential.JSON, &credentialContent); err == nil {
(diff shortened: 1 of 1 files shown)


davies:
@liufuyang Is this PR only used for debug, right?

davies:
Please reopen it if needed in future.

liufuyang:
> @liufuyang Is this PR only used for debug, right?
> 

Yes it is used for debug, but I feel it could be a helpful thing to always log the service account email when mounting with GCS ?

PR #1790, opened 2022-04-14
meta: delay deletion of compacted slices

SandyXSD:


pkg/meta/base.go
```diff
@@ -951,3 +973,18 @@ func (m *baseMeta) doCleanupTrash(force bool) {
...
 	}
 }
+
+func (m *baseMeta) cleanupDelayedSlices() {
+	now := time.Now()
+	edge := now.Unix() - int64(m.fmt.TrashDays)*24*3600
+	logger.Debugf("Cleanup delayed slices: started with edge %d", edge)
+	if count, err := m.en.doCleanupDelayedSlices(edge, 3e5); err == nil {
+		if count > 0 {
+			logger.Infof("Cleanup delayed slices: deleted %d slices in %v", count, time.Since(now))
```
davies, line 983: Debugf

pkg/meta/redis.go
```diff
@@ -2244,6 +2248,62 @@ func (r *redisMeta) doDeleteFileData_(inode Ino, length uint64, tracking string)
...
+			if err := r.txn(ctx, func(tx *redis.Tx) error {
+				_, e := tx.Pipelined(ctx, func(pipe redis.Pipeliner) error {
+					for i, s := range ss {
+						rs[i] = pipe.HIncrBy(ctx, r.sliceRefs(), r.sliceKey(s.Chunkid, s.Size), -1)
+					}
+					pipe.HDel(ctx, r.delSlices(), key)
+					return nil
+				})
+				return e
+			}, r.delSlices()); err != nil {
```
davies, line 2285: don't need to lock r.delSlices(), or we should use  tx.HGet(r.delSlices(), key) inside txn()

pkg/meta/base.go
```diff
@@ -951,3 +973,18 @@ func (m *baseMeta) doCleanupTrash(force bool) {
 		}
 	}
 }
+
+func (m *baseMeta) cleanupDelayedSlices() {
+	now := time.Now()
+	edge := now.Unix() - int64(m.fmt.TrashDays)*24*3600
+	logger.Debugf("Cleanup delayed slices: started with edge %d", edge)
+	if count, err := m.en.doCleanupDelayedSlices(edge, 3e5); err == nil {
```
davies, line 981: When there are more than 300K slices per hour, we can not cleanup them ASAP. Why do we need this limit?

pkg/meta/redis.go
davies: There is no conflict, so no key is needed.

pkg/meta/redis_test.go
```diff
@@ -898,8 +904,27 @@ func testCompaction(t *testing.T, m Meta) {
...
-	if deletes < 30 {
-		t.Fatalf("deleted chunks %d is less then 30", deletes)
+	if trash {
+		if deletes > 10 {
+			t.Fatalf("deleted chunks %d is greater than 10", deletes)
+		}
+		if len(slices[1]) < 200 {
+			t.Fatalf("list delayed slices %d is less than 200", len(slices[1]))
+		}
+		switch m := m.(type) {
```
davies, line 914: m.(engine)

pkg/meta/sql.go
```diff
@@ -2159,6 +2230,31 @@ func (m *dbMeta) ListSlices(ctx Context, slices map[Ino][]Slice, delete bool, sh
 			}
 		}
 	}
+
+	var ds delslices
+	rows2, err := m.db.Rows(&ds)
```
davies, line 2235: table delslices may not be created.

pkg/meta/tkv.go
```diff
@@ -1771,6 +1776,51 @@ func (m *kvMeta) doDeleteFileData(inode Ino, length uint64) {
...
+				rs[i] = tx.incrBy(m.sliceKey(s.Chunkid, s.Size), -1)
+			}
+			tx.dels([]byte(key))
+			return nil
+		}); err != nil {
+			logger.Warnf("Cleanup delayed slices %s: %s", key, err)
+			continue
+		}
+		for i, s := range ss {
+			if rs[i] < 0 {
```
davies, line 1812: <=

pkg/meta/sql.go
```diff
@@ -2159,6 +2230,31 @@ func (m *dbMeta) ListSlices(ctx Context, slices map[Ino][]Slice, delete bool, sh
...
+	if err != nil {
+		return errno(err)
+	}
+	for rows2.Next() {
+		if len(ds.Slices) > 0 {
+			ds.Slices = ds.Slices[:0]
+		}
+		if rows2.Scan(&ds) == nil {
```
davies, line 2243: reset ds.Slices

pkg/meta/redis.go
SandyXSD: At lease one key is required for Redis txn to use pessimistic lock

pkg/meta/base.go
SandyXSD: Same mechanism as cleaning up trash. But yes, we are lack of a way to delete the slices manually.

PR #1792, opened 2022-04-14
speed up redis listall interface

zhijian-pro:


pkg/object/filesystem_test.go
```diff
@@ -64,6 +64,11 @@ func TestHDFS2(t *testing.T) {
 	testFileSystem(t, dfs)
 }
 
+func TestRedis2(t *testing.T) {
```
davies, line 67: Redis don't have file system sematics, don't need to test it.

pkg/object/redis.go
```diff
@@ -79,13 +79,13 @@ func (t *redisStore) ListAll(prefix, marker string) (<-chan Object, error) {
...
 	for {
 		// FIXME: this will be really slow for many objects
-		keys, c, err := t.rdb.Scan(context.TODO(), cursor, "*", 1000).Result()
+		keys, c, err := t.rdb.Scan(context.TODO(), cursor, prefix+"*", 1000).Result()
 		if err != nil {
 			logger.Warnf("redis scan error, coursor %d: %s", cursor, err)
 			return nil, err
 		}
 		for _, key := range keys {
-			if key > marker && strings.HasPrefix(key, prefix) {
```
davies, line 88: We can use Len and Pipeline to speed up it

pkg/object/redis.go
```diff
@@ -94,19 +94,30 @@ func (t *redisStore) ListAll(prefix, marker string) (<-chan Object, error) {
 		}
 		cursor = c
 	}
+
 	sort.Strings(keyList)
-	now := time.Now()
+	p := t.rdb.Pipeline()
```
davies, line 99: 一个 batch 的 数量应该有个限制

pkg/object/redis.go
```diff
@@ -94,19 +94,30 @@ func (t *redisStore) ListAll(prefix, marker string) (<-chan Object, error) {
...
 	for _, key := range keyList {
-		data, err := t.rdb.Get(c, key).Bytes()
-		if err != nil {
-			if err == redis.Nil {
-				continue
+		p.Get(c, key)
```
davies, line 101: 用 Len

pkg/utils/utils.go
```diff
@@ -108,3 +108,21 @@ func StringContains(s []string, e string) bool {
 	}
 	return false
 }
+
+func SliceSplitWithMapFunc(srcSlice []string, batch int, f func(subSlice []string) error) error {
```
davies, line 112: this one is not needed

pkg/object/redis.go
```diff
@@ -94,19 +95,45 @@ func (t *redisStore) ListAll(prefix, marker string) (<-chan Object, error) {
...
 			return nil, err
 		}
 
-		// FIXME: mtime
-		objs <- &obj{key, int64(len(data)), now, strings.HasSuffix(key, "/")}
+		now := time.Now()
+		for idx, cmd := range cmds {
+			if intCmd, ok := cmd.(*redis.IntCmd); ok {
+				size, err := intCmd.Result()
+				if err != nil {
```
davies, line 121: Will err be Nil?

zhijian-pro: err will not be nil but size will be 0 when key does not exist

PR #1802, opened 2022-04-15
add batch limit to pipeline

zhijian-pro:


pkg/meta/redis.go
```diff
@@ -3021,12 +3022,24 @@ func (m *redisMeta) loadEntry(e *DumpedEntry, cs *DumpedCounters, refs map[strin
 		}
 	} else if attr.Typ == TypeDirectory {
 		attr.Length = 4 << 10
-		if len(e.Entries) > 0 {
-			dentries := make(map[string]interface{})
-			for _, c := range e.Entries {
-				dentries[string(unescape(c.Name))] = m.packEntry(typeFromString(c.Attr.Type), c.Attr.Inode)
+		eLen := len(e.Entries)
```
davies, line 3025: dentries := make(map[string]interface{}, batch)
for k := range e.Entries {
   entry := e.Entries[k]
   dentries[string(unescape(k))] = m.packEntry(typeFromString(entry.Attr.Type), entry.Attr.Inode)
   if len(dentries) > 0 {
      p.HSet(ctx, m.entryKey(inode), dentries)
      dentries = make(map[string]interface{}, batch)
   }
}
if len(dentries) > 0 {
   p.HSet(ctx, m.entryKey(inode), dentries)
}

PR #1828, opened 2022-04-19
cmd: use logger to print errors

### LINKED ISSUE #1786: should display clear message when load with exist meta.
**What happened**:

root@bench-01:~# ./juicefs load sqlite3:///tmp/encoding-test3.db backup.dump 
2022/04/14 17:54:18.469497 juicefs[1008732] <INFO>: Meta address: sqlite3:///tmp/encoding-test3.db [interface.go:383]
2022-04-14 17:54:18.470668 I | Database sqlite3 is not empty

 Database sqlite3 is not empty  should be something like "load canceled because database sqlite3:///tmp/encodeing-test3.db is not empty, "
and the log level should be Warning instead of Info.

**Environment**:
- Kernel (e.g. `uname -a`):
- Others:

SandyXSD:
close #1786


cmd/bench.go
```diff
@@ -301,7 +301,7 @@ func bench(ctx *cli.Context) error {
 	setup(ctx, 1)
 	/* --- Pre-check --- */
 	if ctx.Uint("block-size") == 0 || ctx.Uint("threads") == 0 {
-		return os.ErrInvalid
```
davies, line 304: This easier way is to change main():
```
func main() {
	err := Main(os.Args)
	if err != nil {
		log.Fatal(err)
	}
}
```

SandyXSD: Logger prints the line number and is usually helpful to trace the error. Besides, it's good to use only one way (fatal or return) to handle errors in commands.

davies: Main() is also in cmd/ package, we can use logger there, right?

davies: I meant changing `log` to `logger` in main should be enough.

SandyXSD: OK

PR #1854, opened 2022-04-21
meta/sql: fix wrong result from scanning SQL database

davies:
There is a bug in xorm when scanning a binary column, it could return partial or wrong result.

This PR fix that by reset the binary column before scan next row.

### DIFF
--- a/pkg/meta/sql.go
+++ b/pkg/meta/sql.go
@@ -723,6 +723,15 @@ func (m *dbMeta) appendSlice(s *xorm.Session, inode Ino, indx uint32, buf []byte
 	return err
 }
 
+func scanChunk(rows *xorm.Rows, c *chunk) error {
+	// FIXME: there is a bug in xorm, which will fill existing slice with partial result
+	// see line 582 in session.go:582 of xorm
+	if len(c.Slices) > 0 {
+		c.Slices = c.Slices[:0]
+	}
+	return rows.Scan(c)
+}
+
 func (m *dbMeta) Truncate(ctx Context, inode Ino, flags uint8, length uint64, attr *Attr) syscall.Errno {
 	defer timeit(time.Now())
 	f := m.of.find(inode)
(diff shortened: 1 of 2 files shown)


davies:
Create a bug for xorm: https://gitea.com/xorm/xorm/issues/2134

pkg/meta/sql.go
```diff
@@ -1622,6 +1631,7 @@ func (m *dbMeta) doFindStaleSessions(limit int) ([]uint64, error) {
 	}
 	var sids []uint64
 	for rows.Next() {
+		s.Info = nil
```
SandyXSD, line 1634: no need since s.Info is not used

pkg/meta/sql.go
```diff
@@ -764,7 +773,7 @@ func (m *dbMeta) Truncate(ctx Context, inode Ino, flags uint8, length uint64, at
 				return err
 			}
 			for rows.Next() {
-				if err = rows.Scan(&c); err != nil {
+				if err = scanChunk(rows, &c); err != nil {
```
SandyXSD, line 776: no need since Slices is not used

pkg/meta/sql.go
```diff
@@ -1976,7 +1986,7 @@ func (m *dbMeta) doDeleteFileData(inode Ino, length uint64) {
 	}
 	var indexes []uint32
 	for rows.Next() {
-		if rows.Scan(&c) == nil {
+		if scanChunk(rows, &c) == nil {
```
SandyXSD, line 1989: no need

pkg/meta/sql.go
```diff
@@ -2188,6 +2198,7 @@ func (m *dbMeta) ListXattr(ctx Context, inode Ino, names *[]byte) syscall.Errno
 	defer rows.Close()
 	*names = nil
 	for rows.Next() {
+		x.Value = nil
```
SandyXSD, line 2201: no need

pkg/meta/sql.go
davies: It's better to use the same way

PR #1861, opened 2022-04-21
meta/tkv: support large transaction in badger and etcd

davies:


pkg/meta/tkv_badger.go
```diff
@@ -214,11 +214,13 @@ func (c *badgerClient) txn(f func(kvTxn) error) (err error) {
...
 	}()
-	err = f(&badgerTxn{tx, c.client})
+	txn := &badgerTxn{tx, c.client}
+	err = f(txn)
 	if err != nil {
 		return err
 	}
-	return tx.Commit()
+	// tx could be committed
+	return txn.t.Commit()
```
SandyXSD, line 223: The txn.t should the same as tx before, any real change here?

PR #1872, opened 2022-04-22
Load test with big dir using different meta engine

zhoucheng361:


davies:
@zhoucheng361 This action took 118m, it's too slow to run, even for release branch.

PR #1912, opened 2022-04-26
cli flag allow customized storage type

DEvmIb:
allow mount to select storage typ other than default from format.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ dist/
 libjfs.h
 .DS_Store
 docs/node_modules
+cmd/cmd
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -73,6 +73,10 @@ func clientFlags() []cli.Flag {
 	}
 	return []cli.Flag{
 		&cli.StringFlag{
+			Name:  "storage",
+			Usage: "customized storage type (e.g. s3, gcs, oss, cos) to access object store",
+		},
+		&cli.StringFlag{
 			Name:  "bucket",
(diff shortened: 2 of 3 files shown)


zhijian-pro:
@DEvmIb  What kind of scenarios require such behavior ?

DEvmIb:
main server using file for performance. clients in other location can not access files. the bucket switch is not enough, juicefs sets storage type from meta "file://https://example.com/gws/.juicefs/fs/gws/".

so why not let client choose where the files are and let juicefs become more cloud aware. juicefs has many storage types, but after setting up all clients forced to use just this one storage type.

davies:
For a single JuiceFS file system, all the clients of JuiceFS should able read the data written by others. If the storage types are different, how can they work together? 

We can't understand why this option is need.

DEvmIb:
ok let my try to explain it better. 

location A
- direct access to the chunks on disk.
- using mount --storage file --bucket /path/to/chunks redis://test.local/1
- no overhead going through s3,webdav etc

Location B
- no direct access
- using mount --storage webdav --bucket https://locationa/path/to/chunks redis://test.public/1

as i write above the storage is fixed at metata/settings and let me not mix how i want to access the chunk store.

it works for me, that should be enough to write, it should just be a hint :)

davies:
@DEvmIb That make sense, lgtm.

davies (approved):

davies:
@DEvmIb If there is no conflict with main branch, you don't need to merge with it to MERGE the PR.

PR #1938, opened 2022-04-29
CI：move some tests which run for a long time from travis to action

sanwan:


.github/workflows/longtime.yml
```diff
@@ -0,0 +1,93 @@
+name: "longtime"
+
+on:
+  push:
+    branches:
+      - 'release-**'
```
davies, line 6: This should run on main branch also

.github/workflows/longtime.yml
```diff
@@ -0,0 +1,93 @@
+name: "longtime"
```
davies, line 1: Can we have a better name?

for example, Integration tests

sanwan: ok

PR #1939, opened 2022-04-29
hadoop: rename libjfs atomic

tangyoupeng:


sdk/java/src/main/java/io/juicefs/JuiceFileSystemImpl.java
```diff
@@ -562,16 +564,7 @@ public static Libjfs loadLibrary() throws IOException {
...
-          if (tmp.renameTo(new File(dir, name))) {
-            // updated libjfs.so
-            libFile = new File(dir, name);
-          } else {
-            libFile.delete();
-            if (!tmp.renameTo(libFile)) {
-              throw new IOException("Can't update " + libFile);
-            }
-          }
+          Files.move(tmp.toPath(), libFile.toPath(), StandardCopyOption.ATOMIC_MOVE);
```
davies, line 567: we may don't have permission to write, so try File(dir, name) first, and fallback to libFile

PR #1961, opened 2022-05-06
meta/tkv: dump metadata using snapshot

davies:
Currently, we use transaction to scan all the metadata, which is heavy (bookkeeping) and easy to fail (timeout or others), this  PR change it to iterator on read-only snapshot (based on timestamp or revision).

### DIFF
--- a/pkg/meta/tkv.go
+++ b/pkg/meta/tkv.go
@@ -43,7 +43,6 @@ type kvTxn interface {
 	get(key []byte) []byte
 	gets(keys ...[]byte) [][]byte
 	scanRange(begin, end []byte) map[string][]byte
-	scan(prefix []byte, handler func(key, value []byte))
 	scanKeys(prefix []byte) [][]byte
 	scanValues(prefix []byte, limit int, filter func(k, v []byte) bool) map[string][]byte
 	exist(prefix []byte) bool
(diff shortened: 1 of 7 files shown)


pkg/meta/tkv.go
```diff
@@ -2280,26 +2280,26 @@ func (m *kvMeta) DumpMeta(w io.Writer, root Ino) (err error) {
...
-					guessKeyTotal += int64(math.Ceil((float64(used/inodeTotal/(64*1024*1024)) + float64(3)) * float64(inodeTotal)))
+			bUsed, _ := m.get(m.counterKey(usedSpace))
+			bInodes, _ := m.get(m.counterKey(totalInodes))
+			used := parseCounter(bUsed)
+			inodeTotal := parseCounter(bInodes)
+			var guessKeyTotal int64 = 3 // setting, nextInode, nextChunk
+			if inodeTotal > 0 {
+				guessKeyTotal += int64(math.Ceil((float64(used/inodeTotal/(64*1024*1024)) + float64(3)) * float64(inodeTotal)))
+			}
+			bar.SetCurrent(0) // Reset
```
SandyXSD, line 2291: no need to reset now since it won't retry

SandyXSD (approved):

PR #1971, opened 2022-05-07
Revert "meta/tkv: dump metadata using snapshot"

davies:
Reverts juicedata/juicefs#1961

Without creating a safepoint, the GC may remove the data before scan finished, which may end up with partial file system.


SandyXSD:
We should create an issue for this problem if it's difficult to solve for now.

davies:
The scan will fail if GC is newer than snapshot.

PR #1995, opened 2022-05-11
reload configuration for data storage

### LINKED ISSUE #1986: JuiceFS update object storage ak/sk need remount
juicefs config ... update object storage ak/sk need remount, very difficult when the business is large.

expected: hotfix

davies:
closes #1986

### DIFF
--- a/cmd/format.go
+++ b/cmd/format.go
@@ -180,7 +180,7 @@ func fixObjectSize(s int) int {
 	return s
 }
 
-func createStorage(format *meta.Format) (object.ObjectStorage, error) {
+func createStorage(format meta.Format) (object.ObjectStorage, error) {
 	if err := format.Decrypt(); err != nil {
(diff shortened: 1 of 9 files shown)


cmd/format.go
```diff
@@ -180,7 +180,7 @@ func fixObjectSize(s int) int {
 	return s
 }
 
-func CreateStorage(format meta.Format) (object.ObjectStorage, error) {
+func createStorage(format *meta.Format) (object.ObjectStorage, error) {
```
SandyXSD, line 183: we shouldn't pass pointer here, otherwise it will change the value of m.fmt, which will be used in DumpMeta.

davies: Only Decrypt() will change m.fmt, can DumpMeta work with decrypted Format?

davies: To be safe, I will revert the pointer here.

SandyXSD (approved):

PR #2030, opened 2022-05-17
feat: use iota assign to meta type

gaius-qi:
Signed-off-by: Gaius <gaius.qi@gmail.com>

### DIFF
--- a/pkg/meta/interface.go
+++ b/pkg/meta/interface.go
@@ -45,13 +45,13 @@ const (
 )
 
 const (
-	TypeFile      = 1 // type for regular file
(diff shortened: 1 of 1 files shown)


pkg/meta/interface.go
```diff
@@ -45,13 +45,13 @@ const (
 )
 
 const (
-	TypeFile      = 1 // type for regular file
-	TypeDirectory = 2 // type for directory
```
davies, line 49: The number is persisted into database, we should NOT change any of it, so we'd prefer to use the number here.

It's also more readable to people who is not familiar with go.

PR #2048, opened 2022-05-19
Using juicefs directly via SPI in Hadoop

zuston:
Maybe we could set the SPI file to make hadoop fs to discover the juicefs directly.

### DIFF
--- a/sdk/java/src/main/resources/META-INF/services/org.apache.hadoop.fs.FileSystem
+++ b/sdk/java/src/main/resources/META-INF/services/org.apache.hadoop.fs.FileSystem
@@ -1,0 +1,15 @@
+# JuiceFS, Copyright 2022 Juicedata, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


davies:
@zuston Can you update the docs to explain how to use it?

Please sign the CLA, thanks!

tangyoupeng:
We may not want to initialize JuiceFS if the user didn't access JuiceFS. 
In order to compatible with hdfs, we patched some class in runtime which should not work without JuiceFS. If we use SPI and we put the jar in the classpath, it will automatically patched some class.

davies:
@zuston What's the benefits from being discovered automatically?

zuston:
> @zuston What's the benefits from being discovered automatically?

Just dont need to set the conf of `fs.jfs.impl`. 

It is also ok to specify fs explicitly.

zuston:
> We may not want to initialize JuiceFS if the user didn't access JuiceFS. In order to compatible with hdfs, we patched some class in runtime which should not work without JuiceFS. If we use SPI and we put the jar in the classpath, it will automatically patched some class.

Got it. Thanks for your explanation.

davies:
Since we need to have meta-url and other configs in the core-site.xml, this benefits may not worth the side effects. Let's not have this for now.  

Thanks!

PR #2054, opened 2022-05-19
meta/tkv: fix value of delSliceKey

### LINKED ISSUE #2053: delete slice key is wrong
**What happened**:

**Environment**:
- Kernel (e.g. `uname -a`):
- Others:

In tkv.go,  delete slice is wrong,   

delete slice key is Lcccccccctttttttt , but when set delete slice , it is set Lttttttttcccccccc.

func (m *kvMeta) delSliceKey(ts int64, chunkid uint64) []byte {
	return m.fmtKey("L",chunkid,  uint64(ts))
}

SandyXSD:
close #2053

### DIFF
--- a/pkg/meta/tkv.go
+++ b/pkg/meta/tkv.go
@@ -167,7 +167,7 @@ All keys:
   Fiiiiiiii          Flocks
   Piiiiiiii          POSIX locks
   Kccccccccnnnn      slice refs
-  Lcccccccctttttttt  delayed slices
+  Lttttttttcccccccc  delayed slices
   SEssssssss         session expire time
   SHssssssss         session heartbeat // for legacy client
   SIssssssss         session info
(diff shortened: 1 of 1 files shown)


pkg/meta/tkv.go
```diff
@@ -1851,7 +1851,7 @@ func (m *kvMeta) doDeleteFileData(inode Ino, length uint64) {
 }
 
 func (m *kvMeta) doCleanupDelayedSlices(edge int64, limit int) (int, error) {
-	// delayed slices: Lcccccccctttttttt
+	// delayed slices: Lttttttttcccccccc
 	keys, err := m.scanKeys(m.fmtKey("L"))
```
davies, line 1855: we should use scanRange to fetch only those keys we need.

PR #2055, opened 2022-05-19
add obj bench tool

zhijian-pro:
local disk
image


cmd/bench_for_obj.go
```diff
@@ -0,0 +1,873 @@
...
+				Usage: "access key for object storage (env ACCESS_KEY)",
+			},
+			&cli.StringFlag{
+				Name:  "secret-key",
+				Usage: "secret key for object storage (env SECRET_KEY)",
+			},
+			&cli.UintFlag{
+				Name:  "block-size",
+				Value: 4,
+				Usage: "size of each IO block in MiB",
```
davies, line 63: in KB, same with `format`

zhijian-pro:
@davies Local folder as storage requires sodu permission
image

davies:
@zhijian-pro that's expected, maybe we can ignore the case if it's not run by root.

PR #2056, opened 2022-05-19
support warmup from inside of container

davies:


cmd/rmr.go
```diff
@@ -44,16 +43,22 @@ $ juicefs rmr /mnt/jfs/foo`,
...
+	st, err := os.Stat(mp)
+	if err != nil {
+		logger.Fatal(err)
+	}
+	if !st.IsDir() {
+		mp = filepath.Dir(mp)
 	}
-	if err != nil && path != "/" {
-		return openController(filepath.Dir(path))
+	for ; mp != "/"; mp = filepath.Dir(mp) {
```
SandyXSD, line 54: With the change in vfs we can access .control under any directory in juicefs, so we don't need the for loop any more.

cmd/warmup.go
```diff
@@ -150,59 +151,52 @@ func warmup(ctx *cli.Context) error {
...
-			index++
+		if mp == "/" {
+			inode, err := utils.GetFileInode(path)
+			if err != nil {
+				logger.Errorf("lookup inode for %s: %s", mp, err)
+				continue
+			}
+			batch = append(batch, fmt.Sprintf("inode:%d", inode))
+		} else if strings.HasPrefix(path, mp) {
+			batch = append(batch, string(path[start:]))
```
SandyXSD, line 186: path is already a string

cmd/rmr.go
davies: This loop is used to find out the mount point to pass relative pass into client.

SandyXSD: mp is not returned in this func

davies: For backward compatibility, new version should work with existing mount point.

SandyXSD (approved):

PR #2058, opened 2022-05-19
meta: add parentKey to track all parents of a file

SandyXSD:
TODO: dump/load

### DIFF
--- a/pkg/meta/sql.go
+++ b/pkg/meta/sql.go
@@ -1098,18 +1098,6 @@ func (m *dbMeta) doMknod(ctx Context, parent Ino, name string, _type uint8, mode
 	return errno(err)
 }
 
-func (m *dbMeta) pickParent(ctx Context, s *xorm.Session, inode Ino) (Ino, error) {
-	var rows []linkParent
-	if err := s.Limit(1, 0).ForUpdate().Find(&rows, &linkParent{Inode: inode}); err != nil {
-		return 0, err
-	}
-	var ino Ino
-	if len(rows) > 0 {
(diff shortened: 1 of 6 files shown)


pkg/meta/sql.go
```diff
@@ -2062,6 +2141,19 @@ func (m *dbMeta) CopyFileRange(ctx Context, fin Ino, offIn uint64, fout Ino, off
 	return errno(err)
 }
 
+func (m *dbMeta) doGetParents(ctx Context, inode Ino) map[Ino]int {
+	var rows []linkParent
+	if err := m.db.Find(&rows, &linkParent{Inode: inode}); err != nil {
```
davies, line 2146: inside roTxn

pkg/meta/sql.go
```diff
@@ -61,6 +61,12 @@ type edge struct {
 	Type   uint8  `xorm:"notnull"`
 }
 
+type linkParent struct {
+	Id     int64 `xorm:"pk bigserial"`
+	Inode  Ino   `xorm:"notnull"`
```
davies, line 66: index ?

pkg/meta/sql.go
SandyXSD: OK, need merging main branch

pkg/meta/sql.go
SandyXSD: What index? Do you mean count of each parent?

davies: We need to find all parents for given `inode`, so `inode` should be a index

PR #2091, opened 2022-05-24
cmd/objbench: add content check 

zhijian-pro:
add content check for downloaded files in object storage performance tests

### DIFF
--- a/cmd/objbench.go
+++ b/cmd/objbench.go
@@ -71,7 +71,7 @@ $ ACCESS_KEY=myAccessKey SECRET_KEY=mySecretKey juicefs objbench --storage s3 --
 			},
 			&cli.UintFlag{
 				Name:  "big-object-size",
-				Value: 256,
+				Value: 1024,
 				Usage: "size of each big object in MiB",
 			},
 			&cli.UintFlag{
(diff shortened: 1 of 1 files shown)


cmd/objbench.go
```diff
@@ -173,13 +192,9 @@ func objbench(ctx *cli.Context) error {
...
-			part := make([]byte, 5<<20)
-			total := int(math.Ceil(float64(fsize) / float64(len(part))))
-			rand.Read(part)
-			content := make([][]byte, total)
-			for i := 0; i < total; i++ {
-				content[i] = part
-			}
+			partSize := 5 << 20
+			total := int(math.Ceil(float64(fsize) / float64(partSize)))
+			content := makeData(total, partSize)
```
davies, line 197: we should create the content until upload/download, otherwise we can't run large tests.

cmd/objbench.go
```diff
@@ -188,6 +203,19 @@ func objbench(ctx *cli.Context) error {
 			uploadResult[1], uploadResult[2] = colorize("multi-upload", float64(total*5)/cost, cost, 2, tty)
 			uploadResult[1] += " MiB/s"
 			uploadResult[2] += " s/object"
+
+			logger.Infof("downloading the multipart upload file and checking its contents ...")
+			r, err := blob.Get(fname, 0, -1)
+			if err != nil {
+				logger.Fatalf("failed to get multipart upload file: %v", err)
+			}
+			cnt, err := io.ReadAll(r)
```
davies, line 212: This will be too slow for large test.

Maybe we should limit the size of multipart to 200MB?

cmd/objbench.go
zhijian-pro: The timing starts in the next step, and the time to generate the data is not counted.

cmd/objbench.go
zhijian-pro: 200MB is too small for the internal network, here I think you can add a download progress bar, which will give people a better experience.

davies (approved):

PR #2111, opened 2022-05-26
meta: add pessimistic lock to reduce conflicted transaction in database

davies:
This PR bring the pessimistic lock used in Redis engine to all of the meta engines.

### DIFF
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -82,7 +82,7 @@ type baseMeta struct {
 	fmt  Format
 
 	root         Ino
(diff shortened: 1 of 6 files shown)


pkg/meta/base.go
```diff
@@ -81,6 +82,7 @@ type baseMeta struct {
 	fmt  Format
 
 	root         Ino
+	txlocks      [nlocks]sync.Mutex // Pessimistic locks to reduce conflict on Redis
```
SandyXSD, line 85: not only Redis

SandyXSD (approved):

PR #2142, opened 2022-05-30
meta/tkv: reduce memory for loading into kv storage

davies:
Currently, we need to load all the entries into memory, then write them into kv storage one by one. This PR change it to load entries as a stream and write them into kv storage as batches.

For a volume with 10 millions of files, this PR reduce  the memory from 20G to 1G,  reduce  the time to load into badge from 12 minutes to 2 minutes.

### DIFF
--- a/pkg/meta/tkv.go
+++ b/pkg/meta/tkv.go
@@ -2519,7 +2519,7 @@ type pair struct {
 	value []byte
 }
 
-func (m *kvMeta) loadEntry(e *DumpedEntry, cs *DumpedCounters, parents map[Ino][]Ino, refs map[chunkKey]int64, kv chan *pair) error {
+func (m *kvMeta) loadEntry(e *DumpedEntry, kv chan *pair) error {
 	inode := e.Attr.Inode
 	attr := loadAttr(e.Attr)
 	attr.Parent = e.Parents[0]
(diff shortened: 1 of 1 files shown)


pkg/meta/tkv.go
```diff
@@ -2514,74 +2514,142 @@ func (m *kvMeta) DumpMeta(w io.Writer, root Ino) (err error) {
...
-			if cs.NextTrash < int64(inode)-TrashInode {
-				cs.NextTrash = int64(inode) - TrashInode
-			}
+		if err != nil {
+			return nil, fmt.Errorf("decode %v: %s", name, err)
 		}
-		m.Unlock()
+	}
+	if len(e.Parents) == 1 {
+		_ = m.loadEntry(&e, cs, parents, refs, kv)
```
SandyXSD, line 2640: In this case, we use the attribute of the first link, not the newest one any more

pkg/meta/tkv.go
```diff
@@ -2683,9 +2770,28 @@ func (m *kvMeta) LoadMeta(r io.Reader) error {
 		for _, d := range dm.DelFiles {
 			tx.set(m.delfileKey(d.Inode, d.Length), m.packInt64(d.Expire))
 		}
+		// update parents for hardlinks
+		st := make(map[Ino]int64)
+		for i, ps := range parents {
+			if len(ps) > 1 {
+				a := tx.get(m.inodeKey(i))
+				// reset parent
+				binary.BigEndian.PutUint64(a[63:71], 0)
```
SandyXSD, line 2779: a[64:72] ?

SandyXSD (changes requested):

pkg/meta/tkv.go
```diff
@@ -2597,81 +2665,100 @@ func (m *kvMeta) LoadMeta(r io.Reader) error {
...
+				bar.SetTotal(dm.Counters.UsedInodes)
+			}
+		case "Sustained":
+			err = dec.Decode(&dm.Sustained)
+		case "DelFiles":
+			err = dec.Decode(&dm.DelFiles)
+		case "FSTree":
+			_, err = m.decodeEntry(dec, 1, counters, parents, refs, kv, showProgress)
+		case "Trash":
+			_, err = m.decodeEntry(dec, 1, counters, parents, refs, kv, showProgress)
```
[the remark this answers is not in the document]
davies: Why not 1?  the path is /.trash/

pkg/meta/tkv.go
davies: no, there is 8 byte left

SandyXSD (approved):

PR #2147, opened 2022-05-31
Fix nlink of loaded hardlink

davies:


pkg/meta/tkv.go
```diff
@@ -2562,6 +2562,7 @@ type pair struct {
 func (m *kvMeta) loadEntry(e *DumpedEntry, kv chan *pair) error {
 	inode := e.Attr.Inode
 	attr := loadAttr(e.Attr)
+	attr.Nlink = 1
```
SandyXSD, line 2565: Better do it in `decodeEntry`, as a general behavior for all metadata engines.

davies: let's consider this later when we implement this for other engines.

PR #2148, opened 2022-05-31
meta: speed up and reduce memory for loading metadata

SandyXSD:


pkg/meta/sql.go
```diff
@@ -3133,155 +3092,95 @@ func (m *dbMeta) LoadMeta(r io.Reader) error {
...
+	chs[5] <- &counter{totalInodes, counters.UsedInodes}
+	chs[5] <- &counter{"nextInode", counters.NextInode}
+	chs[5] <- &counter{"nextChunk", counters.NextChunk}
+	chs[5] <- &counter{"nextSession", counters.NextSession}
+	chs[5] <- &counter{"nextTrash", counters.NextTrash}
+	for _, d := range dm.DelFiles {
+		chs[5] <- &delfile{d.Inode, d.Length, d.Expire}
 	}
+	for k, v := range refs {
+		chs[3] <- &chunkRef{k.id, k.size, int(v)}
```
davies, line 3160: we should insert those with ref =1, then update them if ref > 1

SandyXSD: `chunkRef` is not inserted until in this for loop, so no need to update

SandyXSD: This could be an issue, if there are too many keys in `refs`.

davies: there could be more rows than node

PR #2156, opened 2022-06-02
meta/redis: Speedup dump for Redis and reduce memory usage

davies:
Currently, we dump all the key-values from Redis into JuiceFS's memory, then build the file system tree and dump them as JSON stream, it require lots of memory to hold all the metadata, for example, 10 millions of files require 9GB memory.

Since Redis holds all the data in memory, we can access them very fast in random order, similar to the memory snapshot it builds. So this PR remove the in-process snapshot and fetch data from Redis in batch and in parallel.

For a test with 11 millions of small files, this PR can reduce the time to dump from 13 minutes to 3.5 minutes, it also reduce the memory from 9GB to 300 MB.

### DIFF
--- a/pkg/meta/dump.go
+++ b/pkg/meta/dump.go
@@ -260,7 +260,7 @@ func (dm *DumpedMeta) writeJsonWithOutTree(w io.Writer) (*bufio.Writer, error) {
 	return bw, nil
 }
 
-func dumpAttr(a *Attr, d *DumpedAttr) *DumpedAttr {
+func dumpAttr(a *Attr, d *DumpedAttr) {
 	d.Type = typeToString(a.Typ)
 	d.Mode = a.Mode
 	d.Uid = a.Uid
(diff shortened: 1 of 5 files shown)


pkg/meta/dump.go
```diff
@@ -260,25 +260,24 @@ func (dm *DumpedMeta) writeJsonWithOutTree(w io.Writer) (*bufio.Writer, error) {
 	return bw, nil
 }
 
-func dumpAttr(a *Attr) *DumpedAttr {
```
SandyXSD, line 263: d could be nil?

davies: Yes, the caller should make sure that it's not nil

pkg/meta/tkv.go
```diff
@@ -2247,7 +2247,7 @@ func (m *kvMeta) dumpEntry(inode Ino, e *DumpedEntry) error {
 		if a == nil && e.Attr != nil {
 			attr.Typ = typeFromString(e.Attr.Type)
 		}
-		e.Attr = dumpAttr(attr) // TODO: reduce memory allocation
+		dumpAttr(attr, e.Attr)
```
SandyXSD, line 2250: here e.Attr may be nil

PR #2172, opened 2022-06-07
meta: add more logs when there is a lot of transaction restarts

SandyXSD:


pkg/meta/redis.go
```diff
@@ -781,7 +781,14 @@ func (m *redisMeta) txn(ctx Context, txf func(tx *redis.Tx) error, keys ...strin
 		err = m.rdb.Watch(ctx, txf, keys...)
 		if m.shouldRetry(err, retryOnFailture) {
 			txRestart.Add(1)
-			logger.Debugf("Transaction failed, restart it (tried %d): %s", i+1, err)
+			msg := fmt.Sprintf("Transaction failed, restart it (tried %d): %s", i+1, err)
+			if i+1 == 10 {
+				logger.Infof(msg)
```
davies, line 786: Maybe we can log it when the transaction succeed after N retries

PR #2214, opened 2022-06-13
cmd/rmr: remove entries in batch

davies:
When removing directories with many files in parallel, they need lots of memory for these entries. 

This PR change to remove them as smaller batches (10000), so it will hold at most 500K entries in memory (100-200MB). 

For smaller directories, it will take one more readdir() to ensure that the directory is empty (better for concurrent adding files).

### DIFF
--- a/pkg/meta/utils.go
+++ b/pkg/meta/utils.go
@@ -226,55 +226,60 @@ func emptyDir(r Meta, ctx Context, inode Ino, count *uint64, concurrent chan int
 	if st := r.Access(ctx, inode, 3, nil); st != 0 {
 		return st
 	}
-	var entries []*Entry
-	if st := r.Readdir(ctx, inode, 0, &entries); st != 0 {
-		return st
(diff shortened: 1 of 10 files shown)


pkg/meta/utils.go
```diff
@@ -226,55 +226,60 @@ func emptyDir(r Meta, ctx Context, inode Ino, count *uint64, concurrent chan int
...
-	var status syscall.Errno
-	// try directories first to increase parallel
-	var dirs int
-	for i, e := range entries {
-		if e.Attr.Typ == TypeDirectory {
-			entries[dirs], entries[i] = entries[i], entries[dirs]
-			dirs++
+	for {
+		var entries []*Entry
+		if st := r.getEngine().doReaddir(ctx, inode, 0, &entries, 10000); st != 0 {
```
SandyXSD, line 231: need `checkRoot` in case it's mounted with subdir

PR #2227, opened 2022-06-15
Fix stats of cached space in writeback mode

### LINKED ISSUE #2225: [disk cache]cache size should not subtract size when the block is staging because the size is not added.
func (cache *cacheStore) remove(key string) {
	cache.Lock()
	path := cache.cachePath(key)
	if cache.keys[key].atime > 0 {                                             --------> if cache.keys[key].atime > 0 && cache.keys[key].size > 0 {
		cache.used -= int64(cache.keys[key].size + 4096)
		delete(cache.keys, key)
	} else if cache.scanned {
		path = "" // not existed
	}
	cache.Unlock()
	if path != "" {
		_ = os.Remove(path)
		stagingPath := cache.stagePath(key)
		if fi, err := os.Stat(stagingPath); err == nil {
			size := fi.Size()
			if err = os.Remove(stagingPath); err == nil {
				stageBlocks.Sub(1)
				stageBlockBytes.Sub(float64(size))
			}
		}
	}
}

func (cache *cacheStore) add(key string, size int32, atime uint32) {
	cache.Lock()
	defer cache.Unlock()
	it, ok := cache.keys[key]
	if ok && it.size > 0 {
		cache.used -= int64(it.size + 4096)
	}
	if atime == 0 {
		// update size of staging block
		cache.keys[key] = cacheItem{size, it.atime}
	} else {
		cache.keys[key] = cacheItem{size, atime}
	}
	if size > 0 {
		cache.used += int64(size + 4096)                            -------------> when size > 0, we add the block size.
	}

	if cache.used > cache.capacity {
		logger.Debugf("Cleanup cache when add new data (%s): %d blocks (%d MB)", cache.dir, len(cache.keys), cache.used>>20)
		cache.cleanup()
	}
}

davies:
closes #2225

### DIFF
--- a/pkg/chunk/disk_cache.go
+++ b/pkg/chunk/disk_cache.go
@@ -301,9 +301,7 @@ func (cache *cacheStore) add(key string, size int32, atime uint32) {
 	} else {
 		cache.keys[key] = cacheItem{size, atime}
 	}
-	if size > 0 {
-		cache.used += int64(size + 4096)
-	}
+	cache.used += int64(size + 4096)
 
 	if cache.used > cache.capacity {
 		logger.Debugf("Cleanup cache when add new data (%s): %d blocks (%d MB)", cache.dir, len(cache.keys), cache.used>>20)
--- a/pkg/chunk/disk_cache.go
+++ b/pkg/chunk/disk_cache.go
@@ -220,8 +220,10 @@ func (cache *cacheStore) createDir(dir string) {
 func (cache *cacheStore) remove(key string) {
 	cache.Lock()
 	path := cache.cachePath(key)
-	if cache.keys[key].atime > 0 {
-		cache.used -= int64(cache.keys[key].size + 4096)
+	if it, ok := cache.keys[key]; ok {
+		if it.size > 0 {
+			cache.used -= int64(it.size + 4096)
+		}
 		delete(cache.keys, key)
 	} else if cache.scanned {
 		path = "" // not existed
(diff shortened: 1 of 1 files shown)


penglaixy99:
I think we should  just  modify  remove function.

func (cache *cacheStore) remove(key string) {
      cache.Lock()
      path := cache.cachePath(key)
      if cache.keys[key].atime > 0 { -------->   replaced with       if cache.keys[key].atime > 0 && cache.keys[key].size > 0 {
                 cache.used -= int64(cache.keys[key].size + 4096)
                 delete(cache.keys, key)
      } else if cache.scanned {
                 path = "" // not existed
      }

cached.used is caculated as only read cached size and should not add staging block. And when add staging block ,size argument is set as < 0.

func (cache *cacheStore) stage(key string, data []byte, keepCache bool) (string, error) {
	stagingPath := cache.stagePath(key)
	if cache.full {
		return stagingPath, errors.New("Space not enough on device")
	}
	err := cache.flushPage(stagingPath, data)
	if err == nil {
		stageBlocks.Add(1)
		stageBlockBytes.Add(float64(len(data)))
		if cache.capacity > 0 && keepCache {
			path := cache.cachePath(key)
			cache.createDir(filepath.Dir(path))
			if err := os.Link(stagingPath, path); err == nil {
				cache.add(key, -int32(len(data)), uint32(time.Now().Unix()))              --------> look at this line
			} else {
				logger.Warnf("link %s to %s failed: %s", stagingPath, path, err)
			}
		}
	}
	return stagingPath, err
}

SandyXSD (approved):

penglaixy99:
Agree with this PR!

PR #2261, opened 2022-06-17
support session token for object storage

### LINKED ISSUE #2236: Access S3 using Session token
Besides AK/SK and IAM roles, it will be better to support session token as well.

davies:
close #2236

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -170,15 +170,15 @@ func config(ctx *cli.Context) error {
 			}
 		case "secret-key": // always update
 			msg.WriteString(fmt.Sprintf("%10s: updated\n", flag))
-			if err := format.Decrypt(); err != nil {
-				logger.Fatalf("decrypt secrets: %s", err)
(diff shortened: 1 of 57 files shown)


cmd/config.go
```diff
@@ -167,8 +171,18 @@ func config(ctx *cli.Context) error {
...
+				logger.Fatalf("decrypt secrets: %s", err)
+			}
 			format.SecretKey = ctx.String(flag)
-			format.KeyEncrypted = false
+			storage = true
+		case "session-token": // always update
+			msg.WriteString(fmt.Sprintf("%10s: updated\n", flag))
+			encrypted = format.KeyEncrypted
+			if err := format.Decrypt(); err != nil {
+				logger.Fatalf("decrypt secrets: %s", err)
```
SandyXSD, line 183: secrets -> session token

davies: it will decrypt all the secrets

PR #2263, opened 2022-06-17
object/azure: fix azure list api

zhijian-pro:


pkg/object/azure.go
```diff
@@ -159,51 +163,29 @@ func newWabs(endpoint, accountName, accountKey string) (ObjectStorage, error) {
...
-	containerName := hostParts[0]
+	var domain string
 	if len(hostParts) > 1 {
 		// Arguments from command line take precedence
-		domain = hostParts[1]
-	} else if domain == "" {
-		if domain, err = autoWasbEndpoint(containerName, accountName, scheme, credential); err != nil {
-			return nil, fmt.Errorf("Unable to get endpoint of container %s: %s", containerName, err)
-		}
+		domain = fmt.Sprintf("blob.%s", hostParts[1])
```
davies, line 183: hostParts[1] may already have 'blob.'

pkg/object/azure.go
```diff
@@ -104,12 +104,16 @@ func (b *wasb) List(prefix, marker string, limit int64) ([]Object, error) {
 
 	limit32 := int32(limit)
 	pager := b.container.ListBlobsFlat(&azblob.ContainerListBlobFlatSegmentOptions{Prefix: &prefix, Marker: &marker, Maxresults: &(limit32)})
+	if pager.Err() != nil {
+		return nil, pager.Err()
+	}
+	var n int
 	if pager.NextPage(ctx) {
 		b.marker = *pager.PageResponse().NextMarker
+		n = len(pager.PageResponse().Segment.BlobItems)
```
davies, line 113: n should be the number of items in current page, n could be > 0 if even NextPage() is false

PR #2294, opened 2022-06-23
cmd/info: Fix repeat path info when inode is 1

201341:


pkg/meta/interface.go
```diff
@@ -475,6 +475,9 @@ func GetPaths(m Meta, ctx Context, inode Ino) []string {
 				c++
 				paths = append(paths, path.Join(dir, string(e.Name)))
 			}
+			if inode == 1 {
```
SandyXSD, line 478: Check `inode == 1` at the beginning of the `GetPaths` should be simpler.

PR #2301, opened 2022-06-23
cmd/info: Add an filter to exclude inode 0

201341:


davies:
Have you run into problem before this patch?

201341:
> Have you run into problem before this patch?

./juicefs -i 0.  It will print some unknown info, which shouldn't be appeared.

PR #2303, opened 2022-06-23
cmd/info: make output more human friendly

### LINKED ISSUE #2300: Add an option for info to make data more human-readable
Add an option for info
**Why is this needed**:
when using info by adding flag --recursive, the size will be too larger to read. So the data should be converted into a human readable format.

SandyXSD:
close #2300

### DIFF
--- a/pkg/vfs/internal.go
+++ b/pkg/vfs/internal.go
@@ -243,19 +243,19 @@ func (v *VFS) handleInternalMsg(ctx meta.Context, cmd uint32, r *utils.Buffer, d
 			return
 		}
 		var w = bytes.NewBuffer(nil)
-		fmt.Fprintf(w, " inode: %d\n", inode)
-		fmt.Fprintf(w, " files:\t%d\n", summary.Files)
-		fmt.Fprintf(w, " dirs:\t%d\n", summary.Dirs)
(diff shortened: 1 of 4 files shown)


pkg/vfs/internal.go
```diff
@@ -243,19 +243,19 @@ func (v *VFS) handleInternalMsg(ctx meta.Context, cmd uint32, r *utils.Buffer, d
 			return
 		}
 		var w = bytes.NewBuffer(nil)
-		fmt.Fprintf(w, " inode: %d\n", inode)
-		fmt.Fprintf(w, " files:\t%d\n", summary.Files)
-		fmt.Fprintf(w, " dirs:\t%d\n", summary.Dirs)
-		fmt.Fprintf(w, " length:\t%d\n", summary.Length)
```
davies, line 249: We may want to know the exact number, so this should be controlled by a flag.

SandyXSD: Both formats will be printed, like:
```
/Users/sandy/jfs :
  inode: 1
  files: 9
   dirs: 2
 length: 105.00 MiB (110100528 Bytes)
   size: 105.04 MiB (110141440 Bytes)
   path: /
```

PR #2326, opened 2022-07-01
cmd/sync: speedup downloading big file

### LINKED ISSUE #1431: Download Large object in parallel
When the destination don't support multiple upload (for example, local disk), it fallback to single thread. But download a object using single tcp connection from object store could be much slower than writing to destination (disk), we could use multiple threads to download and use 1 thread to write.

zhijian-pro:
Close #1431 
image
3x faster performance


pkg/sync/sync.go
```diff
@@ -326,6 +326,153 @@ func checkSum(src, dst object.ObjectStorage, key string, size int64) (bool, erro
...
+	}
+	var bidx int64
+	for {
+		if r.fidx < fsize && bidx < total {
+			c := r.fidx / r.blockSize % int64(r.concurrent)
+			blockIdx := r.fidx % r.blockSize
+			length := r.blockSize - blockIdx
+			if blockIdx == 0 {
+				<-r.readyCh[c]
+				if err != nil {
```
SandyXSD, line 456: `err` is set only for the first read

pkg/sync/sync.go
```diff
@@ -326,6 +326,153 @@ func checkSum(src, dst object.ObjectStorage, key string, size int64) (bool, erro
...
+		for c := 0; c < concurrent; c++ {
+			go func(c int) {
+				var i int64
+				for i = int64(c) * blockSize; i < fsize; i += int64(concurrent) * blockSize {
+					readClose, e := r.src.Get(r.key, i, blockSize)
+					if e != nil {
+						err = e
+					} else {
+						var e2 error
+						if r.cache[i/blockSize%int64(concurrent)], e2 = io.ReadAll(readClose); e2 != nil {
```
SandyXSD, line 432: better to reuse the cache

pkg/sync/sync.go
```diff
@@ -326,6 +326,153 @@ func checkSum(src, dst object.ObjectStorage, key string, size int64) (bool, erro
...
+					_ = readClose.Close()
+				}
+				r.readyCh[c] <- token{}
+				r.preCh[c] <- token{}
+			}
+		}(c)
+	}
+	var counter int64
+	for i := 0; i < int(fsize/blockSize); i++ {
+		if err != nil {
```
SandyXSD, line 377: check error after `readyCh`

pkg/sync/sync.go
```diff
@@ -327,6 +327,43 @@ func checkSum(src, dst object.ObjectStorage, key string, size int64) (bool, erro
...
+			}
+		}
+		if err == nil {
+			err = dst.Put(key, in)
+		}
+		if err != nil {
+			if _, e := src.Head(key); os.IsNotExist(e) {
+				logger.Debugf("Head src %s: %s", key, err)
+				copied.IncrInt64(-1)
+				copiedBytes.IncrInt64(size * -1)
```
SandyXSD, line 360: this is probably not needed

SandyXSD (approved):

PR #2330, opened 2022-07-03
pkg/object: support chacha20 cipher

### LINKED ISSUE #317: Support for modern stream ciphers like Chacha20
Support for modern/lightweight stream ciphers like Chacha20 beside current support of AES.

**Why is this needed**:
ChaCha20 takes a 256-bit key and a 32-bit nonce and then creates a key stream, which is then XORed with the plaintext stream. In software, it is three times faster than AES, and is well suited to lower-powered devices and in real time communications.

Quoted from -> https://medium.com/asecuritysite-when-bob-met-alice/aes-is-great-but-we-need-a-fall-back-meet-chacha-and-poly1305-76ee0ee61895

Also a nice coffee-time basic article https://blog.cloudflare.com/do-the-chacha-better-mobile-performance-with-cryptography/

qingyunha:
close #317

### DIFF
--- a/cmd/format.go
+++ b/cmd/format.go
@@ -242,7 +242,7 @@ func createStorage(format meta.Format) (object.ObjectStorage, error) {
 		if err != nil {
 			return nil, fmt.Errorf("incorrect passphrase: %s", err)
 		}
-		encryptor, err := object.NewDataEncryptor(object.NewRSAEncryptor(privKey), format.EncryptAglo)
+		encryptor, err := object.NewDataEncryptor(object.NewRSAEncryptor(privKey), format.EncryptAlgo)
 		if err != nil {
 			return nil, err
 		}
@@ -385,7 +385,7 @@ func format(c *cli.Context) error {
 				format.HashPrefix = c.Bool(flag)
 			case "storage":
 				format.Storage = c.String(flag)
-			case "encrypt-rsa-key":
+			case "encrypt-rsa-key", "encrypt-algo":
 				logger.Warnf("Flag %s is ignored since it cannot be updated", flag)
 			}
 		}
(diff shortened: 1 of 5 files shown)


pkg/meta/config.go
```diff
@@ -59,6 +59,7 @@ type Format struct {
 	Inodes           uint64 `json:",omitempty"`
 	EncryptKey       string `json:",omitempty"`
 	KeyEncrypted     bool   `json:",omitempty"`
+	ChaCha20         bool   `json:",omitempty"`
```
davies, line 62: I think we should have `EncryptAlgo = chacha20-rsa`, so we can support other algorithm later.

qingyunha: done

davies:
@qingyunha Thanks for contributing this, it look good overall, more work needed for compatibility, and config/load/dump. 

Since we are close to 1.0 release, it's better to merge this after 1.0 released.

qingyunha:
Ok. Thank you.

pkg/meta/config.go
```diff
@@ -58,6 +58,7 @@ type Format struct {
 	Capacity         uint64 `json:",omitempty"`
 	Inodes           uint64 `json:",omitempty"`
 	EncryptKey       string `json:",omitempty"`
+	EncryptAglo      string `json:",omitempty"`
```
SandyXSD, line 61: Should be `EncryptAlgo`

pkg/object/encrypt_test.go
```diff
@@ -126,7 +126,7 @@ func BenchmarkRSA4096Decrypt(b *testing.B) {
 
 func TestAESGCM(t *testing.T) {
 	kc := NewRSAEncryptor(testkey)
-	dc := NewAESEncryptor(kc)
```
SandyXSD, line 129: There are 2 return values

SandyXSD: Could you also add tests for ChaCha20?

cmd/format.go
```diff
@@ -392,6 +400,7 @@ func format(c *cli.Context) error {
 			SecretKey:    c.String("secret-key"),
 			SessionToken: c.String("session-token"),
 			EncryptKey:   loadEncrypt(c.String("encrypt-rsa-key")),
+			EncryptAglo:  c.String("encrypt-aglo"),
```
SandyXSD, line 403: This flag should be ignored like `encrypt-rsa-key` as handled in L388~L389

SandyXSD:
@qingyunha Hi, I leave a few comments, could you continue to complete this PR?

qingyunha:
@SandyXSD Thank you. Updated, please take a look.

SandyXSD (approved):

PR #2360, opened 2022-07-13
cmd/doctor: doctor command supports collecting system, syslog and other info

Ziy1-Tan:
## Description
The following file will be created:
- [x] `cmd/doctor.go` 
- [x] `cmd/doctor_test.go`

## Features
It collects and displays info from multiple dimensions such as the running environment and system logs to report bugs, etc.

- **System info collection:** support collection of os, architecture, hardware, juicefs version, etc.
- **Syslog collection:** support automatic search for log location and filter logs according to the maximum number of records 
-  **Pprof metric collection:** Goroutine stack information, CPU performance statistics, memory allocation statistics, etc.
- **Mount point config collection:** support specifying mount point to obtain the config (filter sensitive info)
- **Mount command collection:** Collect juicefs mount commands and corresponding parameters

## Use cases
```
$ juicefs doctor /mnt/jfs

# Result will be output to /var/log/
$ juicefs doctor --out-dir=/var/log /mnt/jfs

# Get log file up to last 1000 entries
$ juicefs doctor --out-dir=/var/log --collect-log --limit=1000 /mnt/jfs 

# Enable pprof collection
$ juicefs doctor --out-dir=/var/log --collect-log --limit=1000 --collect-pprof /mnt/jfs

```

## How can this PR be tested?
The test files locate at `cmd/doctor_test.go`.

## Basing the PR against the correct JuiceFS version
- [x] This is a new feature and the PR is based against the latest JuiceFS development branch
- [ ] This is a bug fix and the PR is based against the earliest branch in which the bug can be reproduce

### DIFF
--- a/cmd/doctor.go
+++ b/cmd/doctor.go
@@ -56,7 +56,7 @@ func cmdDoctor() *cli.Command {
 		ArgsUsage: "MOUNTPOINT",
 		Usage:     "Collect and display system static and runtime information",
 		Description: `
-It collects and display information from multiple dimensions such as the running environment and system logs, etc.
+It collects and displays information from multiple dimensions such as the running environment and system logs, etc.
 
 Examples:
 $ juicefs doctor /mnt/jfs
@@ -377,15 +377,19 @@ func checkAgent(cmd string) bool {
 }
 
 func geneZipFile(srcPath, destPath string) error {
-	zipfile, err := os.Create(destPath)
+	zipFile, err := os.Create(destPath)
 	if err != nil {
 		return err
 	}
-	defer zipfile.Close()
-	archive := zip.NewWriter(zipfile)
-	defer archive.Close()
+	defer closeFile(zipFile)
+	archive := zip.NewWriter(zipFile)
+	defer func() {
+		if err := archive.Close(); err != nil {
+			logger.Fatalf("error closing zip archive: %v", err)
+		}
+	}()
 
-	filepath.Walk(srcPath, func(path string, info os.FileInfo, _ error) error {
+	if err = filepath.Walk(srcPath, func(path string, info os.FileInfo, _ error) error {
 		if path == srcPath {
 			return nil
 		}
@@ -407,13 +411,16 @@ func geneZipFile(srcPath, destPath string) error {
 		}
 		if !info.IsDir() {
 			file, _ := os.Open(path)
-			defer file.Close()
+			defer closeFile(file)
 			if _, err := io.Copy(writer, file); err != nil {
 				return err
 			}
 		}
 		return nil
-	})
+	}); err != nil {
+		return err
+	}
+
 	return nil
 }
 
--- a/cmd/doctor.go
+++ b/cmd/doctor.go
@@ -126,7 +126,7 @@ func copyConfigFile(srcPath, destPath string, rootPrivileges bool) error {
 	if rootPrivileges {
 		copyArgs = append(copyArgs, "sudo")
 	}
-	copyArgs = append(copyArgs, "bash", "-c", fmt.Sprintf("cp %s %s", srcPath, destPath))
+	copyArgs = append(copyArgs, "/bin/sh", "-c", fmt.Sprintf("cp %s %s", srcPath, destPath))
 	if err := exec.Command(copyArgs[0], copyArgs[1:]...).Run(); err != nil {
 		fmt.Println(strings.Join(copyArgs, " "))
 		return err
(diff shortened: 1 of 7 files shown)


zhijian-pro (approved):

cmd/doctor.go
```diff
@@ -0,0 +1,577 @@
...
+}
+
+func doctor(ctx *cli.Context) error {
+	setup(ctx, 1)
+	mp := ctx.Args().First()
+	inode, err := utils.GetFileInode(mp)
+	if err != nil {
+		return fmt.Errorf("lookup inode for %s: %s", mp, err)
+	}
+	if inode != 1 {
```
SandyXSD, line 410: use `meta.Root`

cmd/doctor.go
```diff
@@ -0,0 +1,577 @@
...
+	outDir := ctx.String("out-dir")
+	// special treatment for non-existing out dir
+	if _, err := os.Stat(outDir); os.IsNotExist(err) {
+		if err := os.Mkdir(outDir, os.ModePerm); err != nil {
+			return fmt.Errorf("failed to create out dir %s: %v", outDir, err)
+		}
+	}
+
+	// double check the file stat
+	outDirInfo, err := os.Stat(outDir)
```
SandyXSD, line 423: no need to check again; instead check its type in the previous `stat`

cmd/doctor.go
```diff
@@ -0,0 +1,577 @@
...
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/juicedata/juicefs/pkg/utils"
+
+	"github.com/urfave/cli/v2"
+)
+
+var defaultOutDir = path.Join(".", "doctor")
```
SandyXSD, line 42: use `filepath` if you want to use it on Windows as well

cmd/doctor.go
```diff
@@ -0,0 +1,577 @@
...
+
+	result := fmt.Sprintf(`Platform: 
+%s %s
+%s
+JuiceFS Version:
+%s`, runtime.GOOS, runtime.GOARCH, sysInfo, ctx.App.Version)
+
+	sysPath := path.Join(currDir, "system-info.log")
+	sysFile, err := os.Create(sysPath)
+	defer closeFile(sysFile)
```
SandyXSD, line 453: check `err` before `defer close`

cmd/doctor.go
```diff
@@ -0,0 +1,577 @@
...
+func copyConfigFile(srcPath, destPath string, rootPrivileges bool) error {
+	if runtime.GOOS == "windows" {
+		return copyVolumeConfWindows(srcPath, destPath)
+	}
+
+	var copyArgs []string
+	if rootPrivileges {
+		copyArgs = append(copyArgs, "sudo")
+	}
+	copyArgs = append(copyArgs, "/bin/sh", "-c", fmt.Sprintf("cat %s > %s", srcPath, destPath))
```
SandyXSD, line 135: just call `cp` ?

cmd/doctor.go
```diff
@@ -0,0 +1,577 @@
...
+`,
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:  "out-dir",
+				Value: defaultOutDir,
+				Usage: "the output directory of the result file",
+			},
+			&cli.Uint64Flag{
+				Name:  "stats-sec",
+				Value: defaultStatsTime,
```
SandyXSD, line 80: set default value here directly, no need for the global const/var

cmd/doctor.go
Ziy1-Tan: check if `stat` is nil? well I code this because `stat` will be nil if the directory doesn't exist.

SandyXSD: The `fileInfo` shouldn't be nil if `stat` return a nil err. If the dir didn't exist and is newly created, it should be a directory, and no need to check it again.

cmd/doctor.go
Ziy1-Tan: `cp` will have permission issues when compressing the result. cp does not change file permissions when mounted via root. `cat` implements copying while changing file permissions.

cmd/doctor.go
```diff
@@ -0,0 +1,574 @@
...
+		}
+		baseUrl := fmt.Sprintf("http://localhost:%d/debug/pprof/", port)
+		trace := ctx.Uint64("trace-sec")
+		profile := ctx.Uint64("profile-sec")
+		metrics := map[string]metricItem{
+			"allocs":       {name: "allocs.pb.gz", url: baseUrl + "allocs"},
+			"blocks":       {name: "block.pb.gz", url: baseUrl + "block"},
+			"cmdline":      {name: "cmdline.txt", url: baseUrl + "cmdline"},
+			"goroutine":    {name: "goroutine.pb.gz", url: baseUrl + "goroutine"},
+			"fullstack":    {name: "full.goroutine.stack.txt", url: baseUrl + "goroutine?debug=2"},
```
SandyXSD, line 538: usually `debug=1` is enough.

Ziy1-Tan: Ok.

SandyXSD (approved):

PR #2383, opened 2022-07-21
obj/redis: fix redis cluster list all api

zhijian-pro:


cmd/objbench.go
```diff
@@ -818,6 +818,26 @@ func functionalTesting(blob object.ObjectStorage, result *[][]string, tty bool)
 			} else if len(objs) != 0 {
 				return fmt.Errorf("list should not return anything, but got %d", len(objs))
 			}
+			for i := 0; i < 4; i++ {
```
davies, line 821: 100

PR #2409, opened 2022-07-27
meta: reset variables at the beginning of transactions

SandyXSD:


davies (approved):

pkg/meta/redis.go
```diff
@@ -802,6 +802,7 @@ func (m *redisMeta) Truncate(ctx Context, inode Ino, flags uint8, length uint64,
 	defer func() { m.of.InvalidateChunk(inode, 0xFFFFFFFF) }()
 	var newSpace int64
 	err := m.txn(ctx, func(tx *redis.Tx) error {
+		newSpace = 0
```
qingyunha, line 805: @SandyXSD Hello, may I ask why the assignment is required? Doesn't newSpace already equal zero?

SandyXSD: The func may run several times in `m.txn`.

qingyunha: ok, thank you.

PR #2445, opened 2022-08-02
pkg/gateway: Support specifying umask for directories

dugusword:
Currently, gateway does not respect the umask flag specified in the start command, instead, it hard codes 755 for all the directories it creates. I have added an option, umask-dir, which will dictate the umask for all the mkdir calls. I keep them as separate flags in case some users would like their umasks different for folders and directories.

The following screenshots show different behaviors when gateway is started with umask-dir 022 vs 000.
image

### DIFF
--- a/cmd/gateway.go
+++ b/cmd/gateway.go
@@ -59,7 +59,12 @@ func cmdGateway() *cli.Command {
 		&cli.StringFlag{
 			Name:  "umask",
 			Value: "022",
-			Usage: "umask for new file in octal",
+			Usage: "umask for new files in octal",
+		},
+		&cli.StringFlag{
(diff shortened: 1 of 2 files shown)


davies:
I think we should use the same umask for both file and dirs. When some user really need a separate umask, we can add that later.

davies (changes requested):
use same umask for both file and directory

dugusword:
> I think we should use the same umask for both file and dirs. When some user really need a separate umask, we can add that later.

Cool. I can make the update. One question I do have is, though, I believe when using a shell, the base permission for file is 0666 while for directories it's 0777. Should we have the same behavior?

davies (approved):

PR #2446, opened 2022-08-02
add / update fstab entry with --update-fstab

timfeirg:


timfeirg:
close in favor of https://github.com/juicedata/juicefs/pull/2462

PR #2457, opened 2022-08-04
release & test: use Go 1.18

SandyXSD:


SandyXSD:
duplicated with #2458

Issue #2432, opened 2022-08-01
Add the automatic mount script


zhijian-pro:
Boot automatic mount steps are tedious, we should use scripts to help users make the setup easier.

timfeirg:
might not be an actual script (and relevant docs), instead I'll try to integrate `--update-fstab` functionality into [mount.go](https://github.com/juicedata/juicefs/blob/main/cmd/mount.go)

PR #2462, opened 2022-08-04
add / update fstab entry with --update-fstab

timfeirg:
- [x] abs path check for sqlite and similars
- [x] correctly handle slice arguments, even though they don't exist in `juicefs mount` yet
- [x] do not run in containers
- [x] unit tests

PR code will produce the following result:

image

permission denied behavior:

image

closes https://github.com/juicedata/juicefs/issues/2432

### DIFF
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -17,6 +17,7 @@
 package cmd
 
 import (
+	"bufio"
 	"fmt"
 	"net"
 	"net/http"
@@ -25,6 +26,7 @@ import (
 	"os/signal"
 	"path/filepath"
 	"runtime"
+	"sort"
 	"strings"
 	"syscall"
 	"time"
@@ -212,26 +214,33 @@ func daemonRun(c *cli.Context, addr string, vfsConf *vfs.Config, m meta.Meta) {
 			}
 		}
 	}
-	embeddedSchemes := []string{"sqlite3://", "badger://"}
-	for _, es := range embeddedSchemes {
-		if strings.HasPrefix(addr, es) {
-			path := addr[len(es):]
-			path2, err := filepath.Abs(path)
-			if err == nil && path2 != path {
-				for i, a := range os.Args {
-					if a == addr {
-						os.Args[i] = es + path2
-					}
-				}
-			}
-		}
-	}
+	_ = expandPathForEmbedded(addr)
 	// The default log to syslog is only in daemon mode.
 	utils.InitLoggers(!c.Bool("no-syslog"))
 	err := makeDaemon(c, vfsConf.Format.Name, vfsConf.Meta.MountPoint, m)
 	if err != nil {
 		logger.Fatalf("Failed to make daemon: %s", err)
 	}
+}
+
+func expandPathForEmbedded(addr string) string {
+	embeddedSchemes := []string{"sqlite3://", "badger://"}
+	for _, es := range embeddedSchemes {
+		if strings.HasPrefix(addr, es) {
+			path := addr[len(es):]
+			absPath, err := filepath.Abs(path)
+			if err == nil && absPath != path {
+				for i, a := range os.Args {
+					if a == addr {
+						expanded := es + absPath
+						os.Args[i] = expanded
+						return expanded
+					}
+				}
+			}
+		}
+	}
+	return addr
 }
 
 func getVfsConf(c *cli.Context, metaConf *meta.Config, format *meta.Format, chunkConf *chunk.Config) *vfs.Config {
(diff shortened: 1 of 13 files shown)


cmd/mount_unix.go
```diff
@@ -136,6 +136,10 @@ func mount_flags() []cli.Flag {
 			Name:  "enable-xattr",
 			Usage: "enable extended attributes (xattr)",
 		},
+		&cli.BoolFlag{
+			Name:  "update-fstab",
+			Usage: "add / update entry in /etc/fstab, will create soft link /sbin/mount.juicefs if not exists",
```
davies, line 141: symlink

cmd/mount.go
```diff
@@ -421,6 +555,18 @@ func mount(c *cli.Context) error {
 	}
 	logger.Infof("Data use %s", blob)
 
+	if c.Bool("update-fstab") {
```
davies, line 558: check current user is root or not

cmd/mount.go
```diff
@@ -421,6 +555,18 @@ func mount(c *cli.Context) error {
...
+		if !calledViaMount(os.Args) {
+			if err = tryToInstallMountExec(); err != nil {
+				return fmt.Errorf("error creating /sbin/mount.juicefs: %s", err)
```
davies, line 561: Don't fail the mount if any error found here

cmd/mount.go
```diff
@@ -421,6 +555,18 @@ func mount(c *cli.Context) error {
...
+			}
+		}
+		err = updateFstab(c)
+		if err != nil {
+			logger.Fatalf("failed to update fstab: %s", err)
```
davies, line 566: don't fatal here, log it as warning

cmd/mount.go
```diff
@@ -392,6 +401,131 @@ func NewReloadableStorage(format *meta.Format, reload func() (*meta.Format, erro
...
+	}
+	return nil
+}
+
+func updateFstab(c *cli.Context) error {
+	if runtime.GOOS != "linux" {
+		logger.Infof("--update-fstab is ignored in %s", runtime.GOOS)
+		return nil
+	}
+	if _, err := os.Stat("/.dockerenv"); err == nil {
```
davies, line 467: we need better way to check whether it's in container or not

cmd/mount.go
```diff
@@ -392,6 +401,131 @@ func NewReloadableStorage(format *meta.Format, reload func() (*meta.Format, erro
...
+		logger.Infoln("--update-fstab is ignored in container")
+		return nil
+	}
+	if calledViaMount(os.Args) {
```
davies, line 471: move this check outside

davies:
Could you also update the docs (command ref and how to mount on boot)

davies (changes requested):
this should not fail the mount

timfeirg:
I usually submit docs update after functionality has been confirmed, I'll work on them now.

timfeirg:
I wish to resolve https://github.com/juicedata/juicefs/issues/2504 first, and work on docs for this feature later on, in a separate PR.

cmd/mount_test.go
```diff
@@ -124,16 +130,46 @@ func umountTemp(t *testing.T) {
...
+	defer umountTemp(t)
+
+	content, err := ioutil.ReadFile(mockFstab.Name())
+	if err != nil {
+		t.Fatalf("error reading mocked fstab: %s", err)
+	}
+	rv := "redis://127.0.0.1:6379/11 /tmp/jfs-unit-test juicefs _netdev,enable-xattr,entry-cache=2,max-uploads=3,max_read=99,no-usage-report,update-fstab,writeback 0 0"
+	lv := string(content)
+	if lv != rv {
+		t.Logf("incorrect fstab entry: %s", lv)
```
davies, line 167: t.Fatalf

cmd/mount.go
```diff
@@ -421,6 +585,28 @@ func mount(c *cli.Context) error {
 	}
 	logger.Infof("Data use %s", blob)
 
+	if c.Bool("update-fstab") {
+		user, err := user.Current()
```
davies, line 589: os.Getuid()

davies (approved):

timfeirg:
failed tests passed without any code change, if things go poorly in the future, remember that current state is able to PASS.

working on docs now, no code change will be introduced.

cmd/mount.go
```diff
@@ -392,6 +401,110 @@ func NewReloadableStorage(format *meta.Format, reload func() (*meta.Format, erro
...
+	if err = scanner.Err(); err != nil {
+		logger.Warnf("scan /proc/1/mountinfo: %s", err)
+	}
+	return false
+}
+
+func updateFstab(c *cli.Context) error {
+	addr := expandPathForEmbedded(c.Args().Get(0))
+	mp := c.Args().Get(1)
+	var fstab = "/etc/fstab"
```
timfeirg, line 464: `fstab` was previously a global var, due to mocking limitations (no good way to mock a local variable in function).

after it's changed to local var, the mock here will not work, and this test will fail.

I'm looking for other ways to setup this mock test

cmd/mount.go
```diff
@@ -392,6 +401,110 @@ func NewReloadableStorage(format *meta.Format, reload func() (*meta.Format, erro
 	return holder, nil
 }
 
+func tellFstabOptions(c *cli.Context) string {
+	opts := []string{"_netdev"}
+	for _, s := range os.Args[2:] {
```
zhijian-pro, line 406: The `-d` argument cannot be used in fstab, otherwise it will fail to mount automatically.

zhijian-pro (changes requested):

cmd/mount.go
timfeirg: `-d` is already assumed in https://github.com/juicedata/juicefs/blob/main/cmd/main.go#L93, but surprisingly, the `--background` option will cause mount to crash, because `background` is classified as a FUSE option. I think this is a bug and I've changed it in [758b57c](https://github.com/juicedata/juicefs/pull/2462/commits/758b57cff1dce34b840dd4b1b1925e4907144526)

zhijian-pro (approved):

timfeirg:
code change has been introduced since review, take another look at the follow-up commits? @davies

docs/zh_cn/tutorials/qcloud.md
```diff
@@ -285,22 +285,14 @@ sudo juicefs umount /mnt/jfs
 
 ## 开机自动挂载
 
-如果你不想每次重启系统都要重新手动挂载 JuiceFS 存储，可以设置自动挂载文件系统。
-
-首先，需要将  `juicefs` 客户端重命名为 `mount.juicefs` 并复制到 `/sbin/` 目录：
-
-```shell
-sudo cp juice/juicefs /sbin/mount.juicefs
+可以用 `juiefs mount --update-fstab` 直接设置出自动挂载，例如：
```
davies, line 288: This will be available in 1.1 release, but user still read the latest docs, so we should add docs for new feature without replacing the old ways.

For these tutorials, we don't need to change them (keep them simple).

docs/en/guide/mount_at_boot.md
```diff
@@ -6,21 +6,31 @@ slug: /mount_juicefs_at_boot_time
 
 # Mount JuiceFS at Boot Time
 
-This is a guide about how to mount JuiceFS automatically at boot.
+After JuiceFS has been successfully mounted, follow this guide to setup auto-mount on boot.
 
 ## Linux
 
-Copy `juicefs` as `/sbin/mount.juicefs`, then edit `/etc/fstab` with following line:
+Using `--update-fstab` flag with the mount command should automatically set things up for you:
```
davies, line 13: This requires 1.1+

davies: Since 1.1 is not released, we should put this new feature at the end of this section.

docs/zh_cn/tutorials/qcloud.md
timfeirg: the old ways are preserved in guide/mount_at_boot.md, I can add minimum version requirements here (and remove after 1.1 has been widely adopted), would this suffice? 

I think putting "the old way" here might just be too much content, after all, there's a dedicated chapter on this topic.

timfeirg:
can I create a separate PR for docs update? review suggestions so far are all concerning release planning & version inconsistency, which won't be an issue if we update docs only after 1.1 release, am I correct? @davies

docs/zh_cn/tutorials/qcloud.md
timfeirg: > For these tutorials, we don't need to change them (keep them simple).

`--update-fstab` is the simplest way to approach auto-mount, shouldn't we recommend its usage in all tutorials? if version inconsistency is the main issue here, can I postpone this docs update until after 1.1 release?

davies:
If you prefer to have another PR, please revert the changes in docs, and we can merge the feature first.

timfeirg:
docs part has been moved to https://github.com/juicedata/juicefs/pull/2554

PR #2474, opened 2022-08-05
proofread: trash.md

rayw000:
Hi @AIXjing , could you help review this PR when you get time? Thank you!

link to: https://github.com/juicedata/docs/issues/115


docs/en/security/trash.md
```diff
@@ -7,23 +7,23 @@ sidebar_position: 2
...
+For storage system, data safety is always one of the key elements to concern. Therefore, JuiceFS enables **trash** feature by default, which will automatically move the user-removed files into a hidden directory named `.trash`, and keep the data for a period of time before cleaning.
 
 ## Configure
 
-When using `juicefs format` command to initialize JuiceFS volume, users may specify `--trash-days <val>` to configure the number of days during which files are kept in the `.trash` directory. Within this period, user-removed files are not actually purged, so you won't see decreased number in `df` output, and can still find blocks in the object storage.
+When using `juicefs format` command to initialize JuiceFS volume, users are allowed to specify `--trash-days <val>` to set the number of days which files are kept in the `.trash` directory. Within this period, user-removed files are not actually purged, so the file system usage shown in the output of `df` command will not decrease, and the blocks in the object storage will still exist in this period.
```
AIXjing, line 14:
```suggestion
When using `juicefs format` command to initialize JuiceFS volume, users are allowed to specify `--trash-days <val>` to set the number of days which files are kept in the `.trash` directory. Within this period, user-removed files are not actually purged, so the file system usage shown in the output of `df` command will not decrease, and the blocks in the object storage will still exist.
```
句首已经提到了 within this period了，所以这里应该可以不需要啦

docs/en/security/trash.md
```diff
@@ -7,23 +7,23 @@ sidebar_position: 2
...
 
-- the default value of `trash-days` is 1, which means files in trash will be automatically purged after ONE day.
+- the value of `trash-days` defaults to 1, which means files in trash will be automatically purged after ONE day.
 - use `--trash-days 0` to disable this feature; the trash will be emptied in a short time, and all files removed afterwards will be purged immediately.
-- trash is disabled for older versions. If you want to enable it for an existed volume, please upgrade **ALL** clients first and then change `--trash-days` to a positive value manually.
+- - For older versions of JuiceFS to use the recycle bin, you need to manually set `--trash-days` to the desired positive integer value via the `config` command after upgrading all mount points.
```
AIXjing, line 18:
```suggestion
- - To use the recycle bin in older versions of JuiceFS, you need to manually set `--trash-days` to the desired positive integer value via the `config` command after upgrading all mount points.
```

docs/en/security/trash.md
```diff
@@ -38,41 +38,41 @@ juicefs status META-URL
...
+There are only two levels under the tree rooted by `.trash`. The first one is a list of directories that are automatically created by JuiceFS and named as `year-month-day-hour` (e.g. `2021-11-30-10`). All files removed within an hour will be moved into the corresponding directory. The second level is just a plain list of removed files and empty directories (usually, `rm -rf <dir>` removes files in `<dir>` first, and then removes the empty `<dir>` itself). Obviously, the original tree structure is lost when files are moved into the trash. To save as much information as possible of the original hierarchy without impact on the performance, JuiceFS renames files in trash to `{parentInode-fileInode-fileName}`. Here `inode` is an internal number used for file system organizing, and can be ignored if you only care about the name of the original file.
```
AIXjing, line 45:
```suggestion
There are only two levels under the tree rooted by `.trash`. The first one is a list of directories that are automatically created by JuiceFS and named as `year-month-day-hour` (e.g. `2021-11-30-10`). All files removed within an hour will be moved into the corresponding directory. The second level is just a plain list of removed files and empty directories (the usual `rm -rf <dir>` command removes files in `<dir>` first, and then removes the empty `<dir>` itself). Obviously, the original tree structure is lost when files are moved into the trash. To save as much information about the original hierarchy as possible without impact on the performance, JuiceFS renames files in trash to `{parentInode-fileInode-fileName}`. Here `inode` is an internal number used for organizing file system, and can be ignored if you only care about the name of the original file.
```

docs/en/security/trash.md
```diff
@@ -38,41 +38,41 @@ juicefs status META-URL
...
+It is not permitted to create files inside the trash. Deleting or purging a file are forbidden as well for non-root users, even if he/she is the owner of this file.
 
 ### Recover/Purge
 
-It is suggested to ask root user to recover files, since root is allowed to move them out of trash with a single `mv` command, and causes no data copy. Other users, however, can only recover a file by reading its content and write it to another new file.
+It is suggested to recover files as root, since root is allowed to move them out of trash with a single `mv` command without any extra data copy. Other users, however, can only recover a file by reading its content and write it to another new file.
 
-JuiceFS client will check the trash every hour and purge old entries. At least one active client is required to make it happen. Like recovering, only root user is allowed to manually purge entries by `rm` command.
+Since it is JuiceFS client which is in charge of checking the trash every hour and purging old entries, you need at least 1 active client to perform checking and purging. Like recovering, only root user is allowed to manually purge entries by `rm` command.
```
AIXjing, line 65:
```suggestion
Since it is JuiceFS client which is in charge of checking the trash every hour and purging old entries, you need at least ONE active client. Like recovering, only the root user is allowed to manually purge entries by the `rm` command.
```

docs/en/security/trash.md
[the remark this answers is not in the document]
rayw000:
```suggestion
- - For older versions of JuiceFS to use the trash, you need to manually set `--trash-days` to the desired positive integer value via the `config` command after upgrading all mount points.
```

rayw000:
```suggestion
- - To use the trash in older versions of JuiceFS, you need to manually set `--trash-days` to the desired positive integer value via the `config` command after upgrading all mount points.
```

docs/en/security/trash.md
rayw000: > 

好像不太常见 the usual command 这种表达？

AIXjing: How about common `rm -rf <dir>` command? 常见的

PR #2477, opened 2022-08-06
coninue:proofread

AIXjing:
I added a few more suggestions from Section 3.2 which I left last time. :)

### DIFF
--- a/docs/en/development/data_structures.md
+++ b/docs/en/development/data_structures.md
@@ -420,7 +420,7 @@ In Redis Keys, integers (including inode numbers) are represented as decimal str
 
 ### 3.3 SQL
 
-Metadata is stored in different tables by type, and each table is named with `jfs_` followed by its specific structure name to form the table name, e.g. `jfs_node`. Some tables use `Id` of the `bigserial` type as primary keys to ensure that each table has a primary key, and the `Id` columns do not contain actual information.
+Metadata is stored in different tables by type, and each table is named with `jfs_` followed by its specific structure name to form the table name, e.g. `jfs_node`. Some tables use `Id` with the `bigserial` type as primary keys to ensure that each table has a primary key, and the `Id` columns do not contain actual information.
 
 #### 3.3.1 Setting
 
--- a/docs/en/development/data_structures.md
+++ b/docs/en/development/data_structures.md
@@ -740,7 +740,7 @@ Here the Value value is only used as a placeholder.
 
 ### 4.1 Finding files by path
 
-According to the design of [Edge](# 3.1.6-Edge), only the direct children of each directory is recorded in the metadata engine. When an application provides a path to access a file, JuiceFS needs to look it up level by level. Now suppose the application wants to open the file `/dir1/dir2/testfile`, then it needs to
+According to the design of [Edge](# 3.1.6-Edge), only the direct children of each directory are recorded in the metadata engine. When an application provides a path to access a file, JuiceFS needs to look it up level by level. Now suppose the application wants to open the file `/dir1/dir2/testfile`, then it needs to
 
 1. search for the entry with name "dir1" in the Edge structure of the root directory (inode number is fixed to 1) and get its inode number N1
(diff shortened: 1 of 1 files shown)


docs/en/development/data_structures.md
```diff
@@ -420,7 +420,7 @@ In Redis Keys, integers (including inode numbers) are represented as decimal str
 
 ### 3.3 SQL
 
-Metadata is stored in different tables by type, and each table is named with `jfs_` followed by its specific structure name to form the table name, e.g. `jfs_node`. Some tables include `Id` columns of type `bigserial` as primary keys, which are only used to ensure that each table has a primary key and do not contain actual information.
+Metadata is stored in different tables by type, and each table is named with `jfs_` followed by its specific structure name to form the table name, e.g. `jfs_node`. Some tables use `Id` of the `bigserial` type as primary keys to ensure that each table has a primary key, and the `Id` columns do not contain actual information.
```
SandyXSD, line 423:
```suggestion
Metadata is stored in different tables by type, and each table is named with `jfs_` followed by its specific structure name to form the table name, e.g. `jfs_node`. Some tables use `Id` with the `bigserial` type as primary keys to ensure that each table has a primary key, and the `Id` columns do not contain actual information.
```

docs/en/development/data_structures.md
```diff
@@ -798,17 +798,17 @@ Slice{pos: 40M, id:  0, size: 24M, off:   0, len: 24M} // can be ommited
...
-- hash is the hash value calculated from basename, which plays a certain role of isolation management
+- hash is the hash value calculated from basename, which plays a role in isolation management
 - basename is the valid name of the object in the format of `${sliceId}_${index}_${size}`, where
   - sliceId is the ID of the Slice to which the object belongs, and each Slice in JuiceFS has a globally unique ID
   - index is the index of the object in the Slice it belongs to, by default a Slice can be split into at most 16 Blocks, so its value range is [0, 16)
-  - size is the size of the Block, by default it takes the value of (0, 4 MiB]
+  - size is the size of the Block, and by default it takes the value of (0, 4 MiB]
 
-There are two hash algorithms currently in use, with the sliceId in basename as the parameter, chosen according to the [HashPrefix](#3.1.1-Setting) of the file system.
+There are two hash algorithms currently in use: with the sliceId in basename as the parameter, or chosen according to the [HashPrefix](#3.1.1-Setting) of the file system.
```
SandyXSD, line 811:
```suggestion
Currently there are two hash algorithm, both use the sliceId in basename as the parameter. The one really applied is chosen according to the [HashPrefix](#3.1.1-Setting) of the file system.
```

docs/en/development/data_structures.md
```diff
@@ -740,25 +740,25 @@ Here the Value value is only used as a placeholder.
 
 ### 4.1 Finding files by path
 
-According to the design of [Edge](# 3.1.6-Edge), only the direct children of each directory are recorded in the metadata engine. When an application provides a path to access a file, JuiceFS needs to look it up level by level. Now suppose the application wants to open the file `/dir1/dir2/testfile`, then it needs to
+According to the design of [Edge](# 3.1.6-Edge), only the direct children of each directory is recorded in the metadata engine. When an application provides a path to access a file, JuiceFS needs to look it up level by level. Now suppose the application wants to open the file `/dir1/dir2/testfile`, then it needs to
```
SandyXSD, line 743: Here the subject is 'children'.
```suggestion
According to the design of [Edge](# 3.1.6-Edge), only the direct children of each directory are recorded in the metadata engine. When an application provides a path to access a file, JuiceFS needs to look it up level by level. Now suppose the application wants to open the file `/dir1/dir2/testfile`, then it needs to
```

docs/en/development/data_structures.md
AIXjing:
```suggestion
Currently there are two hash algorithms, and both use the sliceId in basename as the parameter. Which one will be chosen to use in the application follows the [HashPrefix](#3.1.1-Setting) of the file system.
```
后半句是这个意思不？

SandyXSD (approved):

PR #2493, opened 2022-08-10
Fix jiuce sync

wangliangl:
fix juice sync bug, when not get file uid/gid


SandyXSD:
Please run `go fmt <file>` before submitting code.
Also the CLA should be signed.

PR #2502, opened 2022-08-11
cmd/sync: sync uid/gid for not existed user/group

davies:
Before this PR, the owner and group will be ignored if  the user/group is not found.

This PR will change it to use uid/gid instead.

### DIFF
--- a/pkg/object/file_unix.go
+++ b/pkg/object/file_unix.go
@@ -38,11 +38,11 @@ var mutex sync.Mutex
 func userName(uid int) string {
 	name, ok := uids[uid]
 	if !ok {
-		name = strconv.Itoa(uid)
-		if u, err := user.LookupId(name); err == nil {
+		if u, err := user.LookupId(strconv.Itoa(uid)); err == nil {
 			name = u.Username
 		} else {
 			logger.Warnf("lookup uid %d: %s", uid, err)
+			name = strconv.Itoa(uid)
 		}
 		uids[uid] = name
 	}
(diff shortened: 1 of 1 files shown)


pkg/object/file_unix.go
```diff
@@ -38,21 +38,27 @@ var mutex sync.Mutex
 func userName(uid int) string {
 	name, ok := uids[uid]
 	if !ok {
-		if u, err := user.LookupId(strconv.Itoa(uid)); err == nil {
+		name = strconv.Itoa(uid)
```
SandyXSD, line 41: better put this line in the `else` block

pkg/object/file_unix.go
```diff
@@ -38,21 +38,27 @@ var mutex sync.Mutex
...
 		}
+		uids[uid] = name
 	}
 	return name
 }
 
 func groupName(gid int) string {
 	name, ok := gids[gid]
 	if !ok {
+		name = strconv.Itoa(gid)
```
SandyXSD, line 55: same here

pkg/object/file_unix.go
```diff
@@ -81,6 +87,12 @@ func lookupUser(name string) int {
 	var uid = -1
 	if u, err := user.Lookup(name); err == nil {
 		uid, _ = strconv.Atoi(u.Uid)
+	} else {
+		if g, e := strconv.Atoi(name); e == nil {
+			uid = g
+		} else {
+			logger.Warnf("lookup group %s: %s", name, err)
```
SandyXSD, line 94: 'user' here, not 'group'

SandyXSD (approved):

PR #2582, opened 2022-08-23
chunk: cache: optimize lock type

201341:

### DIFF
--- a/pkg/chunk/mem_cache.go
+++ b/pkg/chunk/mem_cache.go
@@ -31,7 +31,7 @@ type memItem struct {
 }
 
 type memcache struct {
-	sync.Mutex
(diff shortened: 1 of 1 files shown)


pkg/chunk/mem_cache.go
```diff
@@ -31,7 +31,7 @@ type memItem struct {
 }
 
 type memcache struct {
-	sync.Mutex
+	sync.RWMutex
```
davies, line 34: RWMutex is heavier than Mutex, we should not optimize the two not-important cases, but slow down the important ones (add/load).

201341: Got it

PR #2600, opened 2022-08-25
docs on spec-limits

timfeirg:


timfeirg:
is there a number of files limit in JuiceFS CE?

davies:
There is no limit on number of files in single volume.

docs/en/reference/spec-limits.md
```diff
@@ -0,0 +1,14 @@
+# Specification Limits
+
+## File System Limits
+
+Below are theoretical limits for JuiceFS, numbers are very large so don't worry about them in practical use.
+
+* Directory tree depth: unlimited
+* File name length: 225
```
davies, line 8: 255

PR #2608, opened 2022-08-26
cmd/mount: add `root-squash` option, mapping root user to another one

### LINKED ISSUE #2419: disallow access from root
### Discussed in https://github.com/juicedata/juicefs/discussions/2405

<sup>Originally posted by **Abssenta** July 26, 2022</sup>
Hi.

The problem I am facing with JuiceFS is that the mount point is writable by root by default. We normally avoid this for security reasons.  In normal nfs sharing, it should suffice to use root_squash in the export options. However, I don't think JuiceFS has that feature. Are there any options for mounting juicefs without root write permissions?

Thanks in advance.</div>

SandyXSD:
close #2419

### DIFF
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -27,6 +27,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"sort"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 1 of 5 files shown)


SandyXSD:
This can only change the uid/gid of newly created files, but can't prevent root user to access existing ones. The reason is that JuiceFS doesn't do any permission check itself, but leaves the work to kernel (with FUSE `default_permissions` on).

PR #2611, opened 2022-08-26
CI: add version compatible test using hypothesis

zhoucheng361:


.github/scripts/testVersionCompatible.py
```diff
@@ -0,0 +1,380 @@
...
+        print('start status')
+        output = self.exec_check_output([juicefs, 'status', self.meta_url])
+        uuid = json.loads(output.decode('utf8').replace("'", '"'))['Setting']['UUID']
+        assert len(uuid) != 0
+        if self.mounted:
+            sessions = json.loads(output.decode('utf8').replace("'", '"'))['Sessions']
+            assert len(sessions) != 0 
+        print('status succeed')
+
+    @rule(juicefs=st.sampled_from(JFS_BIN))
```
davies, line 186: We only provide backwards compatibility, so old version may not mount volume formatted by new version.

davies (changes requested):
only check backward

.github/scripts/testVersionCompatible.py
zhoucheng361: ok, will check the formatted version when mount.

zhoucheng361: have add backwards check using assume.
image

sanwan (approved):

PR #2646, opened 2022-08-30
meta: allow unlink/rmdir/rename when parent attribute is not available

SandyXSD:


davies:
LGTM, please add a unit test

SandyXSD:
Use `fsck` to repair broken inodes in #2654

PR #2654, opened 2022-08-31
cmd/fsck: add `repair` option to repair broken directories

SandyXSD:


pkg/meta/redis.go
```diff
@@ -2824,6 +2824,32 @@ func (m *redisMeta) ListSlices(ctx Context, slices map[Ino][]Slice, delete bool,
 	return errno(err)
 }
 
+func (m *redisMeta) RepairInode(ctx Context, parent, inode Ino) syscall.Errno {
```
davies, line 2827:
```
Check(ctx Context, inode Ino, repair bool) syscall.Errno
```

SandyXSD: It's kind of difficult to repair an inode directly.
For example, if the inodeKey is lost, we need parent inode & type information to recover it.

cmd/fsck.go
```diff
@@ -43,18 +43,41 @@ It scans all objects in data storage and slices in metadata, comparing them to s
 lost object or broken file.
 
 Examples:
-$ juicefs fsck redis://localhost`,
+$ juicefs fsck redis://localhost
+
+# Repair broken directories
```
davies, line 48: it could be a file

pkg/meta/base.go
```diff
@@ -951,6 +953,52 @@ func (m *baseMeta) GetParents(ctx Context, inode Ino) map[Ino]int {
...
+		}
+		if attr.Full {
+			parent = inode
+			continue
+		}
+
+		// missing attribute
+		p := "/" + path.Join(ps[:i+1]...)
+		if attr.Typ != TypeDirectory { // TODO: determine file size?
+			logger.Errorf("Path %s (inode %d type %d) cannot be auto-repaired, you have to repair it manually or remove it", p, inode, attr.Typ)
```
davies, line 977: calculate the size based on chunks

davies (approved):
LGTM for directories. we can fix files with another PR.

SandyXSD:
> LGTM for directories. we can fix files with another PR.

Sure.

PR #2663, opened 2022-09-02
postgres : only one schema in search_path is allowed

zhijian-pro:
#2636


pkg/meta/sql.go
```diff
@@ -202,7 +215,7 @@ func newSQLMeta(driver, addr string, conf *Config) (Meta, error) {
 	if time.Since(start) > time.Millisecond*5 {
 		logger.Warnf("The latency to database is too high: %s", time.Since(start))
 	}
-
+	engine.SetSchema(searchPath)
```
SandyXSD, line 218: if searchPath != ""

SandyXSD (approved):

PR #2686, opened 2022-09-07
pkg/cache: ignore invalid file in cache dir

zhijian-pro:


pkg/chunk/disk_cache.go
```diff
@@ -515,6 +516,8 @@ func (cache *cacheStore) scanCached() {
 	cache.Unlock()
 }
 
+var pathReg, _ = regexp.Compile(`^chunks(\/|\\)\d+(\/|\\)\d+(\/|\\)\d+_\d+_\d+$`)
```
davies, line 519: replace \ as / before checking, then this pattern will be simpler

davies, line 519: replace \ as / before checking, then this pattern will be simpler

davies (approved):

PR #2699, opened 2022-09-08
cmd/dump: add `keep-secret-key` option

### LINKED ISSUE #2670: Introduce a `--keep-secret-key` on `dump` command

A `--keep-secret-key` command (default: `false`) to keep the secret key intact on a `dump` command.

**Why is this needed**:

Let the administrator choose between safety (the default, as `secret-key` is removed from the `dump`) and usability as the dump can be protected by other means (encrypted or else).

SandyXSD:
close #2670


davies (approved):

solracsf:
Is it normal this PR is not included in v1.0.3?

PR #2713, opened 2022-09-09
add upload delay metrics

### LINKED ISSUE #2614: More metrics for delayed blocks
In writeback mode, some blocks could be delayed to upload, we should have metrics: 1) the number of delayed blocks 2) how long they are delayed.

zhijian-pro:
close #2614

### DIFF
--- a/pkg/chunk/disk_cache.go
+++ b/pkg/chunk/disk_cache.go
@@ -579,6 +579,7 @@ type cacheManager struct {
 	cacheWriteHist  prometheus.Histogram
 	stageBlocks     prometheus.Gauge
 	stageBlockBytes prometheus.Gauge
+	stageBlockDelay prometheus.Counter
 }
 
 func keyHash(s string) uint32 {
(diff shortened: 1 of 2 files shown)


davies (approved):

pkg/chunk/disk_cache.go
```diff
@@ -693,6 +694,10 @@ func newCacheManager(config *Config, reg prometheus.Registerer, uploader func(ke
 			Name: "staging_block_bytes",
 			Help: "Total bytes of blocks in the staging path.",
 		}),
+		stageBlockDelay: prometheus.NewCounter(prometheus.CounterOpts{
+			Name: "staging_block_delay_seconds",
+			Help: "Total seconds of delay for staging blocks",
+		}),
```
qingyunha, line 700: Should this metric be a histogram?

davies: A histogram can provide more information with more metrics, but a counter may be enough here (less metrics).

PR #2749, opened 2022-09-14
cmd/format: ignore bucket schema

### LINKED ISSUE #2750: ignore bucket schema for product ease-of-use
### Discussed in https://github.com/juicedata/juicefs/discussions/2733

<sup>Originally posted by **mark-rushakoff** September 13, 2022</sup>

**What happened**:

I am using juicefs-csi-driver with Nomad, with both meta and data stores in postgres. Adding a volume was failing with a cryptic error:

> `lookup postgres on <my dns server>:53: server misbehaving`

Volume created correctly.

I was using this volume definition for Nomad:

```hcl
id = "demo-jfs"
name = "demo-jfs"

type = "csi"
plugin_id = "juicefs"

capacity_min = "1MiB"
capacity_max = "10MiB"

capability {
  access_mode     = "single-node-writer"
  attachment_mode = "file-system"
}

secrets {
  name = "timestamp-demo-jfs"
  metaurl = "postgres://juicefs:juicefs@sql-server.node.consul/juicefs_meta"
  bucket = "postgres://juicefs:juicefs@sql-server.node.consul/juicefs_data"
  storage = "postgres"
}
```

I later found out that it was incorrect to have the `postgres://` scheme on the secrets.bucket value, but I had to spend a lot of time debugging before I learned that. I eventually found that 

https://github.com/juicedata/juicefs/blob/54279914e423e0c78852e8166937c3ebf54c645f/pkg/object/sql.go#L163

could simply check `strings.HasPrefix(uri, "postgres://")` before adding that prefix to the `uri`.

Alternatively, the volume could have been rejected for already containing a scheme with `storage=postgres`. Either way would have saved me quite a bit of time.

**Environment**:
- JuiceFS version (use `juicefs --version`) or Hadoop Java SDK version: `docker.io/juicedata/juicefs-csi-driver:v0.17.0`
- Cloud provider or hardware configuration running JuiceFS: Locally with Nomad
</div>

zhijian-pro:
close #2750

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -150,11 +150,7 @@ func config(ctx *cli.Context) error {
 				quota = true
 			}
 		case "bucket":
-			new := ctx.String(flag)
-			if format.Storage == "tikv" || format.Storage == "etcd" || format.Storage == "mysql" || format.Storage == "postgres" {
-				new = strings.TrimPrefix(new, format.Storage+"://")
-			}
-			if new != format.Bucket {
+			if new := ctx.String(flag); new != format.Bucket {
 				if format.Storage == "file" {
 					if p, err := filepath.Abs(new); err == nil {
 						new = p + "/"
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -150,7 +150,11 @@ func config(ctx *cli.Context) error {
 				quota = true
 			}
 		case "bucket":
-			if new := ctx.String(flag); new != format.Bucket {
+			new := ctx.String(flag)
+			if format.Storage != "redis" {
+				new = strings.TrimPrefix(new, format.Storage+"://")
+			}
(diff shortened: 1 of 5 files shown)


SandyXSD (approved):

cmd/config.go
```diff
@@ -150,7 +150,11 @@ func config(ctx *cli.Context) error {
 				quota = true
 			}
 		case "bucket":
-			if new := ctx.String(flag); new != format.Bucket {
+			new := ctx.String(flag)
+			if format.Storage == "tikv" || format.Storage == "etcd" || format.Storage == "mysql" || format.Storage == "postgres" {
```
davies, line 154: It's better to support these scheme for these backends.

davies (approved):

Issue #2751, opened 2022-09-14
[Feature] add china unicom oss support


huzk8:
**What would you like to be added**:

**Why is this needed**:

PR #2752, opened 2022-09-14
add china unicom oss support 

huzk8:
related issue: #2751

### DIFF
--- a/pkg/object/cuos.go
+++ b/pkg/object/cuos.go
@@ -1,0 +1,86 @@
+//go:build !nos3
+// +build !nos3
+
+/*
+ * JuiceFS, Copyright 2018 Juicedata, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


SandyXSD:
Please sign the CLA.

It looks like we can use S3 storage type directly to access CUos, isn't it?

pkg/object/cuos.go
```diff
@@ -0,0 +1,86 @@
...
+		 HTTPClient:       httpClient,
+		 Credentials:      credentials.NewStaticCredentials(accessKey, secretKey, token),
+	 }
+ 
+	 ses, err := session.NewSession(awsConfig)
+	 if err != nil {
+		 return nil, fmt.Errorf("china unicom oss session: %s", err)
+	 }
+	 ses.Handlers.Build.PushFront(disableSha256Func)
+	 return &cuos{s3client{bucket, s3.New(ses), ses}}, nil
```
davies, line 80: Since it's compatible with S3, let's use `s3` as the storage, rather than introducing `cuos`.

PR #2771, opened 2022-09-16
add juicefs directory quota function

ldd91:
This pr is to add directory quota for juicefs,here is our design docs: https://docs.google.com/document/d/126-Pzijf4e-XjoslS5_kVDOvwh0LFwi_Mc96Tp74Cf8/edit#


millerlogic:
Hi, thanks for this. I am confused by the description,

> This command provides a faster way to actively build cache for the target files. It reads all objects of the files and then write them into local cache directory.

This is not what I think of when I hear the term `quota` for a file system. I think either the description needs updating, or the command should be more specific, like `cachequota` or `cquota`. Otherwise can you let me know what I'm missing?
I say this without fully knowing what this PR does, I'm just commenting my confusion. Thanks

davies:
@ldd91 We have took over this PR with #3268,  thanks for your contribution!

PR #2778, opened 2022-09-16
sync: make the special char '\' behavior consistent with aws s3

201341:
Fix s3 path s3://test.127.0.0.1/123\123/ 

### DIFF
--- a/cmd/sync.go
+++ b/cmd/sync.go
@@ -313,8 +313,16 @@ func doSync(c *cli.Context) error {
 	go func() { _ = http.ListenAndServe(fmt.Sprintf("127.0.0.1:%d", config.HTTPPort), nil) }()
 
 	// Windows support `\` and `/` as its separator, Unix only use `/`
-	srcURL := strings.Replace(c.Args().Get(0), "\\", "/", -1)
-	dstURL := strings.Replace(c.Args().Get(1), "\\", "/", -1)
+	srcURL := c.Args().Get(0)
+	dstURL := c.Args().Get(1)
+	if runtime.GOOS == "windows" {
+		if !strings.Contains(srcURL, "://") {
+			srcURL = strings.Replace(srcURL, "\\", "/", -1)
+		}
+		if !strings.Contains(dstURL, "://") {
+			dstURL = strings.Replace(dstURL, "\\", "/", -1)
+		}
+	}
 	if strings.HasSuffix(srcURL, "/") != strings.HasSuffix(dstURL, "/") {
 		logger.Fatalf("SRC and DST should both end with path separator or not!")
 	}


davies:
The URI can only accept `/`, a URL with `\` is invalid, so it's not a bug.

201341:
> The URI can only accept `/`, a URL with `\` is invalid, so it's not a bug.
> 
> for more details: https://stackoverflow.com/questions/1589930/so-what-is-the-right-direction-of-the-paths-slash-or-under-windows

s3 object support `\`, see https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html. If the object contains `\`, the juicefs sync will convert it to `/`, But the aws s3 cp/sync will keep it as `\`.

cmd/sync.go
```diff
@@ -313,8 +314,8 @@ func doSync(c *cli.Context) error {
 	go func() { _ = http.ListenAndServe(fmt.Sprintf("127.0.0.1:%d", config.HTTPPort), nil) }()
 
 	// Windows support `\` and `/` as its separator, Unix only use `/`
-	srcURL := strings.Replace(c.Args().Get(0), "\\", "/", -1)
-	dstURL := strings.Replace(c.Args().Get(1), "\\", "/", -1)
+	srcURL := c.Args().Get(0)
```
zhijian-pro, line 317:
```suggestion
        srcURL := c.Args().Get(0)
	dstURL := c.Args().Get(1)
	if runtime.GOOS == "windows" && !strings.Contains(srcURL, "://") && isFilePath(srcURL) {
		srcURL = strings.Replace(srcURL, "\\", "/", -1)
	}
	if runtime.GOOS == "windows" && !strings.Contains(dstURL, "://") && isFilePath(dstURL) {
		dstURL = strings.Replace(dstURL, "\\", "/", -1)
	}
```
Is this better ? `strings.HasSuffix(srcURL, "/") != strings.HasSuffix(dstURL, "/") ` should execute  after path transform .

201341: ok

zhijian-pro (approved):

davies:
@201341 Please update title and description to reflect the new changes.

cmd/sync.go
```diff
@@ -313,8 +313,14 @@ func doSync(c *cli.Context) error {
...
```
zhijian-pro, line 316: My mistake, I missed a case, it should be written like this.
```suggestion
        isrcURL := c.Args().Get(0)
	dstURL := c.Args().Get(1)
	if runtime.GOOS == "windows" {
		if !strings.Contains(srcURL, "://") && isFilePath(srcURL) || strings.HasPrefix(srcURL, "file:") {
			srcURL = strings.Replace(srcURL, "\\", "/", -1)
		}
		if !strings.Contains(dstURL, "://") && isFilePath(srcURL) || strings.HasPrefix(dstURL, "file:") {
			dstURL = strings.Replace(dstURL, "\\", "/", -1)
		}
	}
```
    on windows
    1. juicefs sync file:\\\C:a\b\    s3:xxxxx\
    2. juicefs sync a\b\               s3:xxxxx\

201341: hi, isFilePath has checked this case `C:\a\b\c`?

zhijian-pro: if srcURL is `file:\\\C:a\b\`, ` runtime.GOOS == "windows" && !strings.Contains(srcURL, "://") && isFilePath(srcURL) ` is false.

201341: Most users prefer to use the path C:a\b\ rather than to use these path `file:\\\C:a\b\`, And if we want to support this case, is it better to put these check on isFilePath?

zhijian-pro: Yes, we can make the `isFilePath` function support `file:\\\C:a\b\`, `C:a\b\`, `a\b`.

zhijian-pro: After discussion, we think there are only 3 legal path on windows.
1. `file:///C:/a/b/c`
2. `C:a\b\`
3. `a\b`

I'm really sorry that you repeatedly modified the code because of my problems.

201341: ok. Done

cmd/sync.go
```diff
@@ -195,11 +195,17 @@ func supportHTTPS(name, endpoint string) bool {
 // Check if uri is local file path
 func isFilePath(uri string) bool {
 	// check drive pattern when running on Windows
-	if runtime.GOOS == "windows" &&
-		len(uri) > 1 && (('a' <= uri[0] && uri[0] <= 'z') ||
-		('A' <= uri[0] && uri[0] <= 'Z')) && uri[1] == ':' {
-		return true
+	if runtime.GOOS == "windows" {
```
zhijian-pro, line 198: There's a more direct way. Recovery isFilePath and only use `strings.Contains(srcURL, "://")`  to infer whether it is a local path.

cmd/sync.go
```diff
@@ -313,8 +319,16 @@ func doSync(c *cli.Context) error {
 	go func() { _ = http.ListenAndServe(fmt.Sprintf("127.0.0.1:%d", config.HTTPPort), nil) }()
 
 	// Windows support `\` and `/` as its separator, Unix only use `/`
-	srcURL := strings.Replace(c.Args().Get(0), "\\", "/", -1)
-	dstURL := strings.Replace(c.Args().Get(1), "\\", "/", -1)
+	srcURL := c.Args().Get(0)
+	dstURL := c.Args().Get(1)
+	if runtime.GOOS == "windows" {
+		if !strings.Contains(srcURL, "://") && isFilePath(srcURL) {
```
zhijian-pro, line 325:
```suggestion
		if !strings.Contains(srcURL, "://") {
```

cmd/sync.go
```diff
@@ -313,8 +319,16 @@ func doSync(c *cli.Context) error {
...
+			srcURL = strings.Replace(srcURL, "\\", "/", -1)
+		}
+		if !strings.Contains(dstURL, "://") && isFilePath(dstURL) {
```
zhijian-pro, line 328:
```suggestion
		if !strings.Contains(dstURL, "://") {
```

201341: Hi, `strings.Contains(srcURL, "://")` can't cover this case `file:///C:a/b/c` ?

zhijian-pro:
@201341 `file:///C:a/b/c ` no need to transfrom.

davies (approved):

PR #2846, opened 2022-10-11
vfs: change the error code of ioctl for non-supported commands

SandyXSD:


pkg/vfs/vfs_unix.go
```diff
@@ -305,7 +305,7 @@ func (v *VFS) Ioctl(ctx Context, ino Ino, cmd uint32, arg uint64, bufIn, bufOut
 	defer func() { logit(ctx, "ioctl (%d,0x%X,0x%X,%v,%v): %s", ino, cmd, arg, bufIn, bufOut, strerr(err)) }()
 	switch cmd {
 	default:
-		err = syscall.ENOTTY
+		err = syscall.ENOSYS
```
davies, line 308: ENOSYS means it don't support all ioctl operations, ENOTSUP will be better?

SandyXSD: Waiting for the reporter to see if `ENOTSUP` works on his case.

SandyXSD: Neither `ENOSYS` nor `ENOTSUP` can solve reporter's issue.

davies: Then we should use ENOTSUP, otherwise the first return of ENOSYS will disable this feature.

PR #2892, opened 2022-10-21
pkg/meta: rename nextChunk to nextSlice, distinguish Slice from Chunk…

XiJinyu:
… (#2397)


SandyXSD:
The keys are stored in metadata engine, so we have to keep them intact for backward compatibility.

davies:
Unfortunately we can't change these for backword compatibility.

PR #2908, opened 2022-10-25
load/dump a gzip compressed JSON file

### LINKED ISSUE #2905: load from compressed JSON file
JuiceFS backup the metadata into object store as GZip compressed file, it's better to load these compressed file directly.

davies:
close #2905

### DIFF
--- a/cmd/dump.go
+++ b/cmd/dump.go
@@ -61,15 +61,15 @@ Details: https://juicefs.com/docs/community/metadata_dump_load`,
 
 func dump(ctx *cli.Context) (err error) {
 	setup(ctx, 1)
-	removePassword(ctx.Args().Get(0))
-	var w io.WriteCloser
+	metaUri := ctx.Args().Get(0)
 	dst := ctx.Args().Get(1)
+	removePassword(metaUri)
+	var w io.WriteCloser
 	if ctx.Args().Len() == 1 {
 		w = os.Stdout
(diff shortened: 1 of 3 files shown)


cmd/dump.go
```diff
@@ -56,27 +59,46 @@ Details: https://juicefs.com/docs/community/metadata_dump_load`,
...
+	var w io.WriteCloser
+	dst := ctx.Args().Get(1)
 	if ctx.Args().Len() == 1 {
-		fp = os.Stdout
+		w = os.Stdout
+		dst = "STDOUT"
 	} else {
-		var err error
-		fp, err = os.OpenFile(ctx.Args().Get(1), os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0644)
+		path := ctx.Args().Get(1)
```
SandyXSD, line 71: just use `dst`.

cmd/dump.go
```diff
@@ -56,27 +59,46 @@ Details: https://juicefs.com/docs/community/metadata_dump_load`,
...
+		defer func() {
+			e := fp.Close()
+			if err == nil {
+				err = e
+			}
+		}()
+		if strings.HasSuffix(path, ".gz") {
+			zw := gzip.NewWriter(fp)
+			defer func() {
+				e := zw.Flush()
```
SandyXSD, line 85: Close?

PR #2927, opened 2022-10-31
CI: random test using hdfs as object storage.

zhoucheng361:
Test Result: https://github.com/juicedata/juicefs/actions/runs/3359436839/jobs/5567461795


davies:
It's not that useful to test HDFS.

PR #2956, opened 2022-11-08
gateway: put and delete a directory

zhijian-pro:
Use whether a directory's atime equals 0 to identify if it is an existing S3 object.

### DIFF
--- a/pkg/gateway/gateway.go
+++ b/pkg/gateway/gateway.go
@@ -396,10 +396,8 @@ func (n *jfsObjects) DeleteObject(ctx context.Context, bucket, object string, op
 	}
 	info.Bucket = bucket
 	info.Name = object
-	p := n.path(bucket, object)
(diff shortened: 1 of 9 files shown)


pkg/gateway/gateway.go
```diff
@@ -371,6 +379,17 @@ func (n *jfsObjects) ListObjectsV2(ctx context.Context, bucket, prefix, continua
...
 }
 
+func (n *jfsObjects) setFileAtime(p string, atime int64) {
+	if f, eno := n.fs.Open(mctx, p, 0); eno == 0 {
+		defer f.Close(mctx)
+		if eno := f.Utime(mctx, atime, -1); eno != 0 {
+			logger.Warnf("set file atime failed: %s: %s", p, eno)
+		}
+	} else {
+		logger.Warnf("open failed: %s %s", p, eno)
```
davies, line 389: no warning if not existed

PR #2975, opened 2022-11-14
modify current limit parameters dynamically (volume dimension)

ZLEternity:
Modify juicefs client upload/download limit parameters dynamically.

Add parameters(flags) in format settings, and modify the value through the 'juicefs config' to dynamically modify the traffiic(object storage) limiting bucket setting, which acts on the volume level.

The priority of new configuration is lower than the mount parameter. 

The commit use official rate package(golang) , And allow control of timeout,  but it seems that there is no high necessity., in order to make the use easier, all case set to no timeout.

Test Result:
1. Unlimited upload/download
default 0 in format setting
image

`./juicefs bench /tmp/zmc-1112 -p 4`

image

2. Update upload/download limit(20 Mbps = 2.5MB)
`./juicefs config tikv://xxx:2379/zmc-1112 --volume-upload-limit 20 --volume-download-limit 40`
`./juicefs bench /tmp/zmc-1112 -p 4`
Setting too small, limiting(error) occurs
image
new metric:
image

3.Update upload/download limit(80 Mbps = 10MB)
`./juicefs config tikv://xxx:2379/zmc-1112 --volume-upload-limit 80 --volume-download-limit 100`
image

`./juicefs bench /tmp/zmc-1112 -p 4`
image

4. Close limit
`./juicefs config tikv://xxx:2379/zmc-1112 --volume-upload-limit 0 --volume-download-limit 0`
image

### DIFF
--- a/pkg/fuse/fuse.go
+++ b/pkg/fuse/fuse.go
@@ -459,7 +459,7 @@ func Serve(v *vfs.VFS, options string, xattrs, ioctl bool) error {
 		} else if n == "writeback_cache" || n == "writeback" {
 			opt.EnableWriteback = true
 		} else if strings.TrimSpace(n) != "" {
-			opt.Options = append(opt.Options, n)
+			opt.Options = append(opt.Options, strings.TrimSpace(n))
 		}
 	}
 	opt.Options = append(opt.Options, "default_permissions")


davies:
@ZLEternity Thanks for contributing this!

The idea is to have a default `--max-uploads` and `--max-downloads` per volume for client, and adjust them on-demand. 

To reflect reloaded config, we can register a callback to Config, rather than adding API in different components. 

For example, 
```
config.OnReload(func(f *Config){
    // use the new config
})

```

ZLEternity:
use callback method update ratelimit (exec callback in reload format setting method, the cycle is 1min) & fix test file & resolving conflicts.

retain format setting and mount ratelimit configuration, to support more case.

function test completed.

cmd/format.go
```diff
@@ -169,6 +169,16 @@ Details: https://juicefs.com/docs/community/quick_start_guide`,
 				Name:  "no-update",
 				Usage: "don't update existing volume",
 			},
+			&cli.Int64Flag{
+				Name:  "volume-upload-limit",
```
davies, line 173: We don't need this for `format` command

cmd/config.go
```diff
@@ -95,6 +95,16 @@ $ juicefs config redis://localhost --min-client-version 1.0.0 --max-client-versi
 				Name:  "force",
 				Usage: "skip sanity check and force update the configurations",
 			},
+			&cli.Int64Flag{
+				Name:  "volume-upload-limit",
```
davies, line 99: It's the default value for this volume, we should call it `default-upload-limit`, or just `upload-limit`

davies:
I'm trying to add a API for this case, please use that once merged. : https://github.com/juicedata/juicefs/pull/2994

ZLEternity:
> I'm trying to add a API for this case, please use that once merged. : #2994

OK, I will resubmit after #2994 is merged.
Add 'default-upload-limit/default-download-limit' flag in config command. If user set 'upload-limit/download-limit' in the mount command, the flag in mount command will get higher priority.
can we agree on this logic?

PR #2981, opened 2022-11-15
object/sftp: unescape username and password

zhijian-pro:


pkg/object/sftp.go
```diff
@@ -455,8 +456,10 @@ func newSftp(endpoint, username, pass, token string) (ObjectStorage, error) {
 			username = u.Username
 		}
 	}
+	username, _ = url.QueryUnescape(username)
```
davies, line 459: check the error

PR #2986, opened 2022-11-16
meta/sql: do not use the number of affected rows returned by update

zhijian-pro:
Do not use the number of affected rows returned by update to determine whether the update is successful

### DIFF
--- a/pkg/meta/sql.go
+++ b/pkg/meta/sql.go
@@ -1946,11 +1946,16 @@ func (m *dbMeta) doFindStaleSessions(limit int) ([]uint64, error) {
 
 func (m *dbMeta) doRefreshSession() {
 	_ = m.txn(func(ses *xorm.Session) error {
-		if _, err := ses.Cols("Expire").Update(&session2{Expire: m.expireTime()}, &session2{Sid: m.sid}); err != nil {
-			logger.Errorf("update session: %s", err)
-			return err
+		exist, err := ses.Exist(&session2{Sid: m.sid})
+		if exist {
+			if _, err := ses.Cols("Expire").Update(&session2{Expire: m.expireTime()}, &session2{Sid: m.sid}); err != nil {
+				logger.Errorf("update session: %s", err)
+				return err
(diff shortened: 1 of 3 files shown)


pkg/meta/sql_lock.go
```diff
@@ -211,11 +216,16 @@ func (m *dbMeta) Setlk(ctx Context, inode Ino, owner_ uint64, block bool, ltype
 			}
 			ls := updateLocks(loadLocks(locks[lkey]), lock)
 			var n int64
+			records := dumpLocks(ls)
 			if len(locks[lkey]) > 0 {
-				n, err = s.Cols("records").Update(plock{Records: dumpLocks(ls)},
-					&plock{Inode: inode, Sid: m.sid, Owner: owner})
+				if !bytes.Equal(locks[lkey], records) {
```
davies, line 221: add a  unit test to cover it

pkg/meta/sql.go
```diff
@@ -1946,9 +1946,20 @@ func (m *dbMeta) doFindStaleSessions(limit int) ([]uint64, error) {
 
 func (m *dbMeta) doRefreshSession() {
 	_ = m.txn(func(ses *xorm.Session) error {
-		n, err := ses.Cols("Expire").Update(&session2{Expire: m.expireTime()}, &session2{Sid: m.sid})
-		if err == nil && n == 0 {
-			err = fmt.Errorf("no session found matching sid: %d", m.sid)
+		var err error
+		var ok bool
+		s := session2{Sid: m.sid}
+		if ok, err = ses.ForUpdate().Get(&s); ok {
```
davies, line 1952: It's common to update it with a different value, so we should update it first and double check  it if n is 0 (uncommon)

zhijian-pro: ok

zhijian-pro:
The postgres returns 1 if the updated value is the same as the existing value，but mysql return 0.

davies (approved):

PR #2991, opened 2022-11-17
Docs: fix Redis commands doc

xiaogaozi:


zhijian-pro:
pipline should also be asked for support.

PR #2994, opened 2022-11-18
add a API to reload config

davies:


cmd/mount.go
```diff
@@ -175,18 +175,16 @@ func wrapRegister(mp, name string) (prometheus.Registerer, *prometheus.Registry)
...
-		format.Storage = c.String("storage")
+func updateFormat(c *cli.Context) func(*meta.Format) *meta.Format {
+	return func(format *meta.Format) *meta.Format {
+		if c.IsSet("bucket") {
+			format.Bucket = c.String("bucket")
+		}
+		if c.IsSet("storage") {
+			format.Storage = c.String("storage")
+		}
+		return format
```
SandyXSD, line 186: no need to return since we change the format inline

pkg/meta/config.go
```diff
@@ -66,6 +66,10 @@ type Format struct {
 	MaxClientVersion string `json:",omitempty"`
 }
 
+func (f *Format) OnReload(fn func(f *Format)) {
```
SandyXSD, line 69: not used

SandyXSD: but this may be a better place to register callbacks than `baseMeta`

cmd/gateway.go
```diff
@@ -191,9 +191,7 @@ func initForSvc(c *cli.Context, mp string, metaUrl string) (*vfs.Config, *fs.Fil
 	}
 	registerer, registry := wrapRegister(mp, format.Name)
 
-	blob, err := NewReloadableStorage(format, func() (*meta.Format, error) {
-		return getFormat(c, metaCli)
-	})
+	blob, err := NewReloadableStorage(format, metaCli, nil)
```
SandyXSD, line 194: no `updateFormat`?

SandyXSD (approved):

PR #2996, opened 2022-11-20
meta/tkv: do compaction during scan

SandyXSD:


SandyXSD:
We call `txn` during the scan process, which might cause problems. However, it seems to be supported by all TKV engines except for the MemKV.
An alternative is to do several scans with a limit for each one.

pkg/meta/tkv.go
```diff
@@ -2153,26 +2153,21 @@ func (m *kvMeta) compactChunk(inode Ino, indx uint32, force bool) {
...
-	result, err := r.scanValues(r.fmtKey("A"), -1, func(k, v []byte) bool {
-		return len(k) == klen && k[1+8] == 'C' && len(v) > sliceBytes
-	})
-	if err != nil {
+	if err := m.client.scan(m.fmtKey("A"), func(k, v []byte) {
+		if len(k) == klen && k[1+8] == 'C' && len(v) > sliceBytes {
+			inode := m.decodeInode(k[1:9])
+			indx := binary.BigEndian.Uint32(k[10:])
+			logger.Debugf("compact chunk %d:%d (%d slices)", inode, indx, len(v)/sliceBytes)
+			m.compactChunk(inode, indx, true)
```
davies, line 2164: We can do this in separate goroutine

pkg/meta/tkv_mem.go
```diff
@@ -268,8 +268,8 @@ func (c *memKV) txn(f func(kvTxn) error) error {
 }
 
 func (c *memKV) scan(prefix []byte, handler func(key []byte, value []byte)) error {
-	c.Lock()
-	defer c.Unlock()
```
davies, line 272: we can release the lock before calling handler()

Issue #2625, opened 2022-08-29
info: show locks of a certain file


SandyXSD:
**What would you like to be added**:
Improve `info` command, showing locks of the specified file.

**Why is this needed**:

PR #3001, opened 2022-11-21
info: show locks of a certain file

Hexilee:
Close https://github.com/juicedata/juicefs/issues/2625

```bash
> ./juicefs info /mnt/jfs/example
/mnt/jfs/example :
  inode: 55
  files: 1
   dirs: 0
 length: 1.56 KiB (1596 Bytes)
   size: 4.00 KiB (4096 Bytes)
   path: /mnt/jfs/example
 objects:
+------------+----------------------------+------+--------+--------+
| chunkIndex |         objectName         | size | offset | length |
+------------+----------------------------+------+--------+--------+
|          0 | myjfs/chunks/0/0/34_0_1596 | 1596 |      0 |   1596 |
+------------+----------------------------+------+--------+--------+
  flocks:
+-----+---------------------+------+
| Sid |        Owner        | Type |
+-----+---------------------+------+
| 49  | 4848886172199279596 |    W |
+-----+---------------------+------+
```

### DIFF
--- a/cmd/info.go
+++ b/cmd/info.go
@@ -118,10 +118,6 @@ func info(ctx *cli.Context) error {
 		if err != nil {
 			logger.Fatalf("write message: %s", err)
 		}
-		err = f.Sync()
-		if err != nil {
-			logger.Fatalf("sync control file: %s", err)
-		}
 		var resp vfs.InfoResponse
 		err = resp.Decode(f)
 		_ = f.Close()
--- a/cmd/info.go
+++ b/cmd/info.go
@@ -174,7 +174,7 @@ func info(ctx *cli.Context) error {
 			printResult(results, 1, false)
 		}
 		if len(resp.FLocks) > 0 {
-			fmt.Println("  flocks:")
+			fmt.Println(" flocks:")
 			results := make([][]string, 0, 1+len(resp.FLocks))
 			results = append(results, []string{"Sid", "Owner", "Type"})
 			for _, l := range resp.FLocks {
@@ -187,7 +187,7 @@ func info(ctx *cli.Context) error {
 			printResult(results, 0, false)
 		}
 		if len(resp.PLocks) > 0 {
-			fmt.Println("  plocks:")
+			fmt.Println(" plocks:")
 			results := make([][]string, 0, 1+len(resp.PLocks))
 			results = append(results, []string{"Sid", "Owner", "Type", "Pid", "Start", "End"})
 			for _, l := range resp.PLocks {
--- a/cmd/info.go
+++ b/cmd/info.go
@@ -17,13 +17,11 @@
 package cmd
 
 import (
-	"bytes"
 	"fmt"
 	"os"
 	"path/filepath"
 	"runtime"
 	"strconv"
-	"strings"
 
 	"github.com/juicedata/juicefs/pkg/meta"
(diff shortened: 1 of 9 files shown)


pkg/meta/redis_lock.go
```diff
@@ -206,3 +207,38 @@ func (r *redisMeta) Setlk(ctx Context, inode Ino, owner uint64, block bool, ltyp
 	}
 	return errno(err)
 }
+
+func (r *redisMeta) ListLocks(ctx context.Context, inode Ino) ([]PLockItem, []FLockItem, error) {
```
davies, line 211: Add a unit test for these

cmd/info.go
```diff
@@ -117,42 +118,169 @@ func info(ctx *cli.Context) error {
 		if err != nil {
 			logger.Fatalf("write message: %s", err)
 		}
-		data := make([]byte, 4)
-		n := readControl(f, data)
-		if n == 1 && data[0] == byte(syscall.EINVAL&0xff) {
-			logger.Fatalf("info is not supported, please upgrade and mount again")
+		err = f.Sync()
```
SandyXSD, line 121: Why we need this?

cmd/info.go
```diff
@@ -117,42 +118,169 @@ func info(ctx *cli.Context) error {
...
+		if err != nil {
+			logger.Fatalf("sync control file: %s", err)
+		}
+		var resp vfs.InfoResponse
+		err = resp.Decode(f)
+		_ = f.Close()
+		if err == syscall.EINVAL {
+			legacyInfo(d, path, inode, recursive, raw)
```
SandyXSD, line 129: use `legacyInfo` for the following args.

pkg/vfs/internal.go
```diff
@@ -247,6 +249,59 @@ func (v *VFS) caclObjects(id uint64, size, offset, length uint32) []*obj {
...
+}
+
+type chunkObj struct {
+	ChunkIndex     uint64
+	Key            string
+	Size, Off, Len uint32
+}
+
+func (r *InfoResponse) Encode() []byte {
+	resp, _ := json.MarshalIndent(r, "", "  ")
```
SandyXSD, line 276: `Marshal` should be enough if it's not for printing.

cmd/info.go
Hexilee: To ensure the request data being flushed into FUSE.

pkg/meta/redis_lock.go
Hexilee: Added and passed.

pkg/vfs/internal.go
Hexilee: Ok, we use `Marshal` now.

cmd/info.go
Hexilee: I'm sorry to not understand your change requests. Should I reuse some code?

cmd/info.go
SandyXSD: It's not needed before. Have you met any problem if there's no `sync`?

Hexilee: I have suspected it result in a problem, but I found the true reason later. I'm going to remove this and run tests.

Hexilee: Removed now.

PR #3015, opened 2022-11-24
remove duplicated docs on JuiceFS CSI Driver

timfeirg:
CSI Driver docs has been refurbished, lets maintain csi docs in one place.

the "Mount JuiceFS in the container" part will be re-introduced in CSI docs later.


zwwhdls:
I suggest to keep CSI usage and hostPath usage in this doc and keep a link to CSI docs. About `Mount JuiceFS in the container`, I don't know if it's a common usage in Kubernetes. What do you think? @timfeirg

timfeirg:
I wish all Kubernetes related usage should reside in CSI documentation, including practices like "Mount JuiceFS in the container" or hostPath, although they aren't really under the CSI realm, they are close related to Kubernetes. I want to keep all Kubernetes topics under CSI docs to reduce maintenance.

After that, we can leave references in CE/EE docs so that users don't get lost.

davies:
K8S is just one way of using containers, let's replace the CSI part as a link and keep others here.

timfeirg:
"Mount JuiceFS in the container" is preserved, this pr only deletes CSI tutorials.

PR #3021, opened 2022-11-25
hadoop: add pitest

tangyoupeng:
mvn test-compile org.pitest:pitest-maven:mutationCoverage


sdk/java/src/main/java/io/juicefs/JuiceFileSystemImpl.java
```diff
@@ -1037,6 +1059,12 @@ public void hsync() throws IOException {
       ((FSOutputStream) out).fsync();
     }
 
+    @Override
+    public void close() throws IOException {
```
davies, line 1063: synchronized

PR #3026, opened 2022-11-27
modify current limit parameters dynamically (volume dimension)

ZLEternity:
Modify juicefs client upload/download limit parameters dynamically.

Add parameters(flags) in format settings, and modify the value through the 'juicefs config' to dynamically modify the traffiic(object storage) limiting bucket setting, which acts on the volume level.

The priority of new configuration is lower than the mount parameter.

The commit use official rate package(golang) , And allow control of timeout, but it seems that there is no high necessity., in order to make the use easier, all case set to no timeout.

Test Result:

1. Unlimited upload/download
default 0 in format setting

`./juicefs bench /tmp/zmc-1112 -p 4`
image

2. Update upload/download limit(480 Mbps = 60MB, 960 Mbps = 120 MB)
`./juicefs config tikv://xxx:2379/zmc-1112 --default-upload-limit 480 --default-download-limit 960`
image

3. Update upload/download limit
`./juicefs config tikv://1xxx:2379/zmc-1112 --default-upload-limit 480 --default-download-limit 480`
image

4. Close upload/download limit
`./juicefs config tikv://xxx:2379/zmc-1112 --default-upload-limit 0 --default-download-limit 0` 
image

### DIFF
--- a/pkg/chunk/cached_store.go
+++ b/pkg/chunk/cached_store.go
@@ -795,25 +795,25 @@ func NewCachedStore(storage object.ObjectStorage, config Config, reg prometheus.
 func (store *cachedStore) UpdateCachedStoreRateLimit(limit int64, lastLimit int64, limitType string) {
 	switch limitType {
 	case "upLimit":
-		if (store.upLimit == nil || lastLimit == 0) && limit > 0 {
+		if store.upLimit == nil && limit > 0 {
 			store.upLimit = rate.NewLimiter(rate.Limit(float64(limit)*0.85), int(limit))
 		} else if store.upLimit != nil && limit != lastLimit {
-			if lastLimit > 0 && limit > 0 {
+			if limit > 0 {
 				store.upLimit.SetLimit(rate.Limit(float64(limit) * 0.85))
 				store.upLimit.SetBurst(int(limit))
 			} else if lastLimit > 0 && limit == 0 {
-				store.upLimit = nil
+				store.upLimit.SetLimit(rate.Inf)
 			}
 		}
 	case "downLimit":
-		if (store.downLimit == nil || lastLimit == 0) && limit > 0 {
+		if store.downLimit == nil && limit > 0 {
 			store.downLimit = rate.NewLimiter(rate.Limit(float64(limit)*0.85), int(limit))
 		} else if store.downLimit != nil && limit != lastLimit {
-			if lastLimit > 0 && limit > 0 {
+			if limit > 0 {
 				store.downLimit.SetLimit(rate.Limit(float64(limit) * 0.85))
 				store.downLimit.SetBurst(int(limit))
 			} else if lastLimit > 0 && limit == 0 {
-				store.downLimit = nil
+				store.downLimit.SetLimit(rate.Inf)
 			}
 		}
 	default:
--- a/pkg/chunk/cached_store.go
+++ b/pkg/chunk/cached_store.go
@@ -42,8 +42,6 @@ const SlowRequest = time.Second * time.Duration(10)
 
 var (
 	logger = utils.GetLogger("juicefs")
-	UseMountUploadLimitConf = false
-	UseMountDownloadLimitConf = false
 )
 
 type pendingItem struct {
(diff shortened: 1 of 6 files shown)


pkg/chunk/cached_store.go
```diff
@@ -42,6 +42,8 @@ const SlowRequest = time.Second * time.Duration(10)
 
 var (
 	logger = utils.GetLogger("juicefs")
+	UseMountUploadLimitConf = false
```
davies, line 45: It's better to add a `Reload()` interface for `chunk.Config` and hide these details inside chunk/

Because we may want to reload other fields of chunk.Config, and we don't add these for them.

ZLEternity:
hide set conf logic inside chunk module.
But UseMountUploadLimitConf/UseMountDownloadLimitConf is only set once during mount, and will not be updated. It seems that reload/callback is not need.

pkg/chunk/cached_store.go
```diff
@@ -149,7 +149,11 @@ func (s *rSlice) ReadAt(ctx context.Context, page *Page, off int) (n int, err er
 
 	if s.store.seekable && boff > 0 && len(p) <= blockSize/4 {
 		if s.store.downLimit != nil {
-			s.store.downLimit.Wait(int64(len(p)))
+			err = s.store.downLimit.WaitN(context.TODO(), len(p))
+			if err != nil {
+				s.store.getTokenFromBucketErrors.WithLabelValues("download").Add(1)
+				return 0, fmt.Errorf("method: ReadAt, key = %s, get token from tokenBucket, need token count (page) = %d, error : %s", key, len(p), err)
```
davies, line 155: we should not fail the read() if waitN fail

pkg/chunk/cached_store.go
```diff
@@ -337,7 +341,11 @@ func (s *wSlice) WriteAt(p []byte, off int64) (n int, err error) {
 
 func (store *cachedStore) put(key string, p *Page) error {
 	if store.upLimit != nil {
-		store.upLimit.Wait(int64(len(p.Data)))
+		var err = store.upLimit.WaitN(context.TODO(), len(p.Data))
+		if err != nil {
+			store.getTokenFromBucketErrors.WithLabelValues("upload").Add(1)
+			return fmt.Errorf("method: put, key = %s, get token from tokenBucket, need token count (page.Data) = %d, error : %s", key, len(p.Data), err)
```
davies, line 347: same here

pkg/chunk/cached_store.go
```diff
@@ -589,17 +609,18 @@ type cachedStore struct {
...
+
+	cacheHits                prometheus.Counter
+	cacheMiss                prometheus.Counter
+	cacheHitBytes            prometheus.Counter
+	cacheMissBytes           prometheus.Counter
+	cacheReadHist            prometheus.Histogram
+	objectReqsHistogram      *prometheus.HistogramVec
+	objectReqErrors          prometheus.Counter
+	objectDataBytes          *prometheus.CounterVec
+	getTokenFromBucketErrors *prometheus.CounterVec
```
davies, line 623: it's better to have metric for the wait period, other than errors

pkg/chunk/cached_store.go
```diff
@@ -613,7 +634,11 @@ func (store *cachedStore) load(key string, page *Page, cache bool, forceCache bo
 	compressed := needed > len(page.Data)
 	// we don't know the actual size for compressed block
 	if store.downLimit != nil && !compressed {
-		store.downLimit.Wait(int64(len(page.Data)))
+		err = store.downLimit.WaitN(context.TODO(), len(page.Data))
```
davies, line 637: Why we change to use golang.org/x/time/rate? When it will return error?

pkg/chunk/cached_store.go
```diff
@@ -651,14 +676,24 @@ func (store *cachedStore) load(key string, page *Page, cache bool, forceCache bo
...
-		store.downLimit.Wait(int64(n))
-	}
 	store.objectDataBytes.WithLabelValues("GET").Add(float64(n))
 	store.objectReqsHistogram.WithLabelValues("GET").Observe(used.Seconds())
 	if err != nil {
 		store.objectReqErrors.Add(1)
+		logger.Error("get %s: %s", key, err)
+	}
+	var err1 error
+	if store.downLimit != nil && compressed {
```
davies, line 686: there is a race between UpdateCachedStoreRateLimit and here.

ZLEternity: 1. fixbug rate
2. golang.org/x/time/rate allows update limit or burst, can ensure concurrency.
3. It returns an error if n exceeds the Limiter's burst size, the Context is canceled, or the expected wait time exceeds the Context's Deadline.
cancel and timeout are not set here, so only n(token count) is larger than burst size, can cause error. To ensure the effect of rate.limiter, this case should not be executed normally.
4. about metrics, make tokens at a constant speed(rate.NewLimiter), we may prefer to know whether there are config exception. if it is normal, block until get the needed tokens.

davies:
This PR is took over by #3574 , thanks!

PR #3061, opened 2022-12-08
support idle handles for directory

### LINKED ISSUE #2634: Improve performance for readdir when exported via NFS
nfsd will call opendir/releasedir for each `readdir(off)`, and JuiceFS will list call meta.Readdir() for each `opendir`, it will be very slow for large directory (O(N*N)).

We should have a cache for directory, similar to cache of chunks, they will be invalided when mtime of directory is changed.

Hexilee:

Close #2634

### DIFF
--- a/pkg/vfs/handle.go
+++ b/pkg/vfs/handle.go
@@ -34,7 +34,7 @@ type handle struct {
 	// for dir
 	children   []*meta.Entry
 	readAt     time.Time
-	releasedAt *time.Time
+	releasedAt time.Time
 
 	// for file
 	locks      uint8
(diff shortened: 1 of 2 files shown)


SandyXSD:
We need a simple test report to know how much this PR improves the situation.
Try `ls` a huge directory (e.g. 1m entries) in nfs client, and compare the time costs.

pkg/vfs/handle.go
```diff
@@ -148,6 +150,29 @@ func (h *handle) Close() {
...
 	v.hanleM.Lock()
 	defer v.hanleM.Unlock()
+	for len(v.idleHandles[inode]) > 0 {
+		length := len(v.idleHandles[inode])
+		h := v.idleHandles[inode][length-1]
+		attr := &meta.Attr{}
+		err := func() syscall.Errno {
+			ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+			defer cancel()
+			return v.Meta.GetAttr(meta.WrapContext(ctx), inode, attr)
```
davies, line 160: this blocks v.hanleM

davies:
We should speedup Meta.Readdir, and cached the result (children) in meta.OpenFiles (same to chunks)

davies (changes requested):
We should speedup Meta.Readdir, and cached the result (children) in meta.OpenFiles (same to chunks)

pkg/vfs/handle.go
Hexilee: Fixed

PR #3070, opened 2022-12-12
vfs: return EBADF if handle is not opened for read or write

guaneryu:
If fd isn't opened for read or write, EBADF should be returned according to read(2) and write(2).

Signed-off-by: Eryu Guan <guaneryu@gmail.com>

### DIFF
--- a/pkg/vfs/vfs.go
+++ b/pkg/vfs/vfs.go
@@ -547,7 +547,7 @@ func (v *VFS) Read(ctx Context, ino Ino, buf []byte, off uint64, fh uint64) (n i
 		return
 	}
 	if h.reader == nil {
-		err = syscall.EACCES
+		err = syscall.EBADF
 		return
 	}
 	if !h.Rlock(ctx) {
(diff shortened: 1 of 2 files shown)


guaneryu:
Currently fuse kernel will return EBADF directly to applications, no READ or WRITE request will be sent to juicefs client, so we still see EBADF errno, but it's better to follow the behavior described in manpage in code.

SandyXSD:
`fallocate` should be modified as well.

guaneryu:
> `fallocate` should be modified as well.

Indeed, fixed.

davies (approved):

SandyXSD:
```
    vfs_test.go:251: write off too big: bad file descriptor
--- FAIL: TestVFSIO (0.01s)
```
@guaneryu Could you please fix the related unit tests as well?

guaneryu:
> ```
>     vfs_test.go:251: write off too big: bad file descriptor
> --- FAIL: TestVFSIO (0.01s)
> ```
> 
> @guaneryu Could you please fix the related unit tests as well?

Fixed.

PR #3078, opened 2022-12-13
cache results of readdir in openfiles

Hexilee:

Close https://github.com/juicedata/juicefs/issues/2634, #3061

### Benchmark

- Meta engine

  Redis

- Prepare huge directory

  `> juicefs mdtest --threads=10 --files=100000 --depth=0 redis://localhost /mdtest`

- Export by nfs
- Mount as nfs
- Workload

  `> time ls > /dev/null`

#### Result
- This branch
```bash
hexi@Arch /m/j/m/t/mdtest_tree.0> time ls > /dev/null

________________________________________________________
Executed in    7.23 secs    fish           external
   usr time    5.45 secs    1.96 millis    5.44 secs
   sys time    0.24 secs    0.00 millis    0.24 secs

```
- Main branch
  infinity


pkg/meta/base.go
```diff
@@ -935,9 +935,26 @@ func (m *baseMeta) Close(ctx Context, inode Ino) syscall.Errno {
...
-	var attr Attr
-	if err := m.GetAttr(ctx, inode, &attr); err != 0 {
-		return err
+	attr := &Attr{}
+	if m.conf.OpenCache == 0 || !m.of.OpenCheck(inode, attr) {
+		err := m.GetAttr(ctx, inode, attr)
+		if err != 0 {
+			return err
+		}
+		if m.conf.OpenCache > 0 {
```
davies, line 944: we should cache the result even OpenCache is zero (timeout)

Hexilee: Ok, changed.

davies (approved):

Hexilee:
This PR cannot accomplish our goal because of some problems:
- The attributes of directories are not precise enough, we should check the mtime of all children to ensure their attributes are not outdated
- The children of directores contains their parents, we should also validate all child directories when their parents change.

So, we currently cannot support open cache for `readdirplus`.

PR #3082, opened 2022-12-14
cmd/mount: check log flag when mount by foreground

zhijian-pro:


cmd/mount.go
```diff
@@ -615,6 +615,9 @@ func mount(c *cli.Context) error {
 	if c.Bool("background") && os.Getenv("JFS_FOREGROUND") == "" {
 		daemonRun(c, addr, vfsConf, metaCli)
 	} else {
+		if c.IsSet("log") {
+			logger.Warnf("--log flag is ignored in foreground mode,the juicefs output will be Stdout")
```
SandyXSD, line 619: "--log flag is ignored in foreground mode, the log output will be Stderr"

SandyXSD (approved):

PR #3102, opened 2022-12-19
CI: check "database is locked" in gateway test.

zhoucheng361:
the sqlite3 "database is locked" error is fixed by PR: https://github.com/juicedata/juicefs/pull/3046. should check it in CI.

### DIFF
--- a/.github/workflows/gateway.yml
+++ b/.github/workflows/gateway.yml
@@ -14,6 +14,8 @@ on:
     paths-ignore:
       - 'docs/**'
       - '**.md'
+  schedule:
+    - cron:  '0 19 * * *'
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


.github/workflows/gateway.yml
```diff
@@ -196,6 +198,7 @@ jobs:
           if [ -f ~/.juicefs/juicefs.log ]; then
             tail -300 ~/.juicefs/juicefs.log
             grep "<FATAL>:" ~/.juicefs/juicefs.log && exit 1 || true
+            grep "database is locked" ~/.juicefs/juicefs.log && exit 1 || true
```
davies, line 201: it's OK to have some "database is locked" as warning

PR #3121, opened 2022-12-23
Add internal-inodes-prefix option

### LINKED ISSUE #3120: Disable/change virtual file names

Please add an option to either rename the built-in virtual files (`.accesslog`, `.config`, etc.), or to disable them altogether when mounting the file-system.

**Why is this needed**:

I'm attempting to use JuiceFS as a volume backend for my docker containers. However, some of them (such as MongoDB) attempt to create a `.config` folder at the root of the volume, which conflicts with the `.config` virtual file (ironically, this file was meant to be named `.jfsconfig` initially, avoiding this issue - but was renamed to simply `.config` before PR #652 was finalized).

aronsky:
A proposed change to fix #3120. Implements a new string flag to `juicefs mount`, named `--internal-inodes-prefix`. If provided, it will be prepended to the names of all the internal inodes.

Regular mounting (without this flag, identical to current behavior):

```
$ juicefs mount -d redis://127.0.0.1/1 /mnt/jfs
$ ls -la /mnt/jfs
total 10
drwxrwxrwx    2 root     root          4096 Dec 23 08:42 .
drwxr-xr-x    4 root     root          4096 Dec 21 10:53 ..
-r--------    1 user     user             0 Dec 23 11:19 .accesslog
-r--------    1 user     user          1362 Dec 23 11:19 .config
-rw-r--r--    1 root     root             0 Dec 23 08:56 .juicefs
-r--r--r--    1 user     user             0 Dec 23 11:19 .stats
```

With this proposed flag:

```
$ juicefs mount -d --internal-inodes-prefix .jfs redis://127.0.0.1/1 /mnt/jfs
$ ls -la /mnt/jfs
total 10
drwxrwxrwx    2 root     root          4096 Dec 23 08:42 .
drwxr-xr-x    4 root     root          4096 Dec 21 10:53 ..
-r--------    1 user     user             0 Dec 23 11:19 .jfs.accesslog
-r--------    1 user     user          1362 Dec 23 11:19 .jfs.config
-rw-r--r--    1 root     root             0 Dec 23 08:56 .juicefs
-r--r--r--    1 user     user             0 Dec 23 11:19 .jfs.stats
```

I haven't added any unit-tests, as I'm not thoroughly familiar with either Go, or this code base. If you'll be so kind as to direct me to the relevant files, I'll be happy to oblige.

Furthermore, I only updated the English documentation, as I don't speak Chinese.

I'm aware that this solution (changing the inode names after they are initialized) is a bit hacky, but I don't see any other approach that allows using command flags without fundamentally changing the VFS logic. Also, the `.trash` node is modified in a similar fashion, so I guess it's not too bad.

Finally, my modifications are localized to the `mount` command, since this is the command I'm familiar with and use personally. It's possible that it would be relevant to additional commands - help in that regard is appreciated.

### DIFF
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -360,6 +360,7 @@ func getMetaConf(c *cli.Context, mp string, readOnly bool) *meta.Config {
 		Heartbeat:  duration(c.String("heartbeat")),
 		MountPoint: mp,
 		Subdir:     c.String("subdir"),
+		IntPrefix:  c.String("internal-inodes-prefix"),
 	}
 	if cfg.Heartbeat < time.Second {
 		logger.Warnf("heartbeat should not be less than 1 second")
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -370,10 +370,6 @@ func getMetaConf(c *cli.Context, mp string, readOnly bool) *meta.Config {
 		logger.Warnf("heartbeat shouldd not be greater than 10 minutes")
 		cfg.Heartbeat = time.Minute * 10
 	}
-	if cfg.IntPrefix != "" && cfg.IntPrefix[0] != '.' { // todo: verify the prefix is alpha-numeric
-		logger.Warnf("internal-inodes-prefix should start with a dot (.)")
-		cfg.IntPrefix = "." + cfg.IntPrefix
-	}
 	return cfg
 }
 
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -370,6 +370,10 @@ func getMetaConf(c *cli.Context, mp string, readOnly bool) *meta.Config {
 		logger.Warnf("heartbeat shouldd not be greater than 10 minutes")
 		cfg.Heartbeat = time.Minute * 10
 	}
+	if cfg.IntPrefix != "" && cfg.IntPrefix[0] != '.' { // todo: verify the prefix is alpha-numeric
+		logger.Warnf("internal-inodes-prefix should start with a dot (.)")
+		cfg.IntPrefix = "." + cfg.IntPrefix
+	}
 	return cfg
 }
 
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -245,12 +245,13 @@ func expandPathForEmbedded(addr string) string {
 
 func getVfsConf(c *cli.Context, metaConf *meta.Config, format *meta.Format, chunkConf *chunk.Config) *vfs.Config {
 	cfg := &vfs.Config{
-		Meta:       metaConf,
-		Format:     format,
-		Version:    version.Version(),
-		Chunk:      chunkConf,
-		BackupMeta: duration(c.String("backup-meta")),
-		Port:       &vfs.Port{DebugAgent: debugAgent, PyroscopeAddr: c.String("pyroscope")},
+		Meta:                metaConf,
+		Format:              format,
+		Version:             version.Version(),
+		Chunk:               chunkConf,
+		BackupMeta:          duration(c.String("backup-meta")),
+		Port:                &vfs.Port{DebugAgent: debugAgent, PyroscopeAddr: c.String("pyroscope")},
+		InternalInodePrefix: c.String("internal-inodes-prefix"),
 	}
 	if cfg.BackupMeta > 0 && cfg.BackupMeta < time.Minute*5 {
 		logger.Fatalf("backup-meta should not be less than 5 minutes: %s", cfg.BackupMeta)
@@ -360,7 +361,6 @@ func getMetaConf(c *cli.Context, mp string, readOnly bool) *meta.Config {
 		Heartbeat:  duration(c.String("heartbeat")),
 		MountPoint: mp,
 		Subdir:     c.String("subdir"),
-		IntPrefix:  c.String("internal-inodes-prefix"),
 	}
 	if cfg.Heartbeat < time.Second {
 		logger.Warnf("heartbeat should not be less than 1 second")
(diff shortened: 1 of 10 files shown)


aronsky:
After taking a deeper dive into the code, I made the following changes:

1. Added a relevant unit-test (basically, `TestInternalFile` is run with 2 different cases, once when the prefix is empty, and once when it's `.jfs`).
2. Added a rudimentary validation to make sure the prefix starts with a dot (`.`), if it isn't empty. This is because that's a requirement for internal files, per `IsSpecialName` function used in the unit-test.

The above test should probably be more robust - e.g., slashes should not be allowed. Another question I wonder about - where should these checks be performed? Feels to me that the responsibility might be on the VFS constructor, as well?

aronsky:
OK, not sure what are those `mutate-test` failures - will try to figure it out, but appreciate some pointers in the meantime.

aronsky:
After reading the explanation of mutation testing (https://github.com/juicedata/juicefs/blob/main/.github/scripts/mutate/how_to_use_mutate_test.md), it seems that my changes are not relevant to those failures, and the unit tests in the `vfs` package simply don't cover enough cases (similar failures happened in another PR several weeks ago that touched `vfs.go`, and was approved).

While this can be improved, I believe it's not part of this PR, so I will refrain from making changes attempting to fix that, for now.

Would appreciate a review of my changes.

SandyXSD:
> 

Yes, for now the mutate tests are not compulsory, but serve more like hints. Sorry for the confusion.

pkg/meta/config.go
```diff
@@ -41,6 +42,7 @@ type Config struct {
 	Heartbeat   time.Duration
 	MountPoint  string
 	Subdir      string
+	IntPrefix   string // prefix for internal inodes (.accesslog, .config, etc.)
```
davies, line 45: move this into vfs.Config

davies (changes requested):
The patch sounds good overall, there is a few things to consider: 

Other commands need to access these internal inodes, they should work even this flag is used. So we should provide a way to disover the prefix (via .config?).

pkg/meta/config.go
aronsky: Done.

aronsky:
> The patch sounds good overall, there is a few things to consider:
> 
> Other commands need to access these internal inodes, they should work even this flag is used. So we should provide a way to disover the prefix (via .config?).

I can expose it via `.config` easily, but it's a chicken and egg problem - without knowing the name of `.config`, how can those commands read the prefix? It will solve the issue for commands accessing other internal files (such as `.accesslog` and `.stats`), but commands that seem to use `.config` will be left in the dark.

From what I can see, most of the usage of `.config` is by scripts and debugging utilities, which would run with the default prefix value of none, so it shouldn't pose a problem. Correct me if I'm wrong.

The only place where `.config` seems to be relevant to normal JuiceFS functionality is inside `prepareMp()` (in `cmd/mount.go`):

```go
contents, err := os.ReadFile(path.Join(mp, ".config"))
if err != nil {
	// failed to read juicefs config, continue to mount
	return
}
```

I can add a new function to `internal.go` that takes an unprefixed internal inode name, and returns it prefixed - then this function can be used in the above code to find the correct `.config` file. Sounds good?

davies:
> The only place where `.config` seems to be relevant to normal JuiceFS functionality is inside `prepareMp()` (in `cmd/mount.go`):
> 
> ```go
> contents, err := os.ReadFile(path.Join(mp, ".config"))
> if err != nil {
> 	// failed to read juicefs config, continue to mount
> 	return
> }
> ```
> 
> The problem here is that we don't have access to the existing mount point's VFS object, so we can't check for the correct prefix. I can list the inodes under the root of the mount point, and find the `.config` inode by its number - will that work?

That could work.

aronsky:
The latest commit hasn't been tested yet. There are no unit-tests for `prepareMp`, and I haven't had a chance to verify how that logic worked prior to my changes, and whether it remains functional.

Once I test it, I'll update the results (or, better yet, if I manage to add a reasonable unit test, I'll commit that).

aronsky:
OK, functionally, the latest commit works as expected. Adding a unit test will be complicated, since it's basically an internal function (`prepareMp`) that's never tested. And the function that's actually tested (`mount`) returns `nil` when the mounting should be ignored (the correct behavior), and 0 if mounted successfully (the incorrect behavior), so differentiating based on that is not possible.

Should we merge it as-is, or do you have an idea for a proper unit test?

cmd/mount.go
```diff
@@ -329,7 +330,22 @@ func prepareMp(newCfg *vfs.Config, mp string) (ignore bool) {
 		return
 	}
 
-	contents, err := os.ReadFile(path.Join(mp, ".config"))
+	prefixedConfigName := ".config"
+
+	entries, _ := os.ReadDir(mp)
```
davies, line 335: We should have a fast path for those don't have a special prefix

davies:
@aronsky It's OK to not have a unit test for it.

Can you continue to fix all others commands that access internal files?

cmd/mount.go
aronsky: Best we can do is try getting the Inode of `.config`:

```go
ino, _ := utils.GetFileInode(path.Join(mp, ".config"))
```

and check whether it is a special node.

We can't skip the loop altogether and just read the contents of `.config`, as that would defy the initial purpose of this PR (i.e., the case when the file `.config` exists and clashes with JuiceFS `.config`).

davies: Yes, we can check the inode of `.config`.

aronsky:
Sorry, was busy with work and other things lately - will try to get back and finish this in the coming days.

PR #3157, opened 2023-01-06
chunk: fix the issue that staging file may be uploaded for many times

### LINKED ISSUE #3143: Writeback cache synchronization issue
**What happened**:
JuiceFS appears to have a writeback cache synchronization issue.  Example below may be exacerbated by slow IO and intolerance for misbehavior of the storage provider (STORJ), but it seems to me that the root cause is within JuiceFS.

See chunk 11008_0_4194304 in attached log; several parallel upload attempts (resulting in server errors), continuing to try to upload after file has been deleted (at time 21:32:04), and other general signs of missing synchronization.  Also appears (not observable in log below) as though --upload-delay is not being fully obeyed (uploads begin within about one second of file creation). This example is very repeatable. This problem does not occur without --writeback in the mount command.

One concurrent upload attempt per chunk, and no attempt to continue uploading significantly after chunk has been deleted locally.

Mount command: juicefs mount --no-usage-report --cache-size 512 --writeback -o allow_other --upload-delay 5 --backup-meta 0 --max-uploads 5 --verbose sqlite3://storj-test.db /jfs

Test scenario: create 10 4MB files in rapid succession:
    for i in {1..10}; do dd if=/dev/urandom bs=1M count=4 of=./test${i}; done

**Environment**:
- JuiceFS version: juicefs version 1.0.3+2022-12-27.ec6c8abd
- OS: Linux (AlmaLinux 8.7)
- Kernel: 4.18.0-425.3.1.el8.x86_64
- Object storage: STORJ
- Metadata engine info: SQLite

**Log**:
See [juicefs-storj-writeback-issue.log](https://github.com/juicedata/juicefs/files/10328057/juicefs-storj-writeback-issue.log).  Reflects that I have placed additional DEBUG statements just before and after s3.go:Put()'s call to s3.PutObject() to help clarify the misbehavior.  Redacted my IP with <MY_IP>, juicefs cache path with <JFS_CACHE>

SandyXSD:
close #3143 
For simplicity mutex/atomic is not used to control the status of block uploading or not. It's still possible that a block is uploaded for more than one time, but the possibility and limit should be much smaller.

### DIFF
--- a/pkg/chunk/cached_store.go
+++ b/pkg/chunk/cached_store.go
@@ -710,11 +710,10 @@ func NewCachedStore(storage object.ObjectStorage, config Config, reg prometheus.
 	}
 	store.initMetrics()
 	store.bcache = newCacheManager(&config, reg, func(key, fpath string, force bool) bool {
-		if force {
-			return store.addDelayedStaging(key, fpath, time.Now(), true)
-		} else if fi, err := os.Stat(fpath); err == nil {
-			return store.addDelayedStaging(key, fpath, fi.ModTime(), false)
+		if fi, err := os.Stat(fpath); err == nil {
+			return store.addDelayedStaging(key, fpath, fi.ModTime(), force)
 		} else {
+			logger.Warnf("Stat staging block %s: %s", fpath, err)
 			return false
 		}
 	})
--- a/pkg/chunk/cached_store.go
+++ b/pkg/chunk/cached_store.go
@@ -893,6 +893,7 @@ func (store *cachedStore) addDelayedStaging(key, stagingPath string, added time.
 	}
 	store.pendingMutex.Unlock()
 	if item.uploading {
+		logger.Debugf("Key %s is ignored since it's already being uploaded", key)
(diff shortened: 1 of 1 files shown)


pkg/chunk/cached_store.go
```diff
@@ -709,7 +711,7 @@ func NewCachedStore(storage object.ObjectStorage, config Config, reg prometheus.
 	store.initMetrics()
 	store.bcache = newCacheManager(&config, reg, func(key, fpath string, force bool) bool {
 		if force {
-			return store.addDelayedStaging(key, fpath, time.Time{}, true)
+			return store.addDelayedStaging(key, fpath, time.Now(), true)
```
davies, line 714: fi.ModTime()

SandyXSD: [32ec27a](https://github.com/juicedata/juicefs/pull/3157/commits/32ec27ab6ab51adeb07276fbb5ebaec5db75dc7a)

PR #3160, opened 2023-01-09
doc: update readme

zhijian-pro:


SandyXSD:
Better keep it until the feature is formally released (say, when v1.1 released).

PR #3161, opened 2023-01-09
sync with juicefs without mount point

davies:
This PR allow us sync with JuiceFS without mount point:

```
juicefs sync   a/path/   jfs://meta-url/path/
```

Since meta-url is used as `host`, so it should be escaped twice, for example, sqlite3://a.db should be

```
juicefs sync   a/path/   jfs://sqlite3%253A%252F%252Fa.db/path/
```

To simplify this, we could use environment variable to pass meta-url, for example
```
myfs=sqlite3://a.db juicefs sync  a/path/   jfs://myfs/path/
```

### DIFF
--- a/cmd/mdtest.go
+++ b/cmd/mdtest.go
@@ -35,7 +35,7 @@ import (
 	"github.com/urfave/cli/v2"
 )
 
-var ctx = meta.Background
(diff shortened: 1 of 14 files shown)


zhijian-pro (approved):

SandyXSD (approved):

lidaohang:
mark

PR #3162, opened 2023-01-09
fix prefix of symlink

davies:


zhijian-pro (approved):

zhijian-pro:
backport-1.0 ?

davies:
@zhijian-pro no, prefix is not used for disk and sftp, HDFS does not support symlink.

PR #3170, opened 2023-01-11
fix reading large response from .control

davies:
When a response has more than 2MB, the first 1MB will be dropped before reading.


pkg/vfs/vfs.go
```diff
@@ -510,25 +511,21 @@ func (v *VFS) Read(ctx Context, ino Ino, buf []byte, off uint64, fh uint64) (n i
 				err = syscall.EBADF
 				return
 			}
-			data := h.data
+			h.Lock()
```
SandyXSD, line 514: Is this necessary?

SandyXSD (approved):

PR #3178, opened 2023-01-12
vfs: change conf.Format from a pointer to a struct

### LINKED ISSUE #3175: update `.config` file after  use `juicefs config` cmd
The trashday is not be updated now

SandyXSD:
close #3175 

This PR also fix the issue that secret keys in baseMeta.fmt are removed for a short time [here](https://github.com/juicedata/juicefs/blob/v1.0.3/pkg/vfs/vfs.go#L942).

`meta.Load` can't be used for `.config` file to load the current format, because this method will update the one cached in baseMeta, which fails the detection of format change in `baseMeta.refresh`.

### DIFF
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -218,12 +218,8 @@ func (m *baseMeta) timeit(start time.Time) {
 	m.opDist.Observe(time.Since(start).Seconds())
 }
 
-func (m *baseMeta) GetBase() *baseMeta {
+func (m *baseMeta) getBase() *baseMeta {
 	return m
-}
-
-func (m *baseMeta) GetFormat() Format {
-	return *m.fmt
 }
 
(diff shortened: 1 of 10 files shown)


pkg/meta/base.go
```diff
@@ -218,10 +218,14 @@ func (m *baseMeta) timeit(start time.Time) {
 	m.opDist.Observe(time.Since(start).Seconds())
 }
 
-func (m *baseMeta) getBase() *baseMeta {
+func (m *baseMeta) GetBase() *baseMeta {
 	return m
 }
 
+func (m *baseMeta) GetFormat() Format {
```
davies, line 225: Why not add it to meta.Meta ?

SandyXSD: bfd5225

PR #3197, opened 2023-01-28
meta: recreate the session with info if it was cleaned before

### LINKED ISSUE #3183: Session information is not recreated after network failure recovery
**What happened**:
We injected a 5 minute network failure at a certain node. After the fault is restored, the exception below will always be reported
```
2023/01/13 15:06:57.651702 juicefs[3896811] <ERROR>: update session: no session found matching sid: 969 [sql.go:1904]
```
the sid=969 has been cleared in meta-store, but the mount client has not recreated the session.

mount client can recreated the session information after Network failure recovery 

injected a 5 minute network failure to meta-store

**Environment**:
- JuiceFS version (use `juicefs --version`) or Hadoop Java SDK version: juicefs version 1.0.3+2022-12-27.e4bf15a
- Kernel (e.g. `uname -a`):
- Others:

SandyXSD:
close #3183

### DIFF
--- a/pkg/meta/redis.go
+++ b/pkg/meta/redis.go
@@ -1935,21 +1935,14 @@ func (m *redisMeta) doFindStaleSessions(limit int) ([]uint64, error) {
 func (m *redisMeta) doRefreshSession() error {
 	ctx := Background
 	ssid := strconv.FormatUint(m.sid, 10)
-	if ok, err := m.rdb.HExists(ctx, m.sessionInfos(), ssid).Result(); err != nil {
-		return err
-	} else if !ok {
+	// we have to check sessionInfo here because the operations are not within a transaction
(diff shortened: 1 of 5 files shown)


pkg/meta/redis.go
```diff
@@ -1932,10 +1932,21 @@ func (m *redisMeta) doFindStaleSessions(limit int) ([]uint64, error) {
 	return sids, nil
 }
 
-func (m *redisMeta) doRefreshSession() {
-	m.rdb.ZAdd(Background, m.allSessions(), &redis.Z{
+func (m *redisMeta) doRefreshSession() error {
+	ctx := Background
+	ssid := strconv.FormatUint(m.sid, 10)
+	// we have to check sessionInfo here because the operations are not within a transaction
```
davies, line 1938: Why not use transaction here?

PR #3213, opened 2023-02-06
hadoop: user root in sdk as normal user

tangyoupeng:
If you used user `root` as superuser before. This pr will change the behaviour. If you still want to use `root` as superuser, you can add user `root` to supergroup use config `juicefs.supergroup` and `juicefs.groups`


tangyoupeng:
#3214

davies (approved):

PR #3224, opened 2023-02-09
umount: wait for writting staging chunks

### LINKED ISSUE #2644: wait completion of staging blocks in writeback mode
In writeback mode, we may want all the staging blocks to be uploaded before umount.

We can use the FUSE DESTROY operation as an signal to flush all staging blocks: https://github.com/torvalds/linux/commit/0ec7ca41f6f0f74a394a7d686bc0ee8afef84887

Hexilee:

close #2644

### Example

```bash
hexi@ArchLinux ~/s/juicefs (main)> sudo ./juicefs umount ~/mnt/jfs/
6.6 GiB staging chunks are being written back... 126 MiB/s, left 53s
```


cmd/umount.go
```diff
@@ -91,5 +96,89 @@ func umount(ctx *cli.Context) error {
 	setup(ctx, 1)
 	mp := ctx.Args().Get(0)
 	force := ctx.Bool("force")
+	if !force {
```
davies, line 99: This should be enabled using a flag (--wait-flush), or we changed the previous behavior.

cmd/umount.go
```diff
@@ -91,5 +96,89 @@ func umount(ctx *cli.Context) error {
...
+			return errors.Wrap(err, "failed to read staging directory")
+		}
+		if lastLeft == 0 {
+			lastLeft = size
+		}
+
+		if size == 0 {
+			return nil
+		}
+		speed := lastLeft - size
```
davies, line 156: speed could be negative

cmd/umount.go
```diff
@@ -91,5 +96,89 @@ func umount(ctx *cli.Context) error {
...
+			if os.IsNotExist(err) {
+				continue
+			}
+			return errors.Wrap(err, "failed to read staging directory")
+		}
+		if lastLeft == 0 {
+			lastLeft = size
+		}
+
+		if size == 0 {
```
davies, line 153: it's better to double check, wait until both lastLeft and size are 0

davies (changes requested):

cmd/umount.go
Hexilee: Done

cmd/umount.go
Hexilee: Done

cmd/umount.go
Hexilee: Done

cmd/umount.go
```diff
@@ -90,6 +100,94 @@ func doUmount(mp string, force bool) error {
...
+		if lastLeft > size {
+			speed = lastLeft - size
+		}
+
+		leftTime := 720 * time.Hour
+		if speed != 0 {
+			leftTime = time.Duration(size/speed) * time.Second
+		}
+		clearLastLine()
+		fmt.Printf("\r%s staging chunks are being written back... %s/s, left %s", humanize.IBytes(size), humanize.IBytes(speed), leftTime)
```
zhijian-pro, line 171: use progressbar ?

PR #3241, opened 2023-02-15
make open cache cleanup strategy configurable

### LINKED ISSUE #3220: capacity and cleanup lifecycle should be configurable

Capacity and cleanup lifecycle should be configurable. Currently, the capacity and cleanup lifecycle of `cleanup` are only affected by the size:

```go
func (o *openfiles) cleanup() {
	for {
		o.Lock()
		cutoff := time.Now().Add(-time.Hour).Add(time.Second * time.Duration(len(o.files)/1e4))
		var cnt, expired int
		for ino, of := range o.files {
			if of.refs <= 0 && of.lastCheck.Before(cutoff) {
				delete(o.files, ino)
				expired++
			}
			cnt++
			if cnt > 1e3 {
				break
			}
		}
		o.Unlock()
		time.Sleep(time.Millisecond * time.Duration(1000*(cnt+1-expired*2)/(cnt+1)))
	}
}
```

**Why is this needed**:

Sometimes we need much bigger capacity and lifetime as https://github.com/juicedata/juicefs/pull/3185

Hexilee:
close #3220


cmd/flags.go
```diff
@@ -196,6 +196,16 @@ func clientFlags() []cli.Flag {
 			Value: 0.0,
 			Usage: "open files cache timeout in seconds (0 means disable this feature)",
 		},
+		&cli.Float64Flag{
+			Name:  "open-cache-keep",
```
davies, line 200: open-cache-timeout

cmd/flags.go
```diff
@@ -196,6 +196,16 @@ func clientFlags() []cli.Flag {
 			Value: 0.0,
 			Usage: "open files cache timeout in seconds (0 means disable this feature)",
```
davies, line 197: The period to reuse open file without checking update.

cmd/flags.go
davies: maybe we don't need this

pkg/meta/openfile.go
```diff
@@ -20,36 +20,84 @@ type openFile struct {
...
-			if cnt > 1e3 {
+			if of.refs <= 0 {
+				if time.Since(of.lastCheck) > keepCache {
+					delete(o.files, ino)
+					deleted++
+					continue
+				}
+				expiredFiles[ino] = struct{}{}
+			}
+			if deleted > 1e3 {
```
SandyXSD, line 64: we have to use `cnt` here, otherwise the lock may hold for a long time when there are too many opened files.

pkg/meta/openfile.go
```diff
@@ -20,36 +20,84 @@ type openFile struct {
...
 				break
 			}
 		}
+		for len(expiredFiles) > 0 && uint64(len(o.files)) > limit && deleted <= 1e3 {
```
SandyXSD, line 68: refer [here](https://github.com/juicedata/juicefs/blob/main/pkg/chunk/disk_cache.go#L404) for a simpler implementation.

cmd/flags.go
Hexilee: Should we use a constant timeout?

Hexilee: Renamed

cmd/flags.go
Hexilee: Changed

pkg/meta/openfile.go
Hexilee: As the slice is ordered, we cannot iterate files after the 1e3 files at head.

Hexilee: It's my fault, the files are unordered.

cmd/flags.go
Hexilee: Removed

pkg/meta/openfile.go
Hexilee: Done

pkg/meta/openfile.go
```diff
@@ -20,36 +20,68 @@ type openFile struct {
...
-			if of.refs <= 0 && of.lastCheck.Before(cutoff) {
-				delete(o.files, ino)
-				expired++
-			}
 			cnt++
 			if cnt > 1e3 {
 				break
 			}
+			if of.refs <= 0 {
+				unusedFiles[ino] = of
```
davies, line 57: we can delete the file here rather than save it in a map

Hexilee: done

Hexilee: done

davies (approved):

PR #3242, opened 2023-02-16
object/ibmcos,ks3: check prefix and marker with returned key from S3

zhijian-pro:
ref #2040


davies:
ibmcos and ks3 already guarantee the order, we so don't need to add this check.

btw, we will loose the requirement on order, so we can do this later.

PR #3254, opened 2023-02-20
sync env in juicefs cluster mode

### LINKED ISSUE #3247: sync env to juicefs sync worker process

zhijian-pro:
close #3247

### DIFF
--- a/pkg/sync/cluster.go
+++ b/pkg/sync/cluster.go
@@ -32,6 +32,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/oliverisaac/shellescape"
+
 	"github.com/juicedata/juicefs/pkg/object"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/sync/config.go
```diff
@@ -44,6 +46,77 @@ type Config struct {
...
+}
+
+func envList() []string {
+	return []string{
+		"ACCESS_KEY",
+		"SECRET_KEY",
+		"SESSION_TOKEN",
+
+		// sync JFS to JFS, but neither has a mount point
+		"myfs",
```
davies, line 59: we should pass all the variable that contains "JFS"

zhijian-pro: ok

pkg/sync/cluster.go
```diff
@@ -244,7 +244,22 @@ func launchWorker(address string, config *Config, wg *sync.WaitGroup) {
 				return
 			}
 			// launch itself
-			var args = []string{host, rpath}
+			var args = []string{host}
+			// set env
+			var printEnv []string
+			for k, v := range config.Env {
+				args = append(args, fmt.Sprintf("%s=%s", k, v))
```
davies, line 251: should we encode the `v`? for example, it may contain `&`

zhijian-pro: use JFS='2&222 juicefs sync xxxx'  rather then `JFS=2&222 juicefs sync xxxxx`

image

zhijian-pro: yes

PR #3257, opened 2023-02-20
count space and inodes usage for each directory

Hexilee:


pkg/meta/base.go
```diff
@@ -230,6 +246,90 @@ func (m *baseMeta) checkRoot(inode Ino) Ino {
...
+}
+
+func (m *baseMeta) GetDirUsage(ctx Context, inode Ino) (space, inodes uint64, err error) {
+	return m.en.doGetDirUsage(ctx, m.checkRoot(inode))
+}
+
+func (m *baseMeta) increDirUsage(ctx Context, ino Ino, space int64, inodes int64) {
+	for i := 0; i < 50; i++ {
+		select {
+		case m.dirUsageEvents <- &dirUsageEvent{ino, space, inodes}:
```
SandyXSD, line 275: Most likely it will succeed, so maybe it's better to first try the channel once, and spawn the goroutine only when the channel is full.

pkg/meta/redis.go
```diff
@@ -2215,6 +2243,67 @@ func (m *redisMeta) doGetParents(ctx Context, inode Ino) map[Ino]int {
...
+		return
+	}
+	err = m.rdb.HSet(ctx, inodesKey, field, inodes).Err()
+	return
+}
+
+func (m *redisMeta) doIncreDirUsage(ctx Context, ino Ino, space int64, inodes int64) error {
+	spaceKey := m.dirUsedSpaceKey()
+	inodesKey := m.dirUsedInodesKey()
+	field := strconv.FormatUint(uint64(ino), 16)
```
SandyXSD, line 2265: better use base 10, just like others. see `Ino.String()`

pkg/meta/base.go
```diff
@@ -230,6 +246,90 @@ func (m *baseMeta) checkRoot(inode Ino) Ino {
 	}
 }
 
+func (m *baseMeta) countDirUsage(ctx Context, ino Ino) (space, inodes uint64, err syscall.Errno) {
+	var entries []*Entry
+	if err = m.en.doReaddir(ctx, ino, 1, &entries, -1); err != 0 {
+		return
+	}
+	for _, e := range entries {
+		if string(e.Name) == "." || string(e.Name) == ".." {
```
SandyXSD, line 255: `doReaddir` will not contain these two entries.

pkg/meta/base.go
```diff
@@ -230,6 +246,90 @@ func (m *baseMeta) checkRoot(inode Ino) Ino {
...
+		}
+	}
+}
+
+func (m *baseMeta) dirUsageBatch() {
+	commitChan := make(chan map[Ino]*dirUsageEvent, 16)
+	go func() {
+		for records := range commitChan {
+			for _, r := range records {
+				err := m.en.doIncreDirUsage(WrapContext(context.TODO()), r.ino, r.space, r.inodes)
```
SandyXSD, line 303: We may need to put several inodes within one transaction.

SandyXSD:
Also remove the `jfs.db` from the changed files.

Hexilee:
> Also remove the `jfs.db` from the changed files.

removed

pkg/meta/base.go
Hexilee: Resolved

pkg/meta/base.go
Hexilee: I think batch in application is enough, big transactions more likely conflict with other sessions. In this case, maybe we can call `doIncreDirUsage` parallelly as they won't conflict with each other.

pkg/meta/redis.go
Hexilee: Changed.

pkg/meta/base.go
Hexilee: Removed

pkg/meta/base.go
davies: We should aggregate these updates in memory (a map), don't use channel here.

pkg/meta/tkv.go
```diff
@@ -1917,6 +1951,61 @@ func (m *kvMeta) doGetParents(ctx Context, inode Ino) map[Ino]int {
...
+	}
+	return m.txn(func(tx *kvTxn) error {
+		key := m.dirUsageKey(ino)
+		rawUsage := tx.get(key)
+		if rawUsage == nil {
+			_, _, err := m.doSyncDirUsage(ctx, tx, ino)
+			return err
+		}
+		var usedSpace, usedInodes int64
+		if rawUsage != nil {
```
SandyXSD, line 1977: it's already checked before

pkg/meta/sql.go
```diff
@@ -2231,6 +2259,73 @@ func (m *dbMeta) doGetParents(ctx Context, inode Ino) map[Ino]int {
...
+		)
+		ret, err := s.Exec(sql, space, inodes, ino)
+		if err != nil {
+			return err
+		}
+		if n, _ := ret.RowsAffected(); n > 0 {
+			return nil
+		}
+
+		totalSpace, totalInodes, countErr := m.countDirUsage(ctx, ino)
```
SandyXSD, line 2284: maybe put it outside of the txn in case the directory is huge.

Hexilee: Done

pkg/meta/tkv.go
Hexilee: Removed

pkg/meta/tkv.go
```diff
@@ -181,6 +181,7 @@ All keys:
   SHssssssss         session heartbeat // for legacy client
   SIssssssss         session info
   SSssssssssiiiiiiii sustained inode
+  Uiiiiiiii			 space and inodes usage in directory
```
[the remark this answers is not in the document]
Hexilee: Aligned

pkg/meta/base.go

pkg/meta/base.go
```diff
@@ -230,6 +243,86 @@ func (m *baseMeta) checkRoot(inode Ino) Ino {
...
+	if inodes == 0 && space == 0 {
+		return
+	}
+	event := m.dirUsageEvents[ino]
+	event.space += space
+	event.inodes += inodes
+	m.dirUsageEvents[ino] = event
+}
+
+func (m *baseMeta) tryIncreDirUsage(ctx Context, ino Ino, space int64, inodes int64) bool {
```
davies, line 277: we don't need this, increDirUsage does not block

pkg/meta/base.go
```diff
@@ -769,7 +863,12 @@ func (m *baseMeta) Mknod(ctx Context, parent Ino, name string, _type uint8, mode
 	if m.checkQuota(4<<10, 1) {
 		return syscall.ENOSPC
 	}
-	return m.en.doMknod(ctx, m.checkRoot(parent), name, _type, mode, cumask, rdev, path, inode, attr)
+	err := m.en.doMknod(ctx, m.checkRoot(parent), name, _type, mode, cumask, rdev, path, inode, attr)
+	newSpace := align4K(0)
+	if err == 0 && !m.tryIncreDirUsage(ctx, parent, newSpace, 1) {
+		go m.mustIncreDirUsage(ctx, parent, newSpace, 1)
```
davies, line 869: don't create goroutine

pkg/meta/base.go
Hexilee: Removed

pkg/meta/base.go
Hexilee: Removed

SandyXSD:
@Hexilee We should delete the usage key when removing a directory.

Hexilee:
> @Hexilee We should delete the usage key when removing a directory.

Done

davies (approved):

PR #3258, opened 2023-02-21
dont set the default schema when remove password form titles

zhijian-pro:
fix #3256

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -295,12 +295,7 @@ func setup(c *cli.Context, n int) {
 }
 
 func removePassword(uri string) {
-	var uri2 string
-	if strings.Contains(uri, "://") {
-		uri2 = utils.RemovePassword(uri)
-	} else {
-		uri2 = utils.RemovePassword("redis://" + uri)
-	}
+	uri2 := utils.RemovePassword(uri)
 	if uri2 != uri {
 		for i, a := range os.Args {
 			if a == uri {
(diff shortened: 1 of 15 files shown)


pkg/utils/utils.go
```diff
@@ -101,6 +103,32 @@ func RemovePassword(uri string) string {
...
+				os.Args[i] = uri2
+				break
+			}
+		}
+	}
+	gspt.SetProcTitle(strings.Join(os.Args, " "))
+}
+
+func RemoveObjPwdFromTitle(uri string) {
+	removePwdFromTitle(uri, "file")
```
SandyXSD, line 125: may be 'file' or 'sftp'?

pkg/utils/utils.go
```diff
@@ -101,6 +103,32 @@ func RemovePassword(uri string) string {
 	return uri[:sp+cp] + ":****" + uri[p:]
 }
 
+func removePwdFromTitle(uri, defaultSchema string) {
```
SandyXSD, line 106: IMO it's better to delete 'FromTitle'

pkg/utils/utils.go
zhijian-pro: default  object storage is `file`

cmd/main.go
```diff
@@ -293,21 +293,3 @@ func setup(c *cli.Context, n int) {
 		}
 	}
 }
-
-func removePassword(uri string) {
-	var uri2 string
-	if strings.Contains(uri, "://") {
-		uri2 = utils.RemovePassword(uri)
-	} else {
-		uri2 = utils.RemovePassword("redis://" + uri)
```
davies, line 302: Why we add this? Maybe just remove it.

PR #3262, opened 2023-02-23
meta/tikv: add background GC worker for TiKV

SandyXSD:
The GC cleans up old entries before 3 hours by default, and can be configured with the query string "gc-interval".
It should be disabled by setting "gc-interval=0" if there's already TiDB daemon or any other service that can trigger the GC deployed.

Refs: https://github.com/tikv/client-go/blob/master/examples/gcworker/gcworker.go

TODO: not fully tested.

### DIFF
--- a/pkg/meta/tkv.go
+++ b/pkg/meta/tkv.go
@@ -58,7 +58,7 @@ type tkvClient interface {
 	reset(prefix []byte) error
 	close() error
 	shouldRetry(err error) bool
-	bgJob(arg any)
+	gc(edge time.Time)
 }
 
 type kvTxn struct {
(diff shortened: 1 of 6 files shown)


pkg/meta/tkv.go
```diff
@@ -1935,6 +1937,19 @@ func (m *kvMeta) doFindDeletedFiles(ts int64, limit int) (map[Ino]uint64, error)
 }
 
 func (m *kvMeta) doCleanupSlices() {
+	if m.Name() == "tikv" {
+		interval := time.Hour * 3
+		if dur, err := time.ParseDuration(os.Getenv("TIKV_GC_INTERVAL")); err == nil {
```
davies, line 1942: it's better to move this into tkv_tikv.go, and use url query instead.

PR #3270, opened 2023-02-28
object/obs&tos : fix range get

zhijian-pro:
1. fix obs & tos range get
2. check get response status code  ref https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range
3. improve objbench cmd

### DIFF
--- a/cmd/objbench.go
+++ b/cmd/objbench.go
@@ -681,11 +681,11 @@ func functionalTesting(blob object.ObjectStorage, result *[][]string, colorful b
 			return fmt.Errorf("put object failed: %s", err)
 		}
 		defer blob.Delete(key) //nolint:errcheck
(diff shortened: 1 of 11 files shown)


cmd/objbench.go
```diff
@@ -718,7 +718,7 @@ func functionalTesting(blob object.ObjectStorage, result *[][]string, colorful b
 		}
 		// get middle
 		if d, e := get(blob, key, 2, 2); e != nil || d != "ll" {
-			return fmt.Errorf(`failed to get the intermediate byte of the object, expect "ll", but got %q, error: %s`, d, e)
+			return fmt.Errorf(`failed to get the one byte of the object, expect "ll", but got %q, error: %s`, d, e)
```
zhijian-pro: not one

PR #3202, opened 2023-01-30
meta/redis: ignore check maxMemoryPolicy when meta engine is keydb

zhijian-pro:
When the meta engine is keydb and storage provider is flash, do not adjust the maxMemoryPolicy.
```
# keydb info
redis-cli -h xxxxxxx  -p 6379
47.99.52.222:6379> info
# Server
redis_version:6.3.2
redis_git_sha1:d1dff8c2
redis_git_dirty:0
redis_build_id:ff5573d014b83759
redis_mode:standalone
os:Linux 4.18.0-348.7.1.el8_5.x86_64 x86_64
arch_bits:64
multiplexing_api:epoll
atomicvar_api:atomic-builtin
gcc_version:8.5.0
process_id:10515
process_supervised:no
run_id:cb07e0f23b039299373ff0264436b1847f473eaa
tcp_port:6379
server_time_usec:1675063191723253
uptime_in_seconds:203
uptime_in_days:0
hz:10
configured_hz:10
lru_clock:14118807
executable:/root/keydb-server
config_file:/root/keydb.conf

# Clients
connected_clients:1
cluster_connections:0
maxclients:10000
client_recent_max_input_buffer:16
client_recent_max_output_buffer:0
blocked_clients:0
tracking_clients:0
clients_in_timeout_table:0
current_client_thread:0
thread_0_clients:1
thread_1_clients:0

# Memory
used_memory:3512168
used_memory_human:3.35M
used_memory_rss:24985600
used_memory_rss_human:23.83M
used_memory_peak:3572488
used_memory_peak_human:3.41M
used_memory_peak_perc:98.31%
used_memory_overhead:3643080
used_memory_startup:3447240
used_memory_dataset:18446744073709420704
used_memory_dataset_perc:28411077666537472.00%
allocator_allocated:4683272
allocator_active:5758976
allocator_resident:10866688
total_system_memory:7995740160
total_system_memory_human:7.45G
used_memory_lua:37888
used_memory_lua_human:37.00K
used_memory_scripts:0
used_memory_scripts_human:0B
number_of_cached_scripts:0
maxmemory:3634427345
maxmemory_human:3.38G
maxmemory_policy:allkeys-lru
allocator_frag_ratio:1.23
allocator_frag_bytes:1075704
allocator_rss_ratio:1.89
allocator_rss_bytes:5107712
rss_overhead_ratio:2.30
rss_overhead_bytes:14118912
mem_fragmentation_ratio:7.20
mem_fragmentation_bytes:21516192
[...]

PR #3275, opened 2023-03-01
run keydb-server for TestKeyDB

Hexilee:
In #3202, @zhijian-pro introduced `TestKeyDB` in the unittest, and run `keydb-server` in unittest ci for it. However, the mutate-test may run `TestKeyDB` and also need `keydb-server`.


davies:
We have discussed this and decide to not run these tests in mutest test.

PR #3277, opened 2023-03-01
flush dir stat in less transactions

Hexilee:


pkg/meta/redis.go
```diff
@@ -2260,26 +2260,60 @@ func (m *redisMeta) doSyncDirStat(ctx Context, ino Ino) (space, inodes uint64, e
...
+		}
+	}
+
+	if len(nonexist) > 0 {
+		if err := m.batchCalcDirStat(ctx, nonexist); err != nil {
+			return err
+		}
+	}
+
+	for _, group := range m.groupBatch(batch, 1000) {
```
davies, line 2293: ignore those calculated above

pkg/meta/base.go
```diff
@@ -242,6 +242,38 @@ func (m *baseMeta) checkRoot(inode Ino) Ino {
 	}
 }
 
+func (m *baseMeta) batchCalcDirStat(ctx Context, stats map[Ino]*dirStat) error {
+	var eg errgroup.Group
+	for i, s := range stats {
+		ino, stat := i, s
+		eg.Go(func() error {
```
davies, line 249: They are independent, so we should not cancel others if one of them fail

pkg/meta/redis.go
```diff
@@ -2260,26 +2260,60 @@ func (m *redisMeta) doSyncDirStat(ctx Context, ino Ino) (space, inodes uint64, e
...
+		_, err := m.rdb.Pipelined(ctx, func(pipe redis.Pipeliner) error {
+			for _, ino := range group {
+				field := strconv.FormatUint(uint64(ino), 10)
+				if stat, ok := nonexist[ino]; ok {
+					pipe.HSetNX(ctx, spaceKey, field, stat.space)
```
davies, line 2298: just ignore it

pkg/meta/base.go
```diff
@@ -242,6 +242,38 @@ func (m *baseMeta) checkRoot(inode Ino) Ino {
...
+	return eg.Wait()
+}
+
+func (m *baseMeta) groupBatch(batch map[Ino]dirStat, size int) [][]Ino {
+	var inos []Ino
+	for ino := range batch {
+		inos = append(inos, ino)
+	}
+	var batches [][]Ino
+	for i := 0; i < len(inos); i += size {
```
davies, line 267: sort `inos` to reduce conflicts

pkg/meta/tkv.go
```diff
@@ -1956,30 +1956,53 @@ func (m *kvMeta) doSyncDirStat(ctx Context, ino Ino) (space, inodes uint64, err
...
 }
 
-var errEmptyStat = errors.New("empty stat")
+func (m *kvMeta) doUpdateDirStat(ctx Context, batch map[Ino]dirStat) error {
+	syncMap := make(map[Ino]*dirStat, 0)
+	for _, group := range m.groupBatch(batch, 20) {
+		err := m.txn(func(tx *kvTxn) error {
+			for _, ino := range group {
+				key := m.dirStatKey(ino)
+				rawStat := tx.get(key)
```
davies, line 1965: gets

pkg/meta/base.go
davies: Also, we cant to update the stats right after the calculation, don't wait for other large directories. So it's better to move the parallization out of calculation

pkg/meta/base.go
Hexilee: Done

pkg/meta/tkv.go
Hexilee: Done

pkg/meta/redis.go
Hexilee: Ignored

pkg/meta/base.go

PR #3281, opened 2023-03-02
CI: support skip mutate test for test cases

zhoucheng361:


.github/scripts/mutate/how_to_use_mutate_test.md
```diff
@@ -39,6 +39,13 @@ For example:
 		return "", fmt.Errorf("failed to execute command `lsb_release`: %s", err)
 	}
 
+# how to skip a specific test case?
+if you don't want to run a specific test case, you can add "//skip mutate" after the test case function.
+For example:
+func TestRandomWrite(t *testing.T) {//skip mutate
```
davies, line 45: Can you add this comment for those tests that require external services (not available in mutate test)

PR #3293, opened 2023-03-06
always flush dir stat in background

Hexilee:


pkg/meta/base.go
```diff
@@ -417,6 +417,7 @@ func (m *baseMeta) NewSession() error {
 		logger.Infof("Create read-only session OK with version: %s", version.Version())
 		return nil
 	}
+	go m.flushDirStat()
```
davies, line 420: move it faster  created session

Hexilee: moved

PR #3282, opened 2023-03-02
fsck: support check or repair dir stat

Hexilee:

PR #3296, opened 2023-03-06
info recursively support fast mode

Hexilee:
Optimize `info -r` based on #3257

### DIFF
--- a/cmd/info.go
+++ b/cmd/info.go
@@ -57,12 +57,11 @@ $ juicefs info -i 100`,
 			&cli.BoolFlag{
 				Name:    "recursive",
 				Aliases: []string{"r"},
-				Usage:   "get summary of directories recursively (NOTE: it may take a long time for huge trees)",
+				Usage:   "get summary of directories recursively (NOTE: it may be inaccurate, use --strict to get accurate result)",
 			},
 			&cli.BoolFlag{
-				Name:    "fast",
-				Aliases: []string{"f"},
-				Usage:   "get summary of directories fastly (NOTE: it may be inaccurate)",
+				Name:  "strict",
+				Usage: "get accurate summary of directories (NOTE: it may take a long time for huge trees)",
 			},
 			&cli.BoolFlag{
 				Name:  "raw",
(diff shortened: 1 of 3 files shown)


cmd/info.go
```diff
@@ -108,11 +116,12 @@ func info(ctx *cli.Context) error {
...
 		}
 
-		wb := utils.NewBuffer(8 + 10)
+		wb := utils.NewBuffer(8 + 11)
 		wb.Put32(meta.InfoV2)
-		wb.Put32(10)
+		wb.Put32(11)
 		wb.Put64(inode)
 		wb.Put8(recursive)
+		wb.Put8(fast)
```
davies, line 124: client in old version will use it as raw

cmd/info.go
```diff
@@ -59,6 +59,11 @@ $ juicefs info -i 100`,
 				Aliases: []string{"r"},
 				Usage:   "get summary of directories recursively (NOTE: it may take a long time for huge trees)",
 			},
+			&cli.BoolFlag{
+				Name:    "fast",
```
davies, line 63: Should we use the faster one as default? then we could add a option called `--accurate` or `--strict`

pkg/meta/utils.go
```diff
@@ -369,3 +369,50 @@ func GetSummary(r Meta, ctx Context, inode Ino, summary *Summary, recursive bool
...
+	} else {
+		summary.Files++
+		summary.Length += attr.Length
+		summary.Size += uint64(align4K(attr.Length))
+	}
+	return 0
+}
+
+func fastGetSummary(r Meta, ctx Context, inode Ino, summary *Summary, recursive bool) syscall.Errno {
+	space, _, err := r.GetDirStat(ctx, inode)
```
davies, line 389: This should work in read-only mode

cmd/info.go
Hexilee: Reasonable

pkg/meta/utils.go
Hexilee: In #3282, the `GetDirStat` add a `trySync` parameter, it works in read-only mode if we disable `trySync`.

cmd/info.go
Hexilee: Done

cmd/info.go
Hexilee: Fixed

Issue #3272, opened 2023-02-28
The unit test result did not reach the expected object store


zhijian-pro:
## I tested all of the supported object stores( include file tikv redis pg and mysql etc), and the following is a list of exception results to watch out for.

- [x] UFile: This is a us3 overwrite upload bug
- [x] Azure:  not supported by the server
- [x] B2  #3274
- [x] BOS #3270 
- [x] HDFS: HDFS server does not support `:` and other characters.
- [x] OOS #3273 
- [x] MySQL #3299 #3303
- [x] Disk #3301

### TestUFile
```
=== RUN   TestUFile
    object_storage_test.go:87: PUT testEncodeFile failed: parse "https://zhijian-test4.cn-wlcb.ufileos.com/unit-test/测试编码文件{\"name\":\"zhijian\"}\x1f": net/url: invalid control character in URL
    object_storage_test.go:232: list api error: not supported delimiter
3 
    object_storage_test.go:330: CompleteMultipart failed: CompleteMultipart: status: 400, message: {"RetCode":-148669, "ErrMsg":"validate initidx and partlist [zhijian-test4:c5a134b6-9899-4692-8e6f-4fa8ad6a361d] failed, err: invalid middle block size"}
--- FAIL: TestUFile (14.17s)
```

### TestAzure
```
=== RUN   TestAzure
    object_storage_test.go:87: PUT testEncodeFile failed: ===== INTERNAL ERROR =====
        ===== INTERNAL ERROR =====
        PUT https://zhijiantest1.blob.core.windows.net/testbucket1/unit-test/测试编码文件{"name":"zhijian"}
        --------------------------------------------------------------------------------
        RESPONSE 400: 400 Bad Request
        ERROR CODE UNAVAILABLE
        --------------------------------------------------------------------------------
        <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN""http://www.w3.org/TR/html4/strict.dtd">
        <HTML><HEAD><TITLE>Bad Request</TITLE>
        <META HTTP-EQUIV="Content-Type" Content="text/html; charset=us-ascii"></HEAD>
        <BODY><h2>Bad Request - Invalid URL</h2>
        <hr><p>HTTP Error 400. The request URL is invalid.</p>
        </BODY></HTML>
        
        --------------------------------------------------------------------------------
    object_storage_test.go:232: list api error: not supported delimiter
[...]

PR #3298, opened 2023-03-07
fix MultipartUpload test

zhijian-pro:
For UFile: The size of all parts except the last part cannot be less than 4MB, otherwise calling the FinishMultipartUpload API will fail.

ref #3272

### DIFF
--- a/pkg/object/object_storage_test.go
+++ b/pkg/object/object_storage_test.go
@@ -318,11 +318,19 @@ func testStorage(t *testing.T, s ObjectStorage) {
 			println(len(pending), marker)
 		}
 		part2Size := 1 << 20
+		// ufile: The size of all parts except the last part cannot be less than 4MB,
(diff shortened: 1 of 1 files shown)


davies:
Can we make sure that `sync` use correct part size for UFile?

pkg/object/object_storage_test.go
```diff
@@ -318,11 +318,19 @@ func testStorage(t *testing.T, s ObjectStorage) {
 			println(len(pending), marker)
 		}
 		part2Size := 1 << 20
+		// ufile: The size of all parts except the last part cannot be less than 4MB,
+		// otherwise it will cause the call to the FinishMultipartUpload API to fail.
+		if _, ok := s.(*withPrefix).os.(*ufile); ok {
+			part2Size = partSize
```
davies, line 324: use MinPartSize

PR #3302, opened 2023-03-08
count total length of files in dir stat

Hexilee:


pkg/meta/redis.go
```diff
@@ -528,6 +528,10 @@ func (m *redisMeta) usedSpaceKey() string {
 	return m.prefix + usedSpace
 }
 
+func (m *redisMeta) dirFileLengthKey() string {
+	return m.prefix + "dirFileLength"
```
SandyXSD, line 532: "dirDataLength"

pkg/meta/base.go
```diff
@@ -277,50 +278,49 @@ func (m *baseMeta) groupBatch(batch map[Ino]dirStat, size int) [][]Ino {
...
 		err = errors.Wrap(eno, "calc dir stat")
 		return
 	}
 	for _, e := range entries {
 		inodes += 1
-		var newSpace uint64
+		var l uint64
 		if e.Attr.Typ == TypeFile {
-			newSpace = e.Attr.Length
+			l = e.Attr.Length
```
SandyXSD, line 291: also add length for symlink

pkg/meta/redis.go
Hexilee: Renamed

pkg/meta/base.go
Hexilee: After this PR, both of the fast summary and strict summary won't count the length of symlink.

pkg/meta/base.go
```diff
@@ -280,57 +281,56 @@ func (m *baseMeta) groupBatch(batch map[Ino]dirStat, size int) [][]Ino {
 	return batches
 }
 
-func (m *baseMeta) calcDirStat(ctx Context, ino Ino) (space, inodes uint64, err error) {
+func (m *baseMeta) calcDirStat(ctx Context, ino Ino) (length, space, inodes uint64, err error) {
```
davies, line 284: we should use `dirStat` instead of `length, space, inodes uint64`

davies: Maybe we can do the refactor first, then it will be easy to add `length`

PR #3308, opened 2023-03-09
cmd/objbench: check list error

### LINKED ISSUE #3306: objbench test "list objects" swallows errors with S3 compatible backends
**What happened**:

When running `objbench` against a non-conforming S3 backend, the "list objects" test displays `list failed: not supported` instead of the actual error.

I expect the actual error (e.g., `found invalid key %s from List, prefix: %s, marker: %s`) to be displayed.

```
juicefs objbench --storage s3 --access-key anonymous http://localhost:8000/bucket
```

The problem appears to be at line 570: https://github.com/juicedata/juicefs/blob/585dcb32a1688d7d90264c679571d199f44cbdda/cmd/objbench.go#L565-L581

If `s.List` returns an `err` then it's ignored and replaced with the `err` value from `s.ListAll`, which for an S3 compatible backend is always `notSupported`:
https://github.com/juicedata/juicefs/blob/585dcb32a1688d7d90264c679571d199f44cbdda/pkg/object/s3.go#L220-L222

**Environment**:
- JuiceFS version (use `juicefs --version`) or Hadoop Java SDK version: juicefs version 1.0.3+2022-12-27.e4bf15aca99
- Cloud provider or hardware configuration running JuiceFS: MacBook Pro
- OS (e.g `cat /etc/os-release`): macOS
- Kernel (e.g. `uname -a`): Darwin ... 22.3.0 Darwin Kernel Version 22.3.0: Mon Jan 30 20:42:11 PST 2023; root:xnu-8792.81.3~2/RELEASE_X86_64 x86_64
- Object storage (cloud provider and region, or self maintained): self maintained
- Metadata engine info (version, cloud provider managed or self maintained): SQLite3
- Network connectivity (JuiceFS to metadata engine, JuiceFS to object storage): localhost
- Others:

zhijian-pro:
close #3306

CI fixed by  #3310

### DIFF
--- a/cmd/objbench.go
+++ b/cmd/objbench.go
@@ -564,7 +564,7 @@ func (bm *benchMarkObj) chtimes(key string, startKey int) error {
 
 func listAll(s object.ObjectStorage, prefix, marker string, limit int64) ([]object.Object, error) {
 	r, err := s.List(prefix, marker, "", limit)
-	if err == nil {
+	if !errors.Is(err, utils.ENOTSUP) {
 		return r, nil
 	}
 	ch, err := s.ListAll(prefix, marker)
--- a/cmd/objbench.go
+++ b/cmd/objbench.go
@@ -565,7 +565,7 @@ func (bm *benchMarkObj) chtimes(key string, startKey int) error {
 func listAll(s object.ObjectStorage, prefix, marker string, limit int64) ([]object.Object, error) {
 	r, err := s.List(prefix, marker, "", limit)
 	if !errors.Is(err, utils.ENOTSUP) {
-		return r, nil
+		return r, err
 	}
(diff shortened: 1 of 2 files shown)


cmd/gateway_noop.go
```diff
@@ -31,7 +30,7 @@ func cmdGateway() *cli.Command {
 		Usage:       "Start an S3-compatible gateway (not included)",
 		Description: `This feature is not included. If you want it, recompile juicefs without "nogateway" flag`,
 		Action: func(*cli.Context) error {
-			return errors.New("not supported")
```
davies, line 34: Why? We should minimize the patch for bugfix

davies (approved):

PR #3317, opened 2023-03-10
optimize fast summary

Hexilee:
- fix a bug in `doSyncDirStat` of redis engine
- use `inodes` in dir stat of leaf dirs instead read them

### DIFF
--- a/pkg/meta/redis.go
+++ b/pkg/meta/redis.go
@@ -2241,7 +2241,7 @@ func (m *redisMeta) doGetParents(ctx Context, inode Ino) map[Ino]int {
 }
 
 func (m *redisMeta) doSyncDirStat(ctx Context, ino Ino) (*dirStat, error) {
-	field := strconv.FormatUint(uint64(ino), 16)
+	field := strconv.FormatUint(uint64(ino), 10)
 	space, inodes, err := m.calcDirStat(ctx, ino)
 	if err != nil {
 		return nil, err
--- a/pkg/meta/redis.go
+++ b/pkg/meta/redis.go
@@ -2241,7 +2241,7 @@ func (m *redisMeta) doGetParents(ctx Context, inode Ino) map[Ino]int {
 }
 
 func (m *redisMeta) doSyncDirStat(ctx Context, ino Ino) (*dirStat, error) {
-	field := strconv.FormatUint(uint64(ino), 10)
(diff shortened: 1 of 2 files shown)


pkg/meta/redis.go
```diff
@@ -2241,7 +2241,7 @@ func (m *redisMeta) doGetParents(ctx Context, inode Ino) map[Ino]int {
 }
 
 func (m *redisMeta) doSyncDirStat(ctx Context, ino Ino) (*dirStat, error) {
-	field := strconv.FormatUint(uint64(ino), 16)
+	field := strconv.FormatUint(uint64(ino), 10)
```
SandyXSD, line 2244: use ino.String() as the field

Hexilee: Replaced

pkg/meta/utils.go
```diff
@@ -385,14 +386,23 @@ func FastGetSummary(r Meta, ctx Context, inode Ino, summary *Summary, recursive
...
+	space, inodes, err := r.GetDirStat(ctx, inode)
 	if err != nil {
 		return errno(err)
 	}
 	summary.Size += space
 	summary.Length += space
-	summary.Dirs++
+
+	if !attr.Full {
+		if st := r.GetAttr(ctx, inode, attr); st != 0 {
```
davies, line 398: Some directory could be removed, we could ignore ENOENT

pkg/meta/utils.go
```diff
@@ -385,14 +386,23 @@ func FastGetSummary(r Meta, ctx Context, inode Ino, summary *Summary, recursive
 	return 0
 }
 
-func fastGetSummary(r Meta, ctx Context, inode Ino, summary *Summary, recursive bool) syscall.Errno {
-	space, _, err := r.GetDirStat(ctx, inode)
+func fastGetSummary(r Meta, ctx Context, inode Ino, attr *Attr, summary *Summary, recursive bool) syscall.Errno {
```
davies, line 389: only the attribute of the top directory is useful, we can ignore it.

pkg/meta/utils.go
Hexilee: Ignored

pkg/meta/utils.go
Hexilee: Ignored

PR #3333, opened 2023-03-15
support granting fuse access in unprivileged containers

Hexilee:


cmd/mount_unix.go
```diff
@@ -147,6 +147,10 @@ func mount_flags() []cli.Flag {
 			Name:  "update-fstab",
 			Usage: "add / update entry in /etc/fstab, will create a symlink at /sbin/mount.juicefs if not existing",
 		},
+		&cli.BoolFlag{
+			Name:  "grant-access",
+			Usage: "grant access to the /dev/fuse device (used in unprivileged containers)",
```
davies, line 152: We should try to do that when needed, don't add this option

Hexilee: removed

cmd/mount_unix.go
```diff
@@ -184,13 +184,17 @@ func mount_main(v *vfs.VFS, c *cli.Context) {
 	if os.Getuid() == 0 && os.Getpid() != 1 {
 		disableUpdatedb()
 	}
-
+	fuse.EnsureFuseDev()
```
davies, line 187: we should move these into pkg/fuse

pkg/fuse/device_linux.go
```diff
@@ -0,0 +1,91 @@
+/*
+ * JuiceFS, Copyright 2023 Juicedata, Inc.
```
davies, line 2: Do we copy this file from Chaos Mesh? please keep the original license header

Hexilee: Actually, this file existed in the early version of Chaos Mesh, it has been removed for several years. Should we still keep the original license header?

pkg/fuse/device_linux.go
```diff
@@ -0,0 +1,91 @@
...
+		return errors.Errorf("fail to find device cgroup")
+	}
+
+	deviceCgroupPath = "/sys/fs/cgroup/devices" + deviceCgroupPath + "/devices.allow"
+	f, err := os.OpenFile(deviceCgroupPath, os.O_WRONLY, 0)
+	if err != nil {
+		return errors.Wrapf(err, "open %s", deviceCgroupPath)
+	}
+	// 10, 229 according to https://www.kernel.org/doc/Documentation/admin-guide/devices.txt
+	content := "c 10:229 rwm"
```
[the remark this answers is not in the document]
Hexilee: ## non-privileged

Yes, this grant is necessary. If we `cat` the `devices.list` in cgroup of a non-privileged docker container, we can see following output:

```bash
root@5119aebeaab0:/# cat /sys/fs/cgroup/devices/docker/5119aebeaab082fbc73e86ef62c78cf75dd74114ea470c3bf9bafdfb43af5a97/devices.list
b *:* m
c *:* m
c 1:3 rwm
c 1:5 rwm
c 1:7 rwm
c 1:8 rwm
c 1:9 rwm
c 5:0 rwm
c 5:1 rwm
c 5:2 rwm
c 10:200 rwm
c 136:* rwm
root@5119aebeaab0:/#
```
According to the `devices.list`, we cannot read or write the fuse device (c 10:229). However, if we grant access and `cat` the `devices.list` again, we can see we are allowed to read and write the fuse device:

```bash
root@5119aebeaab0:/# cat /sys/fs/cgroup/devices/docker/5119aebeaab082fbc73e86ef62c78cf75dd74114ea470c3bf9bafdfb43af5a97/devices.list
b *:* m
c *:* m
c 1:3 rwm
c 1:5 rwm
c 1:7 rwm
c 1:8 rwm
c 1:9 rwm
c 5:0 rwm
c 5:1 rwm
c 5:2 rwm
c 10:200 rwm
c 136:* rwm
c 10:229 rwm
```

## privileged

If we `cat` the `devices.list` in cgroup of a privileged docker container, we can see it allow us to read and write all devices:

```bash
root@cc98422c03e6:/# cat /sys/fs/cgroup/devices/docker/cc98422c03e61f5340a77c49e18f28f189bbafd1da2979258f5d64bf64672b5a/devices.list
a *:* rwm
```

The `devices.list` won't change after we write "c 10:229 rwm" into `devices.allow`.

pkg/fuse/device_linux.go
davies: yes, if it's released once

pkg/fuse/device_linux.go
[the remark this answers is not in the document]
davies: So we can skip this if we saw "*:*" in it?

pkg/fuse/device_linux.go
```diff
@@ -0,0 +1,91 @@
...
+		if parts[1] == "devices" {
+			deviceCgroupPath = parts[2]
+		}
+	}
+
+	if len(deviceCgroupPath) == 0 {
+		return errors.Errorf("fail to find device cgroup")
+	}
+
+	deviceCgroupPath = "/sys/fs/cgroup/devices" + deviceCgroupPath + "/devices.allow"
```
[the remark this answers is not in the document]
Hexilee: Moreover, we need capabilities `SYS_NICE` to set priority.

cmd/mount_unix.go
Hexilee: Moved

pkg/fuse/device_linux.go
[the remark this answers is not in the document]
Hexilee: We can skip it if we saw "a \*:\* rwm" in it.

pkg/fuse/device_linux.go
Hexilee: Ok, kept.

pkg/fuse/device_linux.go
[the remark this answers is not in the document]
Hexilee: Completed

PR #3340, opened 2023-03-16
support parallel directory summary

Hexilee:
Both of strict summary and fast summary should run `readdir` parallelly.

## Performance improvement

In engines that support high concurrency, such as TiKV, the parallel summary feature has become 20 times faster than before.

image

### DIFF
--- a/pkg/meta/utils.go
+++ b/pkg/meta/utils.go
@@ -332,43 +332,49 @@ func (m *baseMeta) Remove(ctx Context, parent Ino, name string, count *uint64) s
 }
 
 func GetSummary(r Meta, ctx Context, inode Ino, summary *Summary, recursive bool) syscall.Errno {
-	var attr Attr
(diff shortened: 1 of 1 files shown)


pkg/meta/utils.go
```diff
@@ -379,57 +398,73 @@ func FastGetSummary(r Meta, ctx Context, inode Ino, summary *Summary, recursive
...
-	}
-	for _, e := range entries {
-		if e.Inode == inode || len(e.Name) == 2 && bytes.Equal(e.Name, []byte("..")) {
-			continue
-		}
-		if e.Attr.Typ == TypeDirectory {
-			summary.Dirs++
-			if recursive {
-				if st := fastGetSummary(r, ctx, e.Inode, summary, recursive); st != 0 {
+	const concurrency = 1000
```
davies, line 411: that's too high, may DOS the database, 20-50 should be enough.

Hexilee: Fixed

PR #3349, opened 2023-03-20
Docs: keep the intro video not play automatically

yuhr123:
There are two change in this PR:

1. Repalce the <BilibiliEmbedRenderer /> tag with the native iframe tag, but remain the settings.
2. Switch off the autoplay action.


yuhr123:
It seems that the Markdown lint doesn't allow using iframe in Markdown, I will try to fix it.

xiaogaozi (approved):

PR #3353, opened 2023-03-20
doSyncDirStat return ENOENT if inode is nonexistent

Hexilee:


pkg/meta/tkv.go
```diff
@@ -1977,26 +1977,26 @@ func (m *kvMeta) doGetParents(ctx Context, inode Ino) map[Ino]int {
...
-	stat, err := m.calcDirStat(ctx, ino)
-	if err != nil {
-		return nil, err
+	stat, st := m.calcDirStat(ctx, ino)
+	if st != 0 {
+		return nil, st
 	}
-	err = m.client.txn(func(tx *kvTxn) error {
+	err := m.client.txn(func(tx *kvTxn) error {
+		if !tx.exist(m.inodeKey(ino)) {
```
zhijian-pro, line 1989: better :
```
a := tx.get(m.inodeKey(ino))
    if a == nil {
    return syscall.ENOENT
}
```
`exist` isn't a direct interface to the metadata engine.
`exist` is a confusing name. eg: `m.exist(m.entryKey(parent, name))` is wrong usage in tkv but right  in redis.
```
func (tx *tikvTxn) exist(prefix []byte) bool {
	it, err := tx.Iter(prefix, nextKey(prefix))
	if err != nil {
		panic(err)
	}
	defer it.Close()
	return it.Valid()
}
```

Hexilee: Fixed

pkg/meta/sql.go
```diff
@@ -2304,23 +2304,27 @@ func (m *dbMeta) doUpdateDirStat(ctx Context, batch map[Ino]dirStat) error {
...
+	err := m.txn(func(s *xorm.Session) error {
+		exist, err := s.Exist(&node{Inode: ino})
+		if err != nil {
+			return err
+		}
+		if !exist {
+			return syscall.ENOENT
+		}
+		_, err = s.Insert(&dirStats{ino, stat.length, stat.space, stat.inodes})
 		if err != nil && strings.Contains(err.Error(), "UNIQUE constraint failed") {
```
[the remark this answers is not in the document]
Hexilee: It's incompatible, actually, we even cannot ensure it will work on future version of SQLite. The error returned by doSyncDirStat should be ignored (log only) by most of callers.

PR #3366, opened 2023-03-23
cmd/clone:  clean the detached nodes in the gc command 

zhijian-pro:


pkg/meta/base.go
```diff
@@ -1977,3 +1994,58 @@ func (m *baseMeta) ScanDeletedObject(ctx Context, tss trashSliceScan, pss pendin
 	}
 	return eg.Wait()
 }
+
+func (m *baseMeta) Clone(ctx Context, srcIno, dstParentIno Ino, dstName string, cmode uint8, cumask uint16, count, total *uint64) syscall.Errno {
```
davies, line 1998: Can we move these refactor to another PR?

zhijian-pro: refactor depends on this PR, which will come out after this merger.

pkg/meta/redis.go
```diff
@@ -4311,3 +4301,37 @@ func (m *redisMeta) mkNodeWithAttr(ctx Context, tx *redis.Tx, srcIno Ino, srcAtt
...
+	exists, err := m.rdb.Exists(ctx, m.inodeKey(detachedNode)).Result()
+	if err != nil {
+		return errno(err)
+	}
+	if exists == 1 {
+		rmConcurrent := make(chan int, 10)
+		if eno := m.emptyDir(ctx, detachedNode, true, nil, rmConcurrent); eno != 0 {
+			return eno
+		}
+		if err := m.txn(ctx, func(tx *redis.Tx) error {
```
davies, line 4315: the node could be a file

pkg/meta/sql.go
```diff
@@ -166,6 +166,11 @@ type dirStats struct {
 	UsedInodes int64 `xorm:"notnull"`
 }
 
+type detachedNode struct {
+	Inode  Ino   `xorm:"pk notnull"`
+	Expire int64 `xorm:"notnull"`
```
davies, line 171: Added

pkg/meta/tkv.go
```diff
@@ -184,6 +184,7 @@ All keys:
   SHssssssss         session heartbeat // for legacy client
   SIssssssss         session info
   SSssssssssiiiiiiii sustained inode
+  Niiiiiiii	         detached inodes
```
davies, line 187: align

pkg/meta/redis.go
zhijian-pro: The detached node  is logged into the metadata engine only when `attach = false`.
`attach` is always true if the clone object is a file.
```
mkNodeWithAttr
...
if !attach {
    pipe.ZAdd(ctx, m.detachedNodes(), redis.Z{Member: dstIno.String(), Score: float64(time.Now().Unix())})
}
...
```

pkg/meta/tkv.go
zhijian-pro: image
image

github display cannot be aligned with the local IDE at the same time

zhijian-pro: vim
image
I think the local ide presentation is right.

zhijian-pro: Maybe it was a special character problem before, so delete it and rewrite it

pkg/meta/sql.go
```diff
@@ -4007,9 +3996,56 @@ func (m *dbMeta) mkNodeWithAttr(ctx Context, s *xorm.Session, srcIno Ino, srcNod
...
+			}
+			if _, err := s.Delete(&xattr{Inode: detachedNode}); err != nil {
+				return err
+			}
+			return nil
+		}); err != nil {
+			return errno(err)
+		}
+	}
+	_, err = m.db.Delete(&node{Inode: detachedNode})
```
davies, line 4049: when it exists

pkg/meta/sql.go
```diff
@@ -4007,9 +3996,56 @@ func (m *dbMeta) mkNodeWithAttr(ctx Context, s *xorm.Session, srcIno Ino, srcNod
 		}
 	}
 
+	if !attach {
+		return mustInsert(s, &detachedNode{Inode: *dstIno, Added: time.Now().Unix()})
```
davies, line 4000: move to the end

pkg/meta/redis.go
```diff
@@ -4299,6 +4286,9 @@ func (m *redisMeta) mkNodeWithAttr(ctx Context, tx *redis.Tx, srcIno Ino, srcAtt
 
 	_, err = tx.TxPipelined(ctx, func(pipe redis.Pipeliner) error {
 		pipe.Set(ctx, m.inodeKey(*dstIno), m.marshal(srcAttr), 0)
+		if !attach {
+			pipe.ZAdd(ctx, m.detachedNodes(), redis.Z{Member: dstIno.String(), Score: float64(time.Now().Unix())})
```
davies, line 4290: move to the end
```
if attach {
} else {
}
```

pkg/meta/tkv.go
```diff
@@ -3470,6 +3458,7 @@ func (m *kvMeta) mkNodeWithAttr(ctx Context, tx *kvTxn, srcIno Ino, srcAttr *Att
 		}
 	}
 
+	tx.set(m.detachedKey(*dstIno), m.packInt64(time.Now().Unix()))
```
davies, line 3461: when !attach

pkg/meta/base.go
```diff
@@ -70,6 +70,8 @@ type engine interface {
 	doCleanupSlices()
 	doCleanupDelayedSlices(edge int64) (int, error)
 	doDeleteSlice(id uint64, size uint32) error
+	doFindDetachedNodes(t time.Time) []Ino
```
davies, line 73: add unit test for these

pkg/meta/sql.go
davies: you don't need to delete it, if not existed

pkg/meta/redis.go
```diff
@@ -4306,8 +4296,44 @@ func (m *redisMeta) mkNodeWithAttr(ctx Context, tx *redis.Tx, srcIno Ino, srcAtt
 			tx.HSet(ctx, m.entryKey(dstParentIno), dstName, m.packEntry(srcAttr.Typ, *dstIno))
 			tx.IncrBy(ctx, m.usedSpaceKey(), align4K(0))
 			tx.Incr(ctx, m.totalInodesKey())
+		} else {
+			pipe.ZAdd(ctx, m.detachedNodes(), redis.Z{Member: dstIno.String(), Score: float64(time.Now().Unix())})
```
zhijian-pro, line 4300: This action needs to be done in a transaction and cannot be presented as a common method.

PR #3369, opened 2023-03-23
cmd/clone: check clone path

zhijian-pro:


cmd/clone.go
```diff
@@ -127,3 +144,20 @@ func clone(ctx *cli.Context) error {
...
 	return nil
 }
+func findMountpoint(dir string) (string, error) {
+	odir := dir
+	for dir != "" {
+		inode, err := utils.GetFileInode(dir)
+		if err != nil {
+			return "", fmt.Errorf("get inode of %s: %s", dir, err)
+		}
+		if inode == 1 {
```
SandyXSD, line 154: meta.RootInode

cmd/clone.go
```diff
@@ -127,3 +144,20 @@ func clone(ctx *cli.Context) error {
 	}
 	return nil
 }
+func findMountpoint(dir string) (string, error) {
```
SandyXSD, line 147: unify the implementation for other commands (warmup, rmr, openController, etc)

cmd/clone.go
```diff
@@ -84,6 +84,23 @@ func clone(ctx *cli.Context) error {
 	if err != nil {
 		return fmt.Errorf("abs of %s: %s", dst, err)
 	}
+
+	srcMp, err := findMountpoint(srcAbsPath)
+	if err != nil {
+		return err
+	}
+	dstMp, err := findMountpoint(filepath.Dir(dstAbsPath))
```
SandyXSD, line 92: no need, just check if dstAbsPath has srcMp as the prefix

zhijian-pro: It doesn't apply in this case
```
src=/tmp/mp1/dir1
dst=/tmp/mp10/dir2
```

cmd/clone.go
```diff
@@ -84,6 +83,23 @@ func clone(ctx *cli.Context) error {
...
+	if err != nil {
+		return err
+	}
+	if srcMp != dstMp {
+		return fmt.Errorf("the clone DST path should be at the same mount point as the SRC path")
+	}
+
+	if strings.HasPrefix(strings.TrimLeft(dstAbsPath, dstMp), strings.TrimLeft(srcAbsPath, srcMp)) {
```
davies, line 99: trim is not needed

davies: a  -> ab

SandyXSD (approved):

PR #3380, opened 2023-03-24
Add delimiter for mem storage

zyfjeff:
mem storage does not currently support delimiter, so some test codes cannot rely entirely on mem storage to mock
This PR adds delimiter support to mem storage

### DIFF
--- a/pkg/object/mem.go
+++ b/pkg/object/mem.go
@@ -140,7 +140,7 @@ func (m *memStore) List(prefix, marker, delimiter string, limit int64) ([]Object
 			if delimiter != "" {
 				remainString := strings.TrimPrefix(k, prefix)
 				pos := strings.Index(remainString, delimiter)
-				if pos != -1 && pos != len(remainString) {
+				if pos != -1 && pos != (len(remainString) - 1) {
 					commonPrefix := remainString[0:pos + 1]
 					if _, ok := commonPrefixsMap[commonPrefix]; ok {
 						continue
--- a/pkg/object/mem.go
+++ b/pkg/object/mem.go
@@ -139,9 +139,8 @@ func (m *memStore) List(prefix, marker, delimiter string, limit int64) ([]Object
 			o := m.objects[k]
 			if delimiter != "" {
 				remainString := strings.TrimPrefix(k, prefix)
-				pos := strings.Index(remainString, delimiter)
-				if pos != -1 && pos != (len(remainString) - 1) {
-					commonPrefix := remainString[0:pos + 1]
+				if pos := strings.Index(remainString, delimiter);pos != -1 {
+					commonPrefix := remainString[0 : pos + 1]
 					if _, ok := commonPrefixsMap[commonPrefix]; ok {
 						continue
 					}
@@ -152,9 +151,9 @@ func (m *memStore) List(prefix, marker, delimiter string, limit int64) ([]Object
 							time.Unix(0, 0),
 							strings.HasSuffix(commonPrefix, "/"),
 						},
-						"",
-						"",
-						0,
+						o.owner,
+						o.group,
+						o.mode,
 						false,
 					}
 					objs = append(objs, f)
--- a/pkg/object/mem.go
+++ b/pkg/object/mem.go
@@ -129,16 +129,40 @@ func (m *memStore) Delete(key string) error {
 }
 
 func (m *memStore) List(prefix, marker, delimiter string, limit int64) ([]Object, error) {
-	if delimiter != "" {
-		return nil, notSupportedDelimiter
-	}
 	m.Lock()
 	defer m.Unlock()
 
 	objs := make([]Object, 0)
+	commonPrefixsMap := make(map[string]bool, 0)
 	for k := range m.objects {
 		if strings.HasPrefix(k, prefix) && k > marker {
 			o := m.objects[k]
+			if delimiter != "" {
+				remainString := strings.TrimPrefix(k, prefix)
+				pos := strings.Index(remainString, delimiter)
+				if pos != -1 && pos != len(remainString) {
+					commonPrefix := remainString[0:pos + 1]
+					if _, ok := commonPrefixsMap[commonPrefix]; ok {
+						continue
+					}
+					f := &file{
+						obj{
+							prefix + commonPrefix,
+							0,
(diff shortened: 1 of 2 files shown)


zyfjeff:
**mutate-test/pkg/object/object_storage_test.go-1-1** The reason this test failed doesn't seem to have anything related with my change

ping @Suave

zyfjeff:
//skip mutate seems to have no effect

pkg/object/mem.go
```diff
@@ -129,16 +129,40 @@ func (m *memStore) Delete(key string) error {
...
 	defer m.Unlock()
 
 	objs := make([]Object, 0)
+	commonPrefixsMap := make(map[string]bool, 0)
 	for k := range m.objects {
 		if strings.HasPrefix(k, prefix) && k > marker {
 			o := m.objects[k]
+			if delimiter != "" {
+				remainString := strings.TrimPrefix(k, prefix)
+				pos := strings.Index(remainString, delimiter)
```
zhijian-pro, line 142: The range of pos is `[-1,len(remainString)-1]`, so `pos! = len(remainString)` is not necessary.

pkg/object/mem.go
```diff
@@ -129,16 +129,40 @@ func (m *memStore) Delete(key string) error {
...
+							time.Unix(0, 0),
+							strings.HasSuffix(commonPrefix, "/"),
+						},
+						"",
+						"",
+						0,
+						false,
+					}
+					objs = append(objs, f)
+					commonPrefixsMap[commonPrefix] = true
```
zhijian-pro, line 161: This is not necessary.

zhijian-pro (changes requested):

pkg/object/mem.go
zyfjeff: if pos != -1, but the pos == len(remainString), I don't want to handle this situation, it is considered a complete object and will be handled by the following logic.

pkg/object/mem.go
zyfjeff: `prefix/a/a1/`
`prefix/a/`

The above two objects cause `a/` to be counted repeatedly, so I use commonPrefixMaps to repeat it here

pkg/object/mem.go
zyfjeff: Sorry, I think I know, I should write pos != len(remainString) -1 here, I'll fix him

zhijian-pro:
> **mutate-test/pkg/object/object_storage_test.go-1-1** The reason this test failed doesn't seem to have anything related with my change

The mutate-test can be left unsolved.

pkg/object/mem.go
zhijian-pro: Just determine if the key exists

zhijian-pro: ok

pkg/object/mem.go
zyfjeff: fixed

zhijian-pro:
```
remainString=ab/ 
delimiter=/
pos=2
```
This case should in `commonPrefix`,so just remove `&& pos != (len(remainString)-1)`

zyfjeff: This is a full object that does not belong to commonPrefix. It will proceed to the following logic and return as a full object.

pkg/object/mem.go
```diff
@@ -129,16 +129,40 @@ func (m *memStore) Delete(key string) error {
...
+					continue
+				}
+			}
+
 			f := &file{
```
zyfjeff, line 166: `if pos == len(remainString) -1`, Will go here, return as a full object, so that you can get the real user group and other information

zyfjeff: @zhijian-pro

pkg/object/mem.go
zhijian-pro: The full object  as commonPrefix is ok.
this is a example :
aliyun oss
image

pkg/object/mem.go
zhijian-pro: The user group information is always zero. You'll notice that no one's reset it.

pkg/object/mem.go
zyfjeff: Sure,  I also return this full object here, just instead of going to commonPrefix logic to handle it, but as an object, which can fill more real object properties.
image

pkg/object/mem.go
zyfjeff: mtime and other info，In the future, we may fill in more real info

pkg/object/mem.go
zhijian-pro: It is best to use the same logic as other object stores for handling commprefix. The commprefix can also have more information, albeit with a zero value.
```
if delimiter != "" {
    remainString := strings.TrimPrefix(k, prefix)
    if pos := strings.Index(remainString, delimiter);pos != -1 {
	    commonPrefix := remainString[0 : pos+1]
	    if _, ok := commonPrefixsMap[commonPrefix]; ok {
		    continue
	    }
	    f := &file{
		    obj{
			    prefix + commonPrefix,
			    0,
			    time.Unix(0, 0),
			    strings.HasSuffix(commonPrefix, "/"),
		    },
		    o.owner,
		    o.group,
		    o.mode,
		    false,
	    }
	    objs = append(objs, f)
	    commonPrefixsMap[commonPrefix] = true
	    continue
    }
}
```

zyfjeff: Ok, I fix it.

pkg/object/mem.go
```diff
@@ -129,16 +129,39 @@ func (m *memStore) Delete(key string) error {
...
 	defer m.Unlock()
 
 	objs := make([]Object, 0)
+	commonPrefixsMap := make(map[string]bool, 0)
 	for k := range m.objects {
 		if strings.HasPrefix(k, prefix) && k > marker {
 			o := m.objects[k]
+			if delimiter != "" {
+				remainString := strings.TrimPrefix(k, prefix)
+				if pos := strings.Index(remainString, delimiter);pos != -1 {
```
SandyXSD, line 142: `go fmt` this

zyfjeff: fixed

PR #3393, opened 2023-03-28
statfs support subdir based on dir quota

Hexilee:


pkg/meta/base.go
```diff
@@ -819,8 +819,70 @@ func (m *baseMeta) cleanupSlices() {
...
+		if q == nil {
+			continue
+		}
+		if usage == nil {
+			usage = q
+		}
+		if q.MaxSpace > 0 {
+			ls := q.MaxSpace - q.UsedSpace
+			if ls < 0 {
+				logger.Errorf("invalid quota for inode %d: %+v", root, *q)
```
davies, line 853: it's OK,  left space is 0

pkg/meta/base.go
```diff
@@ -819,8 +819,70 @@ func (m *baseMeta) cleanupSlices() {
...
+func (m *baseMeta) StatFS(ctx Context, totalspace, availspace, iused, iavail *uint64, subdir bool) syscall.Errno {
 	defer m.timeit(time.Now())
+	if st := m.statRootFs(ctx, totalspace, availspace, iused, iavail); st != 0 {
+		return st
+	}
+	if m.root == RootInode || !subdir {
+		return 0
+	}
+	var usage *Quota
+	var leftSpace, leftInodes int64 = -1, -1
```
davies, line 831: use availspace as initial value

Hexilee: We need decide whether override `totalspace` or `iused` based on `leftSpace` or `leftInodes` is less than 0

pkg/meta/base.go
Hexilee: Ok

pkg/meta/base.go
Hexilee: My fault, fixed

PR #3400, opened 2023-03-29
meta/redis: fix transacion of copy_file_range

davies:
A pipeline inside a Redis transaction will end current transaction, so we can't use pipeline to read data.


SandyXSD:
From https://redis.io/docs/manual/transactions/
> When [EXEC](https://redis.io/commands/exec) is called, all keys are [UNWATCH](https://redis.io/commands/unwatch)ed, regardless of whether the transaction was aborted or not. Also when a client connection is closed, everything gets [UNWATCH](https://redis.io/commands/unwatch)ed.

PR #3418, opened 2023-03-31
cmd/format: remove testing dir after test

### LINKED ISSUE #3413: Ignore the testing folder created by test
When formating a volume, JuiceFS will put a object under "testing" to verify permission, the object is deleted later, but the "testing" folder may exists in some object storage.

We should delete the "testing" folder, and and ignore it when we list the bucket.

SandyXSD:
close #3413


amiga23:
:+1: works for me

PR #3423, opened 2023-04-03
improve parallelism for summary

davies:
Improve the parallelism for summary when the structure of directories is not perfect.

### DIFF
--- a/pkg/meta/utils.go
+++ b/pkg/meta/utils.go
@@ -342,7 +342,7 @@ func (m *baseMeta) GetSummary(ctx Context, inode Ino, summary *Summary, recursiv
 	if st := m.GetAttr(ctx, inode, &attr); st != 0 {
 		return st
 	}
-	if !recursive || attr.Typ != TypeDirectory {
+	if attr.Typ != TypeDirectory {
 		if attr.Typ == TypeDirectory {
 			summary.Dirs++
 		} else {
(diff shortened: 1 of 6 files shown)


pkg/meta/utils.go
```diff
@@ -494,116 +447,95 @@ func (m *baseMeta) GetTreeSummary(ctx Context, root *TreeSummary, depth, topN ui
...
+	}
+	wg.Wait()
+	select {
+	case err = <-errCh:
+		return err
+	default:
+	}
+
+	// pick top N
+	for _, c := range tree.Children {
```
Hexilee, line 520: In this implementation, the progress bar of summary does not work, as the root tree will be updated at the last.

pkg/meta/utils.go
```diff
@@ -354,134 +356,85 @@ func (m *baseMeta) GetSummary(ctx Context, inode Ino, summary *Summary, recursiv
...
-			})
-		}
-		if err := eg.Wait(); err != nil {
-			return errno(err)
+	var wg sync.WaitGroup
+	var errCh = make(chan syscall.Errno, 1)
+	for _, e := range entries {
+		if e.Attr.Typ == TypeDirectory {
+			atomic.AddUint64(&summary.Dirs, 1)
+			atomic.AddUint64(&summary.Size, uint64(align4K(0)))
```
Hexilee, line 396: Should not add this in non-strict mode.

pkg/meta/utils.go
```diff
@@ -494,116 +453,100 @@ func (m *baseMeta) GetTreeSummary(ctx Context, root *TreeSummary, depth, topN ui
 		root.Size += uint64(align4K(attr.Length))
 		return 0
 	}
-	root.Type = TypeDirectory
 	root.Dirs++
```
Hexilee, line 456: We should also update progress here.

PR #3426, opened 2023-04-03
obj/sftp: fix size of symlink in sftp

davies:


zhijian-pro:
Need to implement 'SupportSymlink' or you cannot create symlink in a single test.
image

SandyXSD (approved):

zhijian-pro (approved):

PR #3439, opened 2023-04-06
enhance disable updatedb to prevent concurrent modification

sjp00556:
1. juicefs csi driver has already add host /etc/updatedb.conf to mount pod, so we should also disable updatedb in k8s node,so i dropped os.Getpid() != 1
https://github.com/juicedata/juicefs-csi-driver/commit/77dad895a5100b4835982f86dd69d9d4fe026a8f

2. add flock to prevent multiple juicefs clients writing  updatedb.conf at the same time

### DIFF
--- a/cmd/mount_unix.go
+++ b/cmd/mount_unix.go
@@ -162,12 +162,16 @@ func disableUpdatedb() {
 
 	// obtain exclusive and not block flock
 	if err := syscall.Flock(int(file.Fd()), syscall.LOCK_EX|syscall.LOCK_NB); err != nil {
-		return
+		if err == syscall.EAGAIN {
+			return
+		}
+	} else {
+		defer func() {
+			// release flock
+			_ = syscall.Flock(int(file.Fd()), syscall.LOCK_UN)
+		}()
 	}
-	defer func() {
-		// release flock
-		_ = syscall.Flock(int(file.Fd()), syscall.LOCK_UN)
-	}()
+
 	data, err := io.ReadAll(file)
 	if err != nil {
 		return
--- a/cmd/mount_unix.go
+++ b/cmd/mount_unix.go
@@ -21,6 +21,7 @@ package cmd
 
 import (
 	"bytes"
+	"io"
 	"os"
 	"path"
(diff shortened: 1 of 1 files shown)


cmd/mount_unix.go
```diff
@@ -153,7 +154,21 @@ func mount_flags() []cli.Flag {
...
+	}
+	defer file.Close()
+
+	// obtain exclusive and not block flock
+	if err := syscall.Flock(int(file.Fd()), syscall.LOCK_EX|syscall.LOCK_NB); err != nil {
+		return
+	}
+	defer func() {
+		// release flock
+		_ = syscall.Flock(int(file.Fd()), syscall.LOCK_UN)
```
davies, line 169: flock should be released when the file is closed

cmd/mount_unix.go
```diff
@@ -153,7 +154,21 @@ func mount_flags() []cli.Flag {
...
 	path := "/etc/updatedb.conf"
-	data, err := os.ReadFile(path)
+	file, err := os.Open(path)
+	if err != nil {
+		return
+	}
+	defer file.Close()
+
+	// obtain exclusive and not block flock
+	if err := syscall.Flock(int(file.Fd()), syscall.LOCK_EX|syscall.LOCK_NB); err != nil {
```
davies, line 164: Is it possible that Flock is not supported?

cmd/mount_unix.go
sjp00556: syscall.LOCK_UN will release flock

cmd/mount_unix.go
sjp00556: some fs maybe not support flock, It's very rare.  I added condition if err is not syscall.EAGAIN (resource temporarily unavailable), it not use flock

davies:
@sjp00556 Before this PR, juicefs can already update /etc/updatedb.conf in K8S, because the pid of juicefs is not 1. Btw, the pid check is not needed. 

Thanks!

sjp00556:
got it

PR #3458, opened 2023-04-10
meta: batch update quotas

SandyXSD:


pkg/meta/tkv.go
```diff
@@ -2691,23 +2695,26 @@ func (m *kvMeta) doLoadQuotas(ctx Context) (map[Ino]*Quota, error) {
...
-		rawQ := tx.get(key)
-		if len(rawQ) == 0 {
-			logger.Warnf("No quota for inode %d, skip flushing", inode)
-			return nil
-		}
-		if len(rawQ) != 32 {
-			return fmt.Errorf("invalid quota value: %v", rawQ)
+		for _, inode := range inodes {
+			key := m.dirQuotaKey(inode)
+			rawQ := tx.get(key)
```
davies, line 2702: tx.gets() will be faster

pkg/meta/base.go
```diff
@@ -708,21 +710,41 @@ func (m *baseMeta) flushQuotas() {
...
-		// FIXME: merge
-		for ino, q := range quotas {
-			if err := m.en.doFlushQuota(Background, ino, q.newSpace, q.newInodes); err != nil {
-				logger.Warnf("Flush quota of inode %d: %s", ino, err)
+
+		for ino := range quotas {
+			inodes = append(inodes, ino)
+		}
+		batch := 1000
+		if m.en.enType() == "tkv" {
```
davies, line 718: the number of active quota should be large, and updating 100 should not be much slower than updating 20, since all the quotas should belong to single TiKV region, and we can get/put them in single batch.

That means we don't need the batch here.

PR #3465, opened 2023-04-11
cmd/mount: add `single-mode` option to skips the heartbeat and metadata synchronization

### LINKED ISSUE #3329: Add the ability to disable client heartbeat

Please consider making `juicefs mount --heartbeat 0 ...` to disable the client heartbeat feature.

**Why is this needed**:

My use case:

```bash
$ juicefs format \
	--storage sqlite3 \
	--bucket /litestream/data-dir.jfs-bucket.db \
	sqlite3:///litestream/data-dir.jfs-meta.db \
	data-dir
$ litestream replicate -sync-interval 1 /litestream/data-dir.jfs-meta.db s3://mybkt/data-dir.jfs-meta.db &
$ litestream replicate -sync-interval 1 /litestream/data-dir.jfs-bucket.db s3://mybkt/data-dir.jfs-bucket.db &
$ juicefs mount sqlite3:///litestream/data-dir.jfs-meta.db /data
```

The files in `/data` rarely change, but I still expect JuiceFS to reflect to those two databases when it does, and [Litestream](https://litestream.io) to stream WALs to my S3 bucket in time. They actually do work pretty well together. Until I got my S3 bill.

Again, the files in `/data` are rarely change, so I'd like JuiceFS to be quiet until it does, which means it shouldn't be doing any writes to those two databases. But here comes the problem. It turns out that there is a heartbeat feature for the client here. The client always send heartbeats to meta storage at regular intervals (which defaults to 12s, and fixed in the [1, 60] interval). This means that by default every 12s Litestream will stream two WALs to my S3 bucket (I don't know why two). In other words, two S3 class A operations every 12s, or 432,000 class A operations every month, for nothing. And I have hundreds of the above setups currently active...

I know that for network meta storage, such as Redis, the client heartbeat feature is indispensable. But when it comes to SQLite meta storage, I believe this feature is unnecessary, given that there is only local disk I/O.

So I think it's reasonable to add the ability to disable the client heartbeat.

SandyXSD:
close #3329 

We need the heartbeat as a ticker to reload new configurations (conf, quota, etc.). So instead of setting heartbeat to 0, we may add a new single node mode, which skips the management of sessions and synchronization of background tasks.

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -195,6 +195,10 @@ func clientFlags() []cli.Flag {
 			Name:  "no-bgjob",
 			Usage: "disable background jobs (clean-up, backup, etc.)",
 		},
+		&cli.BoolFlag{
+			Name:  "single-mode",
+			Usage: "enable single node mode, which skips the heartbeat and metadata synchronization",
+		},
 		&cli.Float64Flag{
 			Name:  "open-cache",
 			Value: 0.0,
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -353,6 +353,7 @@ func getMetaConf(c *cli.Context, mp string, readOnly bool) *meta.Config {
 	conf.SkipDirNlink = c.Int("skip-dir-nlink")
 	conf.ReadOnly = readOnly
 	conf.NoBGJob = c.Bool("no-bgjob")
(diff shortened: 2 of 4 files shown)


davies:
We don't need single mode for now.

PR #3466, opened 2023-04-11
cmd/fsck: returns non-zero status code if there's error occurred during the fsck

### LINKED ISSUE #3154: Respect the result of repair
`Check` returns 0 even if it fails to repair the node, see: https://github.com/juicedata/juicefs/actions/runs/3844692510/jobs/6548964413#step:7:1022

SandyXSD:
close #3154


pkg/meta/base.go
```diff
@@ -1721,6 +1721,7 @@ func (m *baseMeta) Check(ctx Context, fpath string, repair bool, recursive bool,
 		attr.Parent = parent
 	}
 
+	var done = true
```
davies, line 1724: hasError

davies (approved):

PR #3491, opened 2023-04-18
CI: speed up unit_test

zhijian-pro:


zhijian-pro:
The bottleneck is metadata testing for SQL type databases.

PR #3501, opened 2023-04-20
add method tag for metadata access monitoring

shichanglin5:
a label(method anme) is added to the `opDist` metrics collector of metadata operations at `pkg/meta/base.go:217`

so that we can clearly view the time duration and qps of different methods through monitoring, which is useful for troubleshooting problems and analyzing performance bottlenecks.

### DIFF
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -174,7 +174,7 @@ type baseMeta struct {
 	usedInodesG prometheus.Gauge
 	txDist      prometheus.Histogram
 	txRestart   prometheus.Counter
-	opDist      prometheus.Histogram
+	opDist      *prometheus.HistogramVec
 
 	en engine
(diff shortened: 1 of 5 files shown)


SandyXSD:
Usually we use access log and `juicefs profile` if we want to know details of each operation.

SandyXSD:
@shichanglin5 Could you keep the original `opDist` (so `juicefs stats` still works), and add a new HistogramVec for the metrics with tags?

shichanglin5:
> @shichanglin5 Could you keep the original `opDist` (so `juicefs stats` still works), and add a new HistogramVec for the metrics with tags?

This way is feasible, but there will be two opDist; How about modifying the stats code and merging the different label statistics of this opDist?

davies (approved):

Issue #3240, opened 2023-02-15
Support using atime to track file access


SandyXSD:
**What would you like to be added**:
Add a new option to enable atime.

**Why is this needed**:

PR #3521, opened 2023-04-25
vfs: introduce --atime-mode option to control atime update behavior

eryugey:
Currently juicefs doesn't update atime on access, so introduce --atime-mode option to select how to update atime. And support three modes:

- noatime: don't update atime
- relatime: update atime relative to motify or change time
- strictatime: always update atime

Fixes: #3240
Signed-off-by: Eryu Guan <eguan@linux.alibaba.com>

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -214,6 +214,11 @@ func clientFlags() []cli.Flag {
 			Name:  "subdir",
 			Usage: "mount a sub-directory as root",
 		},
+		&cli.StringFlag{
+			Name:  "atime-mode",
+			Value: "noatime",
+			Usage: "when to update atime, supported mode includes: noatime, relatime, strictatime",
+		},
 	}
 }
 
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -358,6 +358,13 @@ func getMetaConf(c *cli.Context, mp string, readOnly bool) *meta.Config {
 	conf.Heartbeat = duration(c.String("heartbeat"))
 	conf.MountPoint = mp
 	conf.Subdir = c.String("subdir")
+
+	atimeMode := c.String("atime-mode")
+	if atimeMode != meta.RelAtime && atimeMode != meta.StrictAtime && atimeMode != meta.NoAtime {
+		logger.Warnf("unknown atime-mode \"%s\", changed to %s", atimeMode, meta.NoAtime)
+		atimeMode = meta.NoAtime
+	}
+	conf.AtimeMode = atimeMode
 	return conf
 }
 
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -107,7 +107,7 @@ type engine interface {
 	scanPendingFiles(Context, pendingFileScan) error
 
 	GetSession(sid uint64, detail bool) (*Session, error)
-	touchAtime(ctx Context, inode Ino) syscall.Errno
+	touchAtime(ctx Context, inode Ino, attr *Attr) syscall.Errno
 }
 
 type trashSliceScan func(ss []Slice, ts int64) (clean bool, err error)
@@ -1563,15 +1563,15 @@ func (m *baseMeta) Close(ctx Context, inode Ino) syscall.Errno {
 }
 
 func (m *baseMeta) Readdir(ctx Context, inode Ino, plus uint8, entries *[]*Entry) (rerr syscall.Errno) {
+	var attr Attr
 	defer func() {
 		if rerr == 0 {
-			if err := m.en.touchAtime(ctx, inode); err != 0 {
+			if err := m.en.touchAtime(ctx, inode, &attr); err != 0 {
 				logger.Warnf("readdir %v update atime: %s", inode, err)
 			}
 		}
 	}()
 	inode = m.checkRoot(inode)
-	var attr Attr
 	if err := m.GetAttr(ctx, inode, &attr); err != 0 {
 		return err
 	}
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -1495,10 +1495,19 @@ func (m *baseMeta) Rename(ctx Context, parentSrc Ino, nameSrc string, parentDst 
 	return st
 }
 
-func (m *baseMeta) Open(ctx Context, inode Ino, flags uint32, attr *Attr) syscall.Errno {
+func (m *baseMeta) Open(ctx Context, inode Ino, flags uint32, attr *Attr) (rerr syscall.Errno) {
 	if m.conf.ReadOnly && flags&(syscall.O_WRONLY|syscall.O_RDWR|syscall.O_TRUNC|syscall.O_APPEND) != 0 {
 		return syscall.EROFS
 	}
+
+	defer func() {
+		if rerr == 0 {
+			if err := m.en.touchAtime(ctx, inode, attr); err != 0 {
+				logger.Warnf("open %v update atime: %s", inode, err)
+			}
+		}
+	}()
+
 	if m.conf.OpenCache > 0 && m.of.OpenCheck(inode, attr) {
 		return 0
 	}
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -107,6 +107,7 @@ type engine interface {
 	scanPendingFiles(Context, pendingFileScan) error
 
 	GetSession(sid uint64, detail bool) (*Session, error)
+	touchAtime(ctx Context, inode Ino) syscall.Errno
 }
 
 type trashSliceScan func(ss []Slice, ts int64) (clean bool, err error)
@@ -1561,7 +1562,14 @@ func (m *baseMeta) Close(ctx Context, inode Ino) syscall.Errno {
 	return 0
 }
 
-func (m *baseMeta) Readdir(ctx Context, inode Ino, plus uint8, entries *[]*Entry) syscall.Errno {
+func (m *baseMeta) Readdir(ctx Context, inode Ino, plus uint8, entries *[]*Entry) (rerr syscall.Errno) {
+	defer func() {
+		if rerr == 0 {
+			if err := m.en.touchAtime(ctx, inode); err != 0 {
+				logger.Warnf("readdir %v update atime: %s", inode, err)
+			}
+		}
+	}()
 	inode = m.checkRoot(inode)
 	var attr Attr
 	if err := m.GetAttr(ctx, inode, &attr); err != 0 {
--- a/pkg/meta/redis.go
+++ b/pkg/meta/redis.go
@@ -4249,39 +4249,58 @@ func (m *redisMeta) doAttachDirNode(ctx Context, parent Ino, dstIno Ino, name st
 }
 
 // caller makes sure inode is not special inode.
-func (m *redisMeta) touchAtime(ctx Context, ino Ino, cur *Attr) syscall.Errno {
+func (m *redisMeta) touchAtime(ctx Context, ino Ino, cur *Attr) (rerr syscall.Errno) {
 	if (m.conf.AtimeMode != StrictAtime && m.conf.AtimeMode != RelAtime) || m.conf.ReadOnly {
 		return 0
 	}
 
-	attr := &Attr{}
-	return errno(m.txn(ctx, func(tx *redis.Tx) error {
-		if cur == nil {
-			if !m.of.Check(ino, attr) {
-				a, err := tx.Get(ctx, m.inodeKey(ino)).Bytes()
-				if err != nil {
-					return err
-				}
-				m.parseAttr(a, attr)
-			}
-		} else {
-			attr = cur
-		}
+	var attr *Attr
+	newAttr := &Attr{}
+	if cur != nil {
+		attr = cur
+	} else if m.of.Check(ino, newAttr) {
+		attr = newAttr
+	}
 
-		now := time.Now()
+	updated := false
+	defer func() {
+		if rerr == 0 && m.of.IsOpen(ino) && updated {
+			m.of.Update(ino, attr)
+		}
+	}()
+
+	now := time.Now()
+
+	// Already got attr, update atime without transaction
+	if attr != nil {
+		if !m.atimeNeedsUpdate(attr, now) {
+			return 0
+		}
+		attr.Atime = now.Unix()
+		attr.Atimensec = uint32(now.Nanosecond())
+		updated = true
+		return errno(m.rdb.Set(ctx, m.inodeKey(ino), m.marshal(attr), 0).Err())
+	}
+
+	// Should get attr first, do it in transaction
+	return errno(m.txn(ctx, func(tx *redis.Tx) error {
+		attr = newAttr
+		a, err := tx.Get(ctx, m.inodeKey(ino)).Bytes()
+		if err != nil {
+			return err
+		}
+		m.parseAttr(a, attr)
+
 		if !m.atimeNeedsUpdate(attr, now) {
 			return nil
 		}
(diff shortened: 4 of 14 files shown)


pkg/vfs/vfs.go
```diff
@@ -995,6 +1009,54 @@ func (v *VFS) cleanupModified() {
...
+func (v *VFS) atimeNeedsUpdate(attr *Attr, now time.Time) bool {
+	if v.Conf.AtimeMode == RelAtime && relatimeNeedUpdate(attr, now) {
+		return true
+	}
+
+	return v.Conf.AtimeMode == StrictAtime && !now.Equal(time.Unix(attr.Atime, int64(attr.Atimensec)))
+}
+
+// caller makes sure inode is not special inode.
+func (v *VFS) touchAtime(ctx Context, ino Ino) syscall.Errno {
```
davies, line 1021: I think we should move this into meta engine for ACID and performance

pkg/vfs/vfs_unix.go
```diff
@@ -94,6 +94,11 @@ func (v *VFS) Access(ctx Context, ino Ino, mask int) (err syscall.Errno) {
 	}
 
 	err = v.Meta.Access(ctx, ino, uint8(mmask), nil)
+	if err == 0 {
+		if aerr := v.touchAtime(ctx, ino); aerr != 0 {
```
davies, line 98: Access is used to check permission, we should not update atime

eryugey: True, ext4 doesn't update atime on access(2) as well. Fixed

eryugey:
Any advise on how tho fix this markdown lint error?
image

SandyXSD:
> Any advise on how tho fix this markdown lint error?

Add one space around the numbers, and between Chinese word and English word.

pkg/vfs/vfs.go
eryugey: Fixed.

pkg/meta/redis.go
```diff
@@ -2113,7 +2113,14 @@ func (m *redisMeta) doDeleteSustainedInode(sid uint64, inode Ino) error {
 	return err
 }
 
-func (m *redisMeta) Read(ctx Context, inode Ino, indx uint32, slices *[]Slice) syscall.Errno {
+func (m *redisMeta) Read(ctx Context, inode Ino, indx uint32, slices *[]Slice) (rerr syscall.Errno) {
+	defer func() {
+		if rerr == 0 {
+			if err := m.touchAtime(ctx, inode); err != 0 {
```
SandyXSD, line 2119: We'd better cache the result to avoid getting attr from the server for each Read.

eryugey: get attr from open cache and read meta server only when it's not cached?

eryugey: Appended a new patch to do so.

cmd/flags.go
```diff
@@ -214,6 +214,10 @@ func clientFlags() []cli.Flag {
 			Name:  "subdir",
 			Usage: "mount a sub-directory as root",
 		},
+		&cli.StringFlag{
+			Name:  "atime-mode",
```
SandyXSD, line 218: set `Value: "noatime"`, which indicates the default value.

cmd/mount.go
```diff
@@ -358,6 +358,13 @@ func getMetaConf(c *cli.Context, mp string, readOnly bool) *meta.Config {
 	conf.Heartbeat = duration(c.String("heartbeat"))
 	conf.MountPoint = mp
 	conf.Subdir = c.String("subdir")
+
+	atimeMode := c.String("atime-mode")
+	if atimeMode != meta.RelAtime && atimeMode != meta.StrictAtime {
+		logger.Warnf("unknown atime-mode \"%s\", changed to %s", atimeMode, meta.NoAtime)
```
SandyXSD, line 364: don'n warning if it's the default value.

pkg/meta/base.go
```diff
@@ -1561,7 +1562,14 @@ func (m *baseMeta) Close(ctx Context, inode Ino) syscall.Errno {
...
 
-func (m *baseMeta) Readdir(ctx Context, inode Ino, plus uint8, entries *[]*Entry) syscall.Errno {
+func (m *baseMeta) Readdir(ctx Context, inode Ino, plus uint8, entries *[]*Entry) (rerr syscall.Errno) {
+	defer func() {
+		if rerr == 0 {
+			if err := m.en.touchAtime(ctx, inode); err != 0 {
+				logger.Warnf("readdir %v update atime: %s", inode, err)
+			}
+		}
+	}()
```
SandyXSD, line 1572: use the attr got in the later `m.GetAttr`

pkg/meta/redis.go
```diff
@@ -2113,7 +2113,14 @@ func (m *redisMeta) doDeleteSustainedInode(sid uint64, inode Ino) error {
...
+				logger.Warnf("read %v update atime: %s", inode, err)
+			}
+		}
+	}()
```
SandyXSD, line 2123: The attr must be cached in the `of`(openfiles) since the file is still open.

pkg/meta/sql.go
```diff
@@ -3900,3 +3907,37 @@ func (m *dbMeta) doAttachDirNode(ctx Context, parent Ino, inode Ino, name string
 		return err
 	}, parent))
 }
+
+func (m *dbMeta) touchAtime(ctx Context, ino Ino) syscall.Errno {
```
SandyXSD, line 3911: We'd better make the decision of "shall we update the atime" based on different operations (open, readdir, read, etc.), not in a general function. This is because metadata transactions are expensive, and should be avoided whenever possible.

SandyXSD: In other words, this function can be used to UPDATE the atime, while the CHECK should be done before calling it.

pkg/meta/utils.go
```diff
@@ -553,3 +558,34 @@ func (m *baseMeta) getTreeSummary(ctx Context, tree *TreeSummary, depth, topN ui
 	}
 	return 0
 }
+
+func (m *baseMeta) atimeNeedsUpdate(attr *Attr, now time.Time) bool {
+	if m.conf.AtimeMode == RelAtime && relatimeNeedUpdate(attr, now) {
+		return true
+	}
+
+	return m.conf.AtimeMode == StrictAtime && !now.Equal(time.Unix(attr.Atime, int64(attr.Atimensec)))
```
SandyXSD, line 567: Maybe we don't need that strict even in `StrictAtime` mode to reduce updates, i.e, compare `attr.Atime` without `attr.Atimensec` should be enough.

pkg/meta/utils.go
```diff
@@ -553,3 +558,34 @@ func (m *baseMeta) getTreeSummary(ctx Context, tree *TreeSummary, depth, topN ui
...
+		return true
+	}
+
+	// Is ctime younger than atime? If yes, update atime
+	if time.Unix(attr.Ctime, int64(attr.Ctimensec)).After(atime) {
+		return true
+	}
+
+	// Is the previous atime value older than a day? If yes, update atime
+	if now.Sub(atime) > 24*time.Hour {
```
SandyXSD, line 586: just `return now.Sub(atime) > 24*time.Hour` here.

SandyXSD:
We should also check and update the atime in `Open` and `ReadLink`.

cmd/flags.go
eryugey: Fixed

cmd/mount.go
eryugey: Fixed

pkg/meta/base.go
eryugey: Fixed

pkg/meta/redis.go
eryugey: I think it's already been updated by of.Update() in touchAtime().

pkg/meta/sql.go
eryugey: OK, I updated touchAtime() and avoided transaction if possible, but still in check & update manner.

pkg/meta/utils.go
eryugey: Ok, I changed it to update atime only when last update is more than 1s ago.

pkg/meta/utils.go
eryugey: Fixed

eryugey:
> We should also check and update the atime in `Open` and `ReadLink`.

Added touchAtime() in Open() as well.

pkg/meta/redis.go
```diff
@@ -4240,3 +4247,57 @@ func (m *redisMeta) doAttachDirNode(ctx Context, parent Ino, dstIno Ino, name st
...
+	var attr *Attr
+	newAttr := &Attr{}
+	if cur != nil {
+		attr = cur
+	} else if m.of.Check(ino, newAttr) {
+		attr = newAttr
+	}
+
+	defer func() {
+		if rerr == 0 && m.of.IsOpen(ino) {
```
SandyXSD, line 4266: do `Update` only when the atime is really changed.

pkg/meta/sql.go
```diff
@@ -3900,3 +3907,60 @@ func (m *dbMeta) doAttachDirNode(ctx Context, parent Ino, inode Ino, name string
...
+		s := m.db.NewSession()
+		defer s.Close()
+		_, err := s.Cols("atime").Update(&curNode, &node{Inode: ino})
+		return errno(err)
+	}
+
+	// Should get attr first, do it in transaction
+	return errno(m.txn(func(s *xorm.Session) error {
+		attr = newAttr
+		if !m.of.Check(ino, attr) {
```
SandyXSD, line 3948: no need to check again, just like that in Redis

pkg/meta/tkv.go
```diff
@@ -3380,3 +3387,37 @@ func (m *kvMeta) doAttachDirNode(ctx Context, parent Ino, inode Ino, name string
 		return nil
 	}, parent))
 }
+
+func (m *kvMeta) touchAtime(_ctx Context, ino Ino, cur *Attr) syscall.Errno {
+	if (m.conf.AtimeMode != StrictAtime && m.conf.AtimeMode != RelAtime) || m.conf.ReadOnly {
+		return 0
+	}
+
+	attr := &Attr{}
```
SandyXSD, line 3396: adjust it like that in RedisMeta

pkg/meta/sql.go
eryugey: Fixed

pkg/meta/tkv.go
eryugey: I agree, but I don't know how to update a key in tkv without transaction, seems only `tikvTxn` implements `set` method, which calls `KVTxn.Set()`. I didn't find other set implementations outside transaction.

eryugey: > 

Ok, let me try.

eryugey: Fixed

pkg/meta/redis.go
```diff
@@ -4240,3 +4247,60 @@ func (m *redisMeta) doAttachDirNode(ctx Context, parent Ino, dstIno Ino, name st
...
+
+	now := time.Now()
+
+	// Already got attr, update atime without transaction
+	if attr != nil {
+		if !m.atimeNeedsUpdate(attr, now) {
+			return 0
+		}
+		attr.Atime = now.Unix()
+		attr.Atimensec = uint32(now.Nanosecond())
```
davies, line 4280: Update the ctime as well

pkg/meta/redis.go
```diff
@@ -4240,3 +4247,60 @@ func (m *redisMeta) doAttachDirNode(ctx Context, parent Ino, dstIno Ino, name st
...
+		if err != nil {
+			return err
+		}
+		m.parseAttr(a, attr)
+
+		if !m.atimeNeedsUpdate(attr, now) {
+			return nil
+		}
+		attr.Atime = now.Unix()
+		attr.Atimensec = uint32(now.Nanosecond())
```
davies, line 4298: update ctime

pkg/meta/redis.go
```diff
@@ -4240,3 +4247,60 @@ func (m *redisMeta) doAttachDirNode(ctx Context, parent Ino, dstIno Ino, name st
...
+
+	// Already got attr, update atime without transaction
+	if attr != nil {
+		if !m.atimeNeedsUpdate(attr, now) {
+			return 0
+		}
+		attr.Atime = now.Unix()
+		attr.Atimensec = uint32(now.Nanosecond())
+		updated = true
+		return errno(m.rdb.Set(ctx, m.inodeKey(ino), m.marshal(attr), 0).Err())
```
davies, line 4282: We can't call rdb.Set without a transaction, because it may overwrite the change in attr from others.

pkg/meta/redis.go
eryugey: Local filesystems like ext4 will just update atime, I followed that behavior. I just want to confirm that we really want to update ctime as well, right?

pkg/meta/redis.go
eryugey: @SandyXSD suggested to avoid transaction when possible, so you mean only redis always requires transactions? tkv and sql may still avoid transactions when possible?

davies: We can't only update the atime of Attr, so we can't avoid the transaction.

pkg/meta/redis.go
davies: I, see.

pkg/meta/redis.go
eryugey: Fixed, always update atime in transaction, this basically reverts "meta: avoid transaction if possible in touchAtime()".

SandyXSD: My bad. The actual expected behavior should be: avoid updating whenever possible.
In other words, the procedure is like:
1. if we already have the attr in memory, use it to decide whether the atime should be updated
2. if not cached, or the atime SHOULD be updated, use a transaction to update the atime
3. and in the transaction, we should get the latest attr first, and update only the atime if needed

pkg/meta/redis.go
```diff
@@ -4240,3 +4247,49 @@ func (m *redisMeta) doAttachDirNode(ctx Context, parent Ino, dstIno Ino, name st
...
+	}
+
+	var attr *Attr
+	newAttr := &Attr{}
+	if cur != nil {
+		attr = cur
+	} else if m.of.Check(ino, newAttr) {
+		attr = newAttr
+	}
+
```
SandyXSD, line 4264: check `atimeNeedsUpdate` here first, and return directly if the update is not needed.

pkg/meta/redis.go
```diff
@@ -4240,3 +4247,49 @@ func (m *redisMeta) doAttachDirNode(ctx Context, parent Ino, dstIno Ino, name st
...
+	updated := false
+	defer func() {
+		if rerr == 0 && m.of.IsOpen(ino) && updated {
+			m.of.Update(ino, attr)
+		}
+	}()
+
+	return errno(m.txn(ctx, func(tx *redis.Tx) error {
+		if attr == nil {
```
SandyXSD, line 4273: remove the `if` condition, and always get the latest attribute in the transaction

pkg/meta/redis.go
eryugey: Fixed

pkg/meta/redis.go
eryugey: Fixed

SandyXSD (approved):

eryugey:
Thanks all for the reviews!

Issue #3523, opened 2023-04-25
there are some data races when run go test -race


pigletfly:
**What happened**:

there are some data races when run go test -race ./pkg/... on main branch

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**

**Environment**:
- JuiceFS version (use `juicefs --version`) or Hadoop Java SDK version: m
- Cloud provider or hardware configuration running JuiceFS:
- OS (e.g `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Object storage (cloud provider and region, or self maintained):
- Metadata engine info (version, cloud provider managed or self maintained):
- Network connectivity (JuiceFS to metadata engine, JuiceFS to object storage):
- Others:

pigletfly:
the directories are pkg/chunk ,pkg/fs.

SandyXSD:
The races found are caused by test code or acceptable, we'll just leave them intact.

PR #3524, opened 2023-04-25
fix possible data race in pkg/fs

pigletfly:
part of #3523

### DIFF
--- a/pkg/fs/fs.go
+++ b/pkg/fs/fs.go
@@ -40,9 +40,6 @@ import (
 
 var logger = utils.GetLogger("juicefs")
 
-var checkAccessFile = time.Minute
-var rotateAccessLog int64 = 300 << 20 // 300 MiB
-
 type Ino = meta.Ino
 type Attr = meta.Attr
 type LogContext = vfs.LogContext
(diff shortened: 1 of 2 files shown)


pigletfly:
> Please modify only the code that is necessary, and use another standalone PR if you want to do some style refactor.

fixed, please take another look

pkg/fs/fs.go
```diff
@@ -299,16 +300,16 @@ func (fs *FileSystem) flushLog(f *os.File, logBuffer chan string, path string) {
...
 		}
-		if lastcheck.Add(checkAccessFile).After(time.Now()) {
+		if lastcheck.Add(fs.checkAccessFile).After(time.Now()) {
 			continue
 		}
 		lastcheck = time.Now()
 		var fi os.FileInfo
 		fi, err = f.Stat()
-		if err == nil && fi.Size() > rotateAccessLog {
+		if err == nil && fi.Size() > atomic.LoadInt64(&rotateAccessLog) {
```
davies, line 309: The race happens only on tests, we could ignore it.
Even we want to fix that, it's better to save the current value of rotateAccessLog into FileSystem, rather than using atomic.

pigletfly: OK

SandyXSD (approved):

PR #3527, opened 2023-04-25
quota: implement check subcommand

Hexilee:


davies (approved):

pkg/meta/base.go
```diff
@@ -818,7 +819,39 @@ func (m *baseMeta) HandleQuota(ctx Context, cmd uint8, dpath string, quotas map[
...
+			return st
+		}
+		usedInodes := int64(sum.Dirs+sum.Files) - 1
+		usedSpace := int64(sum.Size) - align4K(0) // quota ignore root dir
+		if q.UsedInodes == usedInodes && q.UsedSpace == usedSpace {
+			logger.Infof("quota of %s is consistent", dpath)
+			quotas[dpath] = q
+			return nil
+		}
+		logger.Errorf(
```
SandyXSD, line 841: `warnf` is enough

pkg/meta/base.go
```diff
@@ -818,7 +819,39 @@ func (m *baseMeta) HandleQuota(ctx Context, cmd uint8, dpath string, quotas map[
...
+			"%s: quota(%s, %s) != summary(%s, %s)", dpath,
+			humanize.Comma(q.UsedInodes), humanize.IBytes(uint64(q.UsedSpace)),
+			humanize.Comma(usedInodes), humanize.IBytes(uint64(usedSpace)),
+		)
+		if repair {
+			q.UsedInodes = usedInodes
+			q.UsedSpace = usedSpace
+			quotas[dpath] = q
+			logger.Info("reparing...")
+			return m.en.doSetQuota(ctx, inode, q, false)
```
SandyXSD, line 851: `doSetQuota` will not update the used stats

SandyXSD (changes requested):

pkg/meta/base.go
Hexilee: replaced

pkg/meta/base.go
Hexilee: It seems the used stats will be updated if the `create` flag is true.

Hexilee: done

SandyXSD (approved):

Issue #3526, opened 2023-04-25
Support setting the path for hdfs


kyungwan-nam:
**What would you like to be added**:

juices store data on the subdirectory of the root only (e.g. ``/myjfs``) when setting hdfs as object storage.
https://juicefs.com/docs/community/how_to_setup_object_storage#hdfs
It looks good if we can set the path for object storage like ``--bucket namenode1:8020/user/kyungwan-nam``

```
juicefs format \
    --storage hdfs \
    --bucket namenode1:8020/user/kyungwan-nam \
    --access-key hdfs \
    ... \
    myjfs
```

**Why is this needed**:

I have a multitenant HDFS cluster, and a lot of users share it.
each user has a home directory (e.g. ``/user/a/``, ``/user/b/``) separately. and the other path is not allowed.
thus, I need to change it to a path with permission, not the subdirectory of the root.

PR #3528, opened 2023-04-25
Support setting the path for hdfs

kyungwan-nam:
https://github.com/juicedata/juicefs/issues/3526


tangyoupeng:
Could you also add some docs for this pr?
https://github.com/juicedata/juicefs/blob/main/docs/en/guide/how_to_set_up_object_storage.md

kyungwan-nam:
@tangyoupeng 
Thank you for reviewing.
I updated the document about it.

PR #3536, opened 2023-04-27
object/gs: remove storage class settings

zhijian-pro:
image

The `regional` is an illegal value, which seems to be a bug

### DIFF
--- a/pkg/object/gs.go
+++ b/pkg/object/gs.go
@@ -76,9 +76,8 @@ func (g *gs) Create() error {
 	}
 
 	err := g.client.Bucket(g.bucket).Create(ctx, projectID, &storage.BucketAttrs{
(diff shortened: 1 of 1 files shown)


pkg/object/gs.go
```diff
@@ -76,9 +76,8 @@ func (g *gs) Create() error {
 	}
 
 	err := g.client.Bucket(g.bucket).Create(ctx, projectID, &storage.BucketAttrs{
-		Name:         g.bucket,
-		StorageClass: "regional",
```
davies, line 80: https://cloud.google.com/storage/docs/storage-classes#legacy

zhijian-pro: Standard type is better

PR #3538, opened 2023-04-27
object: add dummy object store

eryugey:
Introduce a dummy object store as juicefs backend, which might be useful for developers to write/debug non-object storage related code such as disk cache.

### DIFF
--- a/pkg/object/dummy.go
+++ b/pkg/object/dummy.go
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2023 Alibaba Cloud, Inc. or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


SandyXSD:
Usually the local filesystem (type 'file') is simple enough as an object store for developing use. Besides, it's more feature completed (have real data write/read).

eryugey:
> 

That's true, but sometimes the disk is slow on dev machine, and I don't want to waste the time and disk space for nothing. A dummy storage would be more efficient.

davies:
We already have in-memory object storage could be used for develop or test.

pkg/object/dummy.go
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright 2023 Alibaba Cloud, Inc. or its affiliates.
```
davies, line 2: The CLA means all the contributed code will be owned by Juicedata Inc, we won't accept any code copyrighted by others, except existing library.

PR #3579, opened 2023-05-10
improve permission check

### LINKED ISSUE #3564: Check permission in JuiceFS
Currently we enable `default_permission` option and let kernel do the access check. With permission squash(#3562), we should do the check in JuiceFS process as well.

Hexilee:
close #3564

- [x] pass all pjdfstest except for fifo open
- [x] almost zero cost when `root-squash` and `non-default-permission` are disabled


cmd/mount_unix.go
```diff
@@ -133,7 +133,11 @@ func mount_flags() []cli.Flag {
 		},
 		&cli.StringFlag{
 			Name:  "root-squash",
-			Usage: "map local root user (uid = 0) to another one in filesystem",
+			Usage: "mapping local root user (uid = 0) to another one specified as <uid>:<gid>",
+		},
+		&cli.BoolFlag{
+			Name:  "non-default-permission",
+			Usage: "disable `default_permissions` option, only for testing",
```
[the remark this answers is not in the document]
Hexilee: Done

pkg/fuse/gidcache.go
```diff
@@ -0,0 +1,115 @@
+/*
+ * JuiceFS, Copyright 2020 Juicedata, Inc.
```
davies, line 2: 2023

pkg/vfs/vfs_unix.go
```diff
@@ -179,13 +182,15 @@ func (v *VFS) SetAttr(ctx Context, ino Ino, set int, opened uint8, mode, uid, gi
 	if set&meta.SetAttrMtime != 0 {
 		attr.Mtime = mtime
 		attr.Mtimensec = mtimensec
-		v.writer.UpdateMtime(ino, time.Unix(mtime, int64(mtimensec)))
-	}
-	if set&meta.SetAttrMtimeNow != 0 {
-		v.writer.UpdateMtime(ino, time.Now())
 	}
 	err = v.Meta.SetAttr(ctx, ino, uint16(set), 0, attr)
```
davies, line 186: When m.Write() is called between SetAttr() and UpdateMtime(), then mtime will be wrong.

pkg/fuse/gidcache.go
Hexilee: done

pkg/vfs/vfs_unix.go
Hexilee: should we always update `mtime` even if the `setattr` is denied?

pkg/fuse/context.go
```diff
@@ -58,7 +62,9 @@ func (fs *fileSystem) newContext(cancel <-chan struct{}, header *fuse.InHeader)
 	ctx.canceled = false
 	ctx.cancel = cancel
 	ctx.header = header
+	ctx.checkPermission = fs.conf.NonDefaultPermission
```
davies, line 65: When uid is 0, checkPermission should be false

pkg/vfs/vfs_unix.go
davies: We could check permission first, then call UpdateMtime before SetAttr

pkg/vfs/vfs_unix.go
```diff
@@ -43,6 +43,9 @@ type Statfs struct {
 }
 
 func (v *VFS) StatFS(ctx Context, ino Ino) (st *Statfs, err syscall.Errno) {
+	if st := v.Meta.Access(ctx, ino, MODE_MASK_R&MODE_MASK_X, nil); st != 0 {
```
davies, line 46: check ctx.CheckPermission()

pkg/vfs/vfs.go
```diff
@@ -448,17 +463,35 @@ func (v *VFS) Truncate(ctx Context, ino Ino, size int64, opened uint8, attr *Att
...
+		}
+		if h.writer == nil {
+			err = syscall.EACCES
+		}
 		if !h.Wlock(ctx) {
 			err = syscall.EINTR
 			return
 		}
 		defer func(h *handle) { h.Wunlock() }(h)
+		_ = h.writer.Flush(ctx)
```
davies, line 492: Why not flush all the fd?

davies: We could leave this one out of current PR

pkg/meta/tkv.go
```diff
@@ -813,6 +813,11 @@ func (m *kvMeta) deleteKeys(keys ...[]byte) error {
 }
 
 func (m *kvMeta) doLookup(ctx Context, parent Ino, name string, inode *Ino, attr *Attr) syscall.Errno {
+	if ctx.CheckPermission() {
```
davies, line 816: move it to base.Lookup?

pkg/meta/tkv.go
```diff
@@ -1561,15 +1624,15 @@ func (m *kvMeta) doRename(ctx Context, parentSrc Ino, nameSrc string, parentDst
...
+			if ctx.CheckPermission() && ctx.Uid() != 0 && dattr.Mode&01000 != 0 && ctx.Uid() != dattr.Uid && ctx.Uid() != tattr.Uid {
 				return syscall.EACCES
 			}
 		} else {
 			if exchange {
 				return syscall.ENOENT
 			}
 		}
-		if ctx.Uid() != 0 && sattr.Mode&01000 != 0 && ctx.Uid() != sattr.Uid && ctx.Uid() != iattr.Uid {
+		if ctx.CheckPermission() && ctx.Uid() != 0 && sattr.Mode&01000 != 0 && ctx.Uid() != sattr.Uid && ctx.Uid() != iattr.Uid {
```
davies, line 1635: We should always do the permission check here

pkg/fuse/context.go
Hexilee: done

pkg/vfs/vfs_unix.go
Hexilee: moved it into meta

pkg/vfs/vfs.go
Hexilee: fixed

pkg/meta/tkv.go
Hexilee: done with readdir, too

pkg/meta/tkv.go
Hexilee: done

pkg/meta/context.go
```diff
@@ -78,6 +79,10 @@ func (c *wrapContext) WithValue(k, v interface{}) {
 	c.Context = context.WithValue(c.Context, k, v)
 }
 
+func (c *wrapContext) CheckPermission() bool {
+	return true
```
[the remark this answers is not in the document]
Hexilee: done

pkg/vfs/vfs_unix.go
Hexilee: done

Hexilee: done

pkg/vfs/vfs_unix.go
```diff
@@ -179,6 +179,13 @@ func (v *VFS) SetAttr(ctx Context, ino Ino, set int, opened uint8, mode, uid, gi
 	if set&meta.SetAttrMtime != 0 {
 		attr.Mtime = mtime
 		attr.Mtimensec = mtimensec
+	}
+	if ctx.CheckPermission() {
+		if err = v.Meta.CheckSetAttr(ctx, ino, uint16(set), *attr); err != 0 {
```
davies, line 184: only for UpdateMtime

pkg/meta/base.go
```diff
@@ -2444,3 +2478,162 @@ LOOP:
...
+		}
+		if cur.Gid != attr.Gid {
+			if ctx.CheckPermission() && ctx.Uid() != 0 && !containsGid(ctx, attr.Gid) {
+				return syscall.EPERM
+			}
+		}
+	}
+	if set&SetAttrUID != 0 {
+		if cur.Uid != attr.Uid {
+			if ctx.CheckPermission() && ctx.Uid() != 0 {
```
davies, line 2596: even ctx.CheckPermission() is false, we should still check permission here (no regression)

pkg/meta/base.go
```diff
@@ -2444,3 +2478,162 @@ LOOP:
 	}
 	return eno
 }
+
+func (m *baseMeta) setAttr(ctx Context, inode Ino, set uint16, cur, attr *Attr, now time.Time) (*Attr, syscall.Errno) {
+	dirtyAttr := *cur
```
SandyXSD, line 2483: why not just change the `cur` here?

Hexilee: Because we need the fields (`uid` and `gid`) of `cur` to check permission, I think changing them in the mid may cause bugs. We can also rearrange the control flow, however, I think keep the `cur` immutable is a better solution.

pkg/vfs/vfs_unix.go
Hexilee: done

pkg/meta/base.go
```diff
@@ -2444,3 +2485,104 @@ LOOP:
...
+		if attr.Mode != cur.Mode {
+			if ctx.Uid() != 0 && ctx.Uid() != cur.Uid &&
+				(cur.Mode&01777 != attr.Mode&01777 || attr.Mode&02000 > cur.Mode&02000 || attr.Mode&04000 > cur.Mode&04000) {
+				return nil, syscall.EPERM
+			}
+			dirtyAttr.Mode = attr.Mode
+			changed = true
+		}
+	}
+	if set&SetAttrAtimeNow != 0 || (set&SetAttrAtime) != 0 && attr.Atime < 0 {
```
SandyXSD, line 2533: when would be `attr.Atime < 0`?

Hexilee: Actually, I don't know when it occurs in the real world, but I see the [setattrStr](https://github.com/juicedata/juicefs/blob/6f8c47cc2cf9a43453e4250abde2b1b85c6024af/pkg/vfs/vfs_unix.go#LL114C37-L114C37) and its unit tests assume it will.

PR #3584, opened 2023-05-10
meta/redis: support redis connection pool config to avoid connection …

cmmp6:
…explosion


cmmp6:
> 

y, you are right. I will close pr and request.

PR #3585, opened 2023-05-10
meta: move touchAtime to base meta

SandyXSD:


pkg/meta/redis.go
```diff
@@ -4282,52 +4280,23 @@ func (m *redisMeta) doAttachDirNode(ctx Context, parent Ino, dstIno Ino, name st
...
 			return nil
 		}
 		attr.Atime = now.Unix()
 		attr.Atimensec = uint32(now.Nanosecond())
-		_, err = tx.TxPipelined(ctx, func(pipe redis.Pipeliner) error {
-			pipe.Set(ctx, m.inodeKey(ino), m.marshal(attr), 0)
-			return nil
-		})
-		updated = true
+		if err = tx.Set(ctx, m.inodeKey(inode), m.marshal(attr), 0).Err(); err == nil {
```
davies, line 4296: we should use pipeline to end a transaction, if there is only one operation

PR #3590, opened 2023-05-11
CI: config codecov threshold

zhijian-pro:
threshold Allow the coverage to drop by X%, and posting a success status.
doc: https://docs.codecov.com/docs/commit-status#threshold


zhijian-pro:
Configure in organization Settings

PR #3591, opened 2023-05-11
meta: track atime for ReadLink as well

SandyXSD:


davies:
Please add unit test to cover these

PR #3657, opened 2023-05-22
cmd/restore: restore files in trash

davies:


cmd/recover.go
```diff
@@ -0,0 +1,123 @@
...
+			&cli.IntFlag{
+				Name:  "threads",
+				Value: 10,
+				Usage: "number of threads",
+			},
+		},
+	}
+}
+
+func recover(ctx *cli.Context) error {
```
Hexilee, line 41: this function name conflicts with build-in function `recover() error`.

Hexilee: and the debug command try to use the build-in `recover` https://github.com/juicedata/juicefs/blob/main/cmd/debug.go#L242

Hexilee (approved):
Rest LGTM

PR #3658, opened 2023-05-22
cmd/debug: set default log limit 

zhijian-pro:
set default log limit and return error when find more than one mount process


cmd/debug.go
```diff
@@ -424,6 +414,7 @@ func collectPprof(ctx *cli.Context, cmd string, pid string, amp string, requireR
 		return fmt.Errorf("failed to get pprof port: %v", err)
 	}
 	baseUrl := fmt.Sprintf("http://localhost:%d/debug/pprof/", port)
+	logger.Infof("The pprof base url: %s", baseUrl)
```
davies, line 417: debug?

PR #3661, opened 2023-05-22
doc: Automating Mounting with systemd.mount

xyb:


docs/en/guide/mount_at_boot.md
```diff
@@ -27,6 +27,42 @@ sudo chkconfig --add netfs
 
 :::
 
+### Automating Mounting with systemd.mount
```
davies, line 30: Please add the Chineses version as well

xyb: Sure, missing Chinese version added now.

PR #3686, opened 2023-05-26
cmd/config: add option to enable/disable dir stats

Hexilee:


cmd/format.go
```diff
@@ -179,6 +179,10 @@ Details: https://juicefs.com/docs/community/quick_start_guide`,
 				Name:  "no-update",
 				Usage: "don't update existing volume",
 			},
+			&cli.BoolFlag{
+				Name:  "no-dir-stats",
```
davies, line 183: remove it from `format`, user should use `config` instead.

pkg/meta/base_test.go
```diff
@@ -1999,9 +1999,14 @@ func testCheckAndRepair(t *testing.T, m Meta) {
 }
 
 func testDirStat(t *testing.T, m Meta) {
+	format := m.GetFormat()
+	format.DirStats = true
```
davies, line 2003: it should be enabled in tests

pkg/meta/sql.go
```diff
@@ -322,8 +322,38 @@ func (m *dbMeta) Init(format *Format, force bool) error {
...
 			return fmt.Errorf("json: %s", err)
 		}
+		if old.DirStats && !format.DirStats {
+			// remove dir stats
+			_, err = m.db.Where("TRUE").Delete(new(dirStats))
+			if err != nil {
+				return errors.Wrap(err, "drop table dirStats")
+			}
+		}
+		if !old.DirStats && format.DirStats {
```
davies, line 332: this could be slow, and it should not block `Init()`.

The stats could be rebuild using `fsck`

pkg/meta/base.go
```diff
@@ -705,6 +723,9 @@ func (m *baseMeta) checkDirQuota(ctx Context, inode Ino, space, inodes int64) bo
 }
 
 func (m *baseMeta) updateDirQuota(ctx Context, inode Ino, space, inodes int64) {
+	if !m.GetFormat().DirStats {
```
davies, line 726: We should enable it when first quota was set.

cmd/config.go
```diff
@@ -289,6 +299,12 @@ func config(ctx *cli.Context) error {
 				return fmt.Errorf("Aborted.")
 			}
 		}
+		if originDirStats && !format.DirStats {
```
davies, line 302: we can't disable it if any quota is set

pkg/meta/base_test.go
Hexilee: The `Init` method receives `format` as final state, we call `Init` for many times in tests and the `DirStat` is false by default. So if we want to ensure `DirStat` is enabled we should set it in each time we call `Init`.

cmd/config.go
Hexilee: done

pkg/meta/sql.go
Hexilee: done

pkg/meta/base.go
Hexilee: done

pkg/meta/base.go
```diff
@@ -2029,6 +2068,9 @@ func (m *baseMeta) resolve(ctx Context, dpath string, inode *Ino) syscall.Errno
 }
 
 func (m *baseMeta) GetFormat() Format {
+	if m.fmt == nil {
+		return Format{}
```
davies, line 2072: panic?

pkg/meta/tkv.go
```diff
@@ -411,8 +411,38 @@ func (m *kvMeta) Init(format *Format, force bool) error {
 		if err != nil {
 			return fmt.Errorf("json: %s", err)
 		}
+		if old.DirStats && !format.DirStats {
+			qs, err := m.doLoadQuotas(Background)
```
davies, line 415: move this check into cmd/config.go

cmd/config.go
```diff
@@ -234,6 +238,11 @@ func config(ctx *cli.Context) error {
 				format.TrashDays = new
 				trash = true
 			}
+		case "dir-stats":
+			if new := ctx.Bool(flag); new != format.DirStats {
+				msg.WriteString(fmt.Sprintf("%10s: %t -> %t\n", flag, format.DirStats, new))
+				format.DirStats = new
```
davies, line 244: tell user what to do (calculate the stats for existing dirs) after dirStat is enabled

pkg/meta/base.go
```diff
@@ -768,6 +797,16 @@ func (m *baseMeta) HandleQuota(ctx Context, cmd uint8, dpath string, quotas map[
 
 	switch cmd {
 	case QuotaSet:
+		format, err := m.Load(false)
+		if err != nil {
+			return errors.Wrap(err, "load format")
+		}
+		if !format.DirStats {
+			format.DirStats = true
```
davies, line 805: tell user the dir stats is enabled, ask them to update the stats manually, or only for current dir with quota

cmd/config.go
Hexilee: achieve consensus, no need

pkg/meta/base.go
Hexilee: achieve consensus, no need

pkg/meta/base.go
Hexilee: reverted

pkg/meta/tkv.go
Hexilee: done

pkg/meta/base.go
Hexilee: now we will load format if nil

PR #3693, opened 2023-05-27
sync: list objects in parallel

davies:


pkg/object/file.go
```diff
@@ -318,7 +321,42 @@ func readDirSorted(dirname string) ([]*mEntry, error) {
...
+		}
+		return nil, err
+	}
+	var objs []Object
+	for _, e := range entries {
+		p := filepath.Join(dir, e.Name())
+		if e.IsDir() {
+			p = filepath.ToSlash(p + "/")
+		}
+		key := p[len(d.root):]
```
SandyXSD, line 344: it will panic if `len(d.root) > len(dir) + len(name)`.
e.g. List `/a/bb`, while there's another entry named `/a/b`

SandyXSD (approved):

PR #3695, opened 2023-05-29
cmd/clone: fix clone properties not being applied

zhijian-pro:


davies:
update the unit tests to cover it

PR #3701, opened 2023-05-29
meta/tkv: fix the issue that the same slice may be appended for multiple times

### LINKED ISSUE #3008: meta/redis: in deleteChunk do dedup of the slices to make sliceRefs right

**Why is this needed**:
It seems to be possible that sometimes a Redis transaction is executed multiple times, leaving several identical slices in one chunk. In this case, sliceRef is reduced more than actually need.

Confirmation required.

SandyXSD:
close #3008


davies:
We have not find a clue that the write transaction is repeated, but better to double check it.

PR #3707, opened 2023-05-29
fix quota staging usage loss

### LINKED ISSUE #3705: should not writable when quota reached.
**What happened**:
```

#!/bin/bash

set -e 
umount /jfs | true
rm test.db
rm /var/jfs/myjfs -rf 

./juicefs format sqlite3://test.db myjfs
./juicefs mount -d sqlite3://test.db /jfs --heartbeat 4
mkdir -p /jfs/d
./juicefs quota set sqlite3://test.db --path /d --capacity 1
sleep 5s
dd if=/dev/zero of=/jfs/d/test1 bs=1G count=1
sleep 5s
./juicefs quota get sqlite3://test.db --path /d
used=$(./juicefs quota get sqlite3://test.db --path /d 2>&1 | grep "/d" | awk -F'|' '{print $5}'  | tr -d '[:space:]')
[[ $used != "100%" ]] && echo "used should be 100%" && exit 1 || true
echo a | tee -a /jfs/d/test1 2>error.log && echo "echo should fail on out of space" && exit 1 || true
```

the last command   " echo a | tee -a /jfs/d/test1 "  should fail. 

**Environment**:
- Kernel (e.g. `uname -a`):
- Others:

Hexilee:
close #3705

### DIFF
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -731,8 +731,8 @@ func (m *baseMeta) flushQuotas() {
 		time.Sleep(time.Second * 3)
 		m.quotaMu.RLock()
 		for ino, q := range m.dirQuotas {
-			newSpace = atomic.SwapInt64(&q.newSpace, 0)
-			newInodes = atomic.SwapInt64(&q.newInodes, 0)
+			newSpace = atomic.LoadInt64(&q.newSpace)
+			newInodes = atomic.LoadInt64(&q.newInodes)
 			if newSpace != 0 || newInodes != 0 {
 				quotas[ino] = &Quota{newSpace: newSpace, newInodes: newInodes}
 			}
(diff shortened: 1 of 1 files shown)


davies (approved):

pkg/meta/base.go
```diff
@@ -741,11 +741,17 @@ func (m *baseMeta) flushQuotas() {
 
 		if err := m.en.doFlushQuotas(Background, quotas); err != nil {
 			logger.Warnf("Flush quotas: %s", err)
+		} else {
```
SandyXSD, line 744: so it's still possible to get wrong quota result? e.g. load usedSpace after flush, then add the `snap.newSpace` again to the `usedSpace`

davies: yes, the chance is lower than before

PR #3715, opened 2023-05-30
meta/sql: fix `doSyncDirStat` does not update existed dir stats

### LINKED ISSUE #3704: quota and dir stats repair can't work in SQL engine
**What happened**:

  ```
    set -e
    ./juicefs format sqlite3://test.db myjfs 
    ./juicefs mount -d sqlite3://test.db /jfs
    mkdir -p /jfs/d
    ./juicefs quota set sqlite3://test.db --path /d --capacity 1
    dd if=/dev/zero of=/jfs/d/test1 bs=1G count=1
    pid=$(ps -ef | grep "juicefs mount" | grep -v grep | awk '{print $2}')
    kill -9 $pid
    sleep 3s
    ./juicefs quota check sqlite3://test.db --path /d --strict && echo "quota check should fail" && exit 1 || true
    ./juicefs quota check sqlite3://test.db --path /d --repair
    ./juicefs quota check sqlite3://test.db --path /d --strict
```

The fast check of quota won't work event if we executed `juicefs fsck --sync-dir-stat --repair`.

**Environment**:
- Kernel (e.g. `uname -a`):
- Others:

### PR DESCRIPTION
**Pull request by Hexilee**
close #3704

Currently, the `doSyncDirStat` of SQL engine won't update existed dir stats, while other engines will.

### CODE REVIEW

File: pkg/meta/sql.go [hunk 1]
```diff
@@ -2479,10 +2479,15 @@ func (m *dbMeta) doSyncDirStat(ctx Context, ino Ino) (*dirStat, syscall.Errno) {
 		if !exist {
 			return syscall.ENOENT
 		}
-		_, err = s.Insert(&dirStats{ino, stat.length, stat.space, stat.inodes})
-		if err != nil && isDuplicateEntryErr(err) {
-			// other client synced
-			err = nil
+		exist, err = s.Exist(&dirStats{Inode: ino})
```
SandyXSD, line 2482: try `insert` first, and do update if it returns 'duplicated entry error'

pkg/meta/base.go
```diff
@@ -1999,17 +1999,30 @@ func (m *baseMeta) Check(ctx Context, fpath string, repair bool, recursive bool,
 					statBroken = true
 				}
 
+				if statAll {
+					s, st := m.calcDirStat(ctx, inode)
```
SandyXSD, line 2003: this slows the repair since we call the `calcDirStat` again within the `doSyncDirStat` func in L2017.

pkg/meta/sql.go
Hexilee: done

pkg/meta/base.go
Hexilee: fixed

SandyXSD (approved):

SandyXSD:
@Hexilee please add a unit test to cover this case.

Hexilee:
> @Hexilee please add a unit test to cover this case.

It may be difficult to cover this case in unit tests because I have no good idea to stop the bg process `doFlushDirStat`, we can only interrupt it by obtain `m.dirStatsLock`, which rely on type assertion in base_test.

@zhoucheng361 will cover this case in e2e tests.

pkg/meta/base.go
```diff
@@ -1999,17 +1999,30 @@ func (m *baseMeta) Check(ctx Context, fpath string, repair bool, recursive bool,
...
+					if stat.space != s.space || stat.inodes != s.inodes {
+						logger.Warnf("usage stat of %s should be %v, but got %v", path, s, stat)
+						statBroken = true
+					}
+				}
+
 				if repair {
 					if statBroken || statAll {
 						if _, st := m.en.doSyncDirStat(ctx, inode); st == 0 || st == syscall.ENOENT {
-							logger.Debugf("Stat of path %s (inode %d) is successfully synced", path, inode)
```
davies, line 2005: There will be too many logging

Hexilee: log level reverted

PR #3728, opened 2023-06-01
cmd: add `log-level` option to customize the log level

SandyXSD:
Sometimes (in tests) we want to set log level to `error` or even higher.


SandyXSD:
@davies Shall we hide this flag?

davies:
we can hide the flag for now.

Issue #3756, opened 2023-06-05
mysql: <FATAL>: Write 5041 beans in channel 0: Error 1390 (HY000): Prepared statement contains too many placeholders


zhoucheng361:
**What happened**:
https://github.com/juicedata/juicefs/actions/runs/5174034823/jobs/9319904723
```
runner@fv-az794-769:~/work/juicefs/juicefs$ ./juicefs load mysql://root:root@(127.0.0.1)/test  2M_emtpy_files.dump
2023/06/05 06:52:01.542308 juicefs[2953] <INFO>: Meta address: mysql://root:****@(127.0.0.1)/test [interface.go:494]
2023/06/05 06:52:01.930637 juicefs[2953] <INFO>: Loading from file ... [dump.go:332]
2023/06/05 06:52:02.104711 juicefs[2953] <FATAL>: Write 5041 beans in channel 0: Error 1390 (HY000): Prepared statement contains too many placeholders [sql.go:3761]
```
runner@fv-az794-769:~/work/juicefs/juicefs$ 

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**

**Environment**:
- JuiceFS version (use `juicefs --version`) or Hadoop Java SDK version:
- Cloud provider or hardware configuration running JuiceFS:
- OS (e.g `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Object storage (cloud provider and region, or self maintained):
- Metadata engine info (version, cloud provider managed or self maintained):
- Network connectivity (JuiceFS to metadata engine, JuiceFS to object storage):
- Others:

zhoucheng361:
Fixed by #3757

PR #3654, opened 2023-05-22
SQL engine: atime, mtine and ctime support nanoseconds

Hexilee:

PR #3757, opened 2023-06-05
load/sql: update max fields count to reduce max batch size

Hexilee:
in #3654 we introduce three new fields for the `node` table, so the numbers of placeholders may exceed the limitation of SQL engine like https://github.com/juicedata/juicefs/actions/runs/5049460379/jobs/9058941017


zhoucheng361:
fix #3756

zhijian-pro (approved):

PR #3768, opened 2023-06-06
sdk: skip permission check for root

davies:
When the path starts with `/`, it will check permission for root.

We assume root is accessible for everyone.

### DIFF
--- a/pkg/fs/fs.go
+++ b/pkg/fs/fs.go
@@ -701,7 +701,7 @@ func (fs *FileSystem) resolve(ctx meta.Context, p string, followLastSymlink bool
 			parent = inode
 			break
 		}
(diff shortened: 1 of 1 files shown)


pkg/fs/fs.go
```diff
@@ -701,7 +701,7 @@ func (fs *FileSystem) resolve(ctx meta.Context, p string, followLastSymlink bool
 			parent = inode
 			break
 		}
-		if i > 0 {
+		if parent > 1 {
```
SandyXSD, line 704: `i > 0` already skips the check for root directory?
What's the case (p = ?) are you trying to fix?

SandyXSD (approved):

PR #3783, opened 2023-06-08
hadoop: change back the so version

SandyXSD:
`soVer = 8` is used for v1.0.4.


sdk/java/src/main/java/io/juicefs/JuiceFileSystemImpl.java
```diff
@@ -581,7 +581,7 @@ public static Libjfs loadLibrary() throws IOException {
     LibraryLoader<Libjfs> libjfsLibraryLoader = LibraryLoader.create(Libjfs.class);
     libjfsLibraryLoader.failImmediately();
 
-    int soVer = 8;
+    int soVer = 7;
```
davies, line 584: 9?

SandyXSD: 9 is used for 1.1-beta1. As discussed with tommy, we'd like to always use 7 for the main branch, and +1 every time a new version is released.

tangyoupeng (approved):

PR #3823, opened 2023-06-19
nfs: support object store and full-featured sync

### LINKED ISSUE #3702: Support NFS as object store
Similar to sftp, we can use `sync` to work with NFS directly.

Hexilee:
close #3702


pkg/object/file_unix.go
```diff
@@ -36,6 +37,9 @@ func getOwnerGroup(info os.FileInfo) (string, string) {
 	case *sftp.FileStat:
 		owner = utils.UserName(int(st.UID))
 		group = utils.GroupName(int(st.GID))
+	case *nfs.Fattr:
```
davies, line 40: the build will fail with tag !nonfs

Hexilee: Fixed

pkg/sync/sync.go
```diff
@@ -556,11 +555,7 @@ func worker(tasks <-chan object.Object, src, dst object.ObjectStorage, config *C
 			}
 			var err error
 			if config.Links && obj.IsSymlink() {
-				if err = copyLink(src, dst, key); err == nil {
-					copied.Increment()
-					break
-				}
-				logger.Errorf("copy link failed: %s", err)
+				err = copyLink(src, dst, key)
```
davies, line 558: checkSum() doe not work with symlinks, we should skip it or fix it

Hexilee: skipped

Hexilee: Is there any idea to check the copied symlink?

PR #3835, opened 2023-06-20
objbench: fix ceph panic when put empty file

zhijian-pro:
image

### DIFF
--- a/pkg/object/ceph.go
+++ b/pkg/object/ceph.go
@@ -139,6 +139,9 @@ func (c *ceph) Put(key string, in io.Reader) error {
 		if b, ok := in.(*bytes.Reader); ok {
 			v := reflect.ValueOf(b)
 			data := v.Elem().Field(0).Bytes()
+			if len(data) == 0 {
+				return errors.New("ceph: can't put empty file")
+			}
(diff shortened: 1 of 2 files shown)


pkg/object/ceph.go
```diff
@@ -139,6 +139,9 @@ func (c *ceph) Put(key string, in io.Reader) error {
 		if b, ok := in.(*bytes.Reader); ok {
 			v := reflect.ValueOf(b)
 			data := v.Elem().Field(0).Bytes()
+			if len(data) == 0 {
+				return errors.New("ceph: can't put empty file")
+			}
 			return ctx.WriteFull(key, data)
 		}
 		buf := cephPool.Get().([]byte)
```
davies, line 147: `in` could also be empty string, we should return an error

zhijian-pro: `ctx.WriteFull` will panic when  `len(data)=0`

zhijian-pro: image

davies: I meant we should also fix the case that it's not `*bytes.Reader`

PR #3840, opened 2023-06-21
support GlusterFS as object store

davies:


SandyXSD:
@davies Please review this commit: f4732e9.
1. add back `Chtimes` and `Chown` to make it satisfy `FileSystem` interface, since it does behavior like an fs
2. adjust the `readDirSorted`, removing '.' and '..', cuz `os.File.ReadDir` removes them
3. minor adjust in the `file.go`, PTAL

PR #3855, opened 2023-06-27
object/ceph: fix ceph range get and get no-exist

### LINKED ISSUE #3836: object/ceph: repair ceph object storage
image

- [x] get non-exist
- [x] get partial object
- [x] put a big object
- [x] put empty object  https://github.com/juicedata/juicefs/pull/3835

zhijian-pro:
close #3836 
image

```
      // use ctx.WriteFull(key, data)

        for i := 85; i < 95; i++ {
		body := make([]byte, i<<20)
		rand.Read(body)
		err = s.Put("test", bytes.NewReader(body))
		if err != nil {
			t.Errorf("PUT %dM failed: %s", i, err.Error())
		} else {
			t.Logf("PUT %dM success", i)
		}
	}
```
image

### DIFF
--- a/pkg/object/ceph.go
+++ b/pkg/object/ceph.go
@@ -98,6 +98,9 @@ type cephReader struct {
 }
 
 func (r *cephReader) Read(buf []byte) (n int, err error) {
+	if r.limit == 0 {
+		return 0, io.EOF
+	}
 	if r.limit > 0 && int64(len(buf)) > r.limit {
 		buf = buf[:r.limit]
 	}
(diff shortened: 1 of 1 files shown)


pkg/object/ceph.go
```diff
@@ -121,6 +124,9 @@ func (r *cephReader) Close() error {
 }
 
 func (c *ceph) Get(key string, off, limit int64) (io.ReadCloser, error) {
+	if _, err := c.Head(key); err != nil {
```
davies, line 127: we can use it in the cephReader

PR #3857, opened 2023-06-28
hadoop: avoid create too many threads for multiple JuiceFileSystem instance

tangyoupeng:


sdk/java/src/main/java/io/juicefs/JuiceFileSystem.java
```diff
@@ -70,24 +68,23 @@ private synchronized static void patchDistCpChecksum() {
   public void initialize(URI uri, Configuration conf) throws IOException {
     super.initialize(uri, conf);
     fileChecksumEnabled = Boolean.parseBoolean(getConf(conf, "file.checksum", "false"));
-    startTrashEmptier(uri, conf);
+    if (Boolean.parseBoolean(getConf(conf, "enable-trash-emptier", "true"))) {
```
davies, line 71: disable-trash-emptier

davies: use `false` as default value

davies (approved):

PR #3874, opened 2023-07-04
refactor hdfs kerberos keytab load

tangyoupeng:
load decodedKeytab directly in memory to avoid possible multi clients conflicts

### DIFF
--- a/pkg/object/hdfs_kerberos.go
+++ b/pkg/object/hdfs_kerberos.go
@@ -35,10 +35,6 @@ func getKerberosClient() (*krb.Client, error) {
 	keytabBase64 := os.Getenv("KRB5KEYTAB_BASE64")
 	principal := os.Getenv("KRB5PRINCIPAL")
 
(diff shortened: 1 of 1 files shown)


pkg/object/hdfs_kerberos.go
```diff
@@ -35,30 +33,26 @@ func getKerberosClient() (*krb.Client, error) {
...
-	principal := os.Getenv("KRB5PRINCIPAL")
-	if keytabPath != "" && principal != "" {
-		kt, err := keytab.Load(keytabPath)
+	} else if keytabPath != "" {
+		kt, err = keytab.Load(keytabPath)
 		if err != nil {
 			return nil, err
 		}
+	}
+	if kt != nil && principal != "" {
```
davies, line 55: when KRB5PRINCIPAL is not set, we will ignore  KRB5KEYTAB or KRB5KEYTAB_BASE64, is that expected?

davies (approved):

PR #3888, opened 2023-07-06
dont try set maxmemory_policy if setting does not exist

p1u3o:
Apache KVRocks supports the needed Redis commands (see https://kvrocks.apache.org/docs/supported-commands) to be a backend for JuiceFS, however because it uses RocksDB, there is no max_memory policy, but JuiceFS will try and set this, resulting in an non-critical error at startup.

### DIFF
--- a/pkg/meta/redis.go
+++ b/pkg/meta/redis.go
@@ -3548,7 +3548,7 @@ func (m *redisMeta) checkServerConfig() {
 	if err != nil {
 		logger.Warnf("parse info: %s", err)
 	}
-	if rInfo.storageProvider == "" && rInfo.maxMemoryPolicy != "noeviction" {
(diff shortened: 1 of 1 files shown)


zhijian-pro:
Replacing https://github.com/juicedata/juicefs/blob/7d660d9ccf04e751a3a024f4926e0417c9bdb648/pkg/meta/redis.go#L3551
with
```
if rInfo.storageProvider == "" && rInfo.maxMemoryPolicy != "" && rInfo.maxMemoryPolicy != "noeviction" {
is logically simpler

p1u3o:
I agree, I'll change this MR.

p1u3o:
I have updated this MR with the suggested change and confirmed it works at silencing the error.

PR #3933, opened 2023-07-25
sync: read large file from sftp in parallel

davies:


zhijian-pro:
There is a 50% performance improvement
image

PR #3966, opened 2023-08-04
Configure parameters to regularly clear cache that has not been accessed for a long time.

feeyman:
We encountered a scenario where cached files are only used for a certain period of time and are no longer needed for a long time, such as three days. However, the files will remain in the cache until the cache reaches the quota before triggering deletion. We need to set a time, such as setting a file system for a certain format. 
### For example:  
If the cached files are not accessed for more than five days, they will be automatically deleted, Of course, it can also be set for a longer time.
juicefs format **--cached-stay-days 10** --storage s3  --bucket *****


davies:
This is kind of cache expiring policy, should be a parameter of mount point (not part of format). The cached blocks could be evicted once  it's not accessed after a period of time (--cache-expire xxd)

feeyman:
> This is kind of cache expiring policy, should be a parameter of mount point (not part of format). The cached blocks could be evicted once it's not accessed after a period of time (--cache-expire xxd)

Okay, I just submitted the code and set the '--cache-expire' parameter during the mount operation.

cmd/config.go
```diff
@@ -219,6 +219,14 @@ func config(ctx *cli.Context) error {
 				format.TrashDays = new
 				trash = true
 			}
+		case "cached-stay-days":
```
davies, line 222: We don't need to global config for this.

cmd/flags.go
```diff
@@ -323,10 +331,56 @@ func expandFlags(compoundFlags ...[]cli.Flag) []cli.Flag {
 	return flags
 }
 
+func parseDuration(input string) (time.Duration, error) {
```
davies, line 334: Should we use time.ParseDuration() here?

SandyXSD:
Taken over by #4255

PR #3967, opened 2023-08-04
write syslog in background

davies:
Syslog could be slow or unavailable sometimes, it should not block logging.


SandyXSD:
Somehow the SDKTest always failed.

SandyXSD:
The test fails on the main branch as well.

PR #3990, opened 2023-08-11
Makefile: use commit time for revision date

eryugey:
Author time might be an old timestamp, commit time is always monotonically increasing and useful to identify a revision.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ export GO111MODULE=on
 all: juicefs
 
 REVISION := $(shell git rev-parse --short HEAD 2>/dev/null)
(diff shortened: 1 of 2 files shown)


SandyXSD:
I don't see the necessity of this change. In what scenario do you think that it makes a notable difference?

davies:
Commit date and author date will be different when cherry-pick a commit from main to a release branch, specially when the author date of commit is older than last commit of a release branch, that will be confusing that a newer release has a date which could be older than a previous release.

+1 for the change, can you update all the '%ad' to '%cd'?

davies:
@eryugey Thanks!

PR #3993, opened 2023-08-14
cmd/gateway&webdav: gateways and webdav support daemon running

### LINKED ISSUE #3865: background option for the juicefs gateway command

when start S3 gateway service, the command does not have option background like juicefs mount. then if the gateway starting command terminal is terminated, the gateway service will stop.

I suggest we add an option background like juicefs mount 

**Why is this needed**:

to make the gateway service running in background and avoid interrupted by terminating session.

zhijian-pro:
close #3865

### DIFF
--- a/cmd/webdav.go
+++ b/cmd/webdav.go
@@ -22,6 +22,7 @@ package cmd
 
 import (
 	"os"
+	"path"
 
 	"github.com/juicedata/juicefs/pkg/fs"
 	"github.com/urfave/cli/v2"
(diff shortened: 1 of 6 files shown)


cmd/webdav.go
```diff
@@ -45,10 +46,20 @@ func cmdWebDav() *cli.Command {
...
+			Usage: "path for WebDAV log",
+			Value: path.Join(getDefaultLogDir(), "juicefs-webdav.log"), //nolint:typecheck
+		},
 		&cli.StringFlag{
 			Name:  "access-log",
 			Usage: "path for JuiceFS access log",
 		},
+		&cli.BoolFlag{
+			Name:    "d",
+			Aliases: []string{"background"},
```
davies, line 60: use the short one as alias

PR #4000, opened 2023-08-15
use crypto/rand.Read, and remove math/rand.Seed

SandyXSD:
Lint error for Go 1.20:
```
Error: SA1019: rand.Seed has been deprecated since Go 1.20 and an alternative has been available since Go 1.0: Programs that call Seed and then expect a specific sequence of results from the global random source (using functions such as Int) can be broken when a dependency changes how much it consumes from the global random source. To avoid such breakages, programs that need a specific result sequence should use NewRand(NewSource(seed)) to obtain a random generator that other packages cannot access.  (staticcheck)
Error: SA1019: rand.Read has been deprecated since Go 1.20 because it shouldn't be used: For almost all use cases, crypto/rand.Read is more appropriate.  (staticcheck)
```

### DIFF
--- a/pkg/object/restful.go
+++ b/pkg/object/restful.go
@@ -38,7 +38,6 @@ var resolver = dnscache.New(time.Minute)
 var httpClient *http.Client
 
 func init() {
-	rand.Seed(time.Now().Unix())
 	httpClient = &http.Client{
 		Transport: &http.Transport{
 			Proxy:                 http.ProxyFromEnvironment,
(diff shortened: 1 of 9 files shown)


Hexilee (approved):

pkg/object/restful.go
```diff
@@ -61,7 +60,7 @@ func init() {
 				}
 				var conn net.Conn
 				n := len(ips)
-				first := rand.Intn(n)
+				first := rand.New(rand.NewSource(time.Now().Unix())).Intn(n)
```
davies, line 63: rand.NewSource could be slow

PR #4001, opened 2023-08-15
update product comparison naviagtor, dont duplicate content from blog

Suave:
don't duplicate content from blog, link to it directly.


davies:
It's OK to have duplicated content between docs and blogs, because docs should be updated regularly, but blog will not.

PR #4002, opened 2023-08-16
Time duration can identify the number of days

feeyman:
In some scenarios, the time parameter supports days, which is better. For example, data is rarely modified and the backup interval is expected to be one week. Although 168h can also be used, 7d is more intuitive, such as --backup-meta 7d.

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -19,8 +19,10 @@ package cmd
 import (
 	"os"
 	"path"
+	"regexp"
 	"runtime"
 	"strconv"
+	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


cmd/flags.go
```diff
@@ -323,7 +325,51 @@ func expandFlags(compoundFlags ...[]cli.Flag) []cli.Flag {
 	return flags
 }
 
+func parseDuration(input string) (time.Duration, error) {
```
davies, line 328: We can use time.ParseDuration to cover most of the cases, and handle 7d here.

davies:
took over by #4246

PR #4018, opened 2023-08-23
cmd: fix options parsing of `mount.juicefs`

Hexilee:


cmd/main.go
```diff
@@ -143,14 +149,17 @@ func handleSysMountArgs(args []string) []string {
...
+				} else if _, ok := cmdFlagsLookup[fields[0]]; ok {
 					newArgs = append(newArgs, fmt.Sprintf("--%s=%s", fields[0], fields[1]))
 				} else {
 					fuseOptions = append(fuseOptions, opt)
 				}
 			} else if flagName, ok := optionToCmdFlag[opt]; ok {
 				newArgs = append(newArgs, fmt.Sprintf("--%s", flagName))
-			} else if isBool, ok := cmdFlagsLookup[opt]; ok && isBool {
+			} else if isBool, ok := cmdFlagsLookup[opt]; ok {
+				if !isBool {
```
SandyXSD, line 160: no need for this, just pass it to the fuseOptions

cmd/main.go
```diff
@@ -143,14 +149,17 @@ func handleSysMountArgs(args []string) []string {
 				fields := strings.SplitN(opt, "=", 2)
 				if flagName, ok := optionToCmdFlag[fields[0]]; ok {
 					newArgs = append(newArgs, fmt.Sprintf("--%s=%s", flagName, fields[1]))
-				} else if isBool, ok := cmdFlagsLookup[fields[0]]; ok && !isBool {
```
davies, line 146: we don't want to support `--debug=true` or `--debug=1`, it's better return an error for this case

PR #4040, opened 2023-09-05
utils/logger: support append user specified ID in log

eryugey:
So that we could filter logs from the same juicefs instance. And use "--log-id random" to generate random uuid as log id.

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -47,6 +47,10 @@ func globalFlags() []cli.Flag {
 			Usage:  "set log level (trace, debug, info, warn, error, fatal, panic)",
 			Hidden: true,
 		},
+		&cli.StringFlag{
+			Name:  "log-id",
+			Usage: "append the given log id in log, use \"random\" to use random uuid",
+		},
 		&cli.BoolFlag{
 			Name:  "no-agent",
 			Usage: "disable pprof (:6060) agent",
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -27,6 +27,7 @@ import (
 	"syscall"
 
 	"github.com/erikdubbelboer/gspt"
+	"github.com/google/uuid"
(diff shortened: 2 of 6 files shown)


SandyXSD:
Usually we use `pid` to separate different clients. When do you think is that not enough?

eryugey:
> 

When running in container, pid is always 1, so there's no way to distinguish different juicefs instances from pid. And logs from multiple juicefs containers might be collected to a central place, e.g. aliyun SLS, so log contents are merged together, we need log id to tell which log comes from which instance.

davies:
Can we added something to identify different containers when collecting the logging?

eryugey:
> 

That should work too, but I'm not aware any standard way to do this (I think that depends on log services, and probably not all log services provide the required function). So adding log id in juicefs works everywhere.

davies:
The pid (1) is useless in container, so we can change it to use session id in the logging.

eryugey:
In my case Session ID is not that useful either, I have multiple containers in a Pod, e.g. 2 sidecar containers and 1 main container, I want to use a unique log id for the 3 containers, so I could know the logs with this log id came from the same pod.

Session ID could only identify juicefs, but I still didn't know which pod it was in. So a user specified log id is better. Maybe session ID could be the default log id, not a random uuid.

davies:
How do you use `--log-id` in your case?

eryugey:
> How do you use `--log-id` in your case?

I use pod name as log id, e.g.

```yaml
env:
      - name: POD_NAME
        valueFrom:
          fieldRef:
            fieldPath: metadata.name
      - name: POD_NS
        valueFrom:
          fieldRef:
            fieldPath: metadata.namespace
...
   args: juicefs mount ... --log-id "${POD_NS}-${POD_NAME}"
```

davies:
Sounds reasonable, merging it, thanks!

PR #4069, opened 2023-09-26
meta/mysql: dont escape for mysql password

zhijian-pro:


cmd/format.go
```diff
@@ -70,6 +70,16 @@ $ juicefs format mysql://jfs:mypassword@(127.0.0.1:3306)/juicefs myjfs
 # A safer alternative
 $ META_PASSWORD=mypassword juicefs format mysql://jfs:@(127.0.0.1:3306)/juicefs myjfs
 
+# MySQL password containing special characters
+$ juicefs format 'mysql://jfs:123@45@(127.0.0.1:3306)/juicefs' myjfs
```
davies, line 74: don't add special case here, it will be too long

davies: update the docs

SandyXSD (approved):

PR #4071, opened 2023-09-26
fix password leakage problem when password contains `@ `symbol

zhijian-pro:


pkg/utils/utils.go
```diff
@@ -119,7 +119,7 @@ func WithTimeout(f func() error, timeout time.Duration) error {
 }
 
 func RemovePassword(uri string) string {
-	p := strings.Index(uri, "@")
+	p := strings.LastIndex(uri, "@")
```
davies, line 122: add a test case

SandyXSD (approved):

PR #4076, opened 2023-10-01
pkg/sync: quickly sync data between different buckets within the same cloud provider and region

zxh326:
This PR uses the `CopyObject/UploadPartCopy` API to sync data when the source and destination are from the same cloud vendor and region.

- file size < 100MB, use the copyObject API to directly copy whole object 
- file size >= 100MB, use the UploadPartCopy API to copy the file in parts, with each part being 100MB.

The `100MB` size was obtained from my local test, and the effect is good. There may be better values, but it should not be too large. cause the API copy request is very slow. Sometimes it even times out and fails.

Below are some test results.

> - with config `Threads: 10, ListThreads: 1, ListDepth: 10, Limit: -1`
> - `\ds` mean sync elapsed time
> - `X` mean unneeded tests
> - `TIMEOUT` mean sync api timeout

[test script](https://gist.github.com/zxh326/3b172b45a4cdf6bbad6004e5e7e6cb6a)

1. different bucket

| target\action  | CopyObject | 5MB part copy | 100MB part copy | 500MB part copy | 1GB part copy | old(internal) |
| -------------- | ---------- | ------------- | --------------- | --------------- | ------------- | ------------- |
| 10KB-file*1000 | 10.81s     | 25.07s        | X               | X               | X             | **2.67s**     |
| 100MB-file     | 3.01s      | **1.02s**     | 2.85s       | X               | X             | 1.13s         |
| 500MB-file     | 17.34s     | 3.87s         | **3.14s**       | 12.09s          | X             | 3.46s         |
| 1GB-file       | TIMEOUT    | 6.28s         | **4.72s**       | 21.21s          | 24.50s        | 5.17s         |
| 5GB-file       | TIMEOUT    | 27s           | **19s**         | 20.74s          | TIMEOUT       | 24.1s         |

2. same bucket

| targetaction  | CopyObject | 5MB part copy | 100MB part copy | 500MB part copy | 1GB part copy |
| -------------- | ---------- | ------------- | --------------- | --------------- | ------------- |
| 10KB-file*1000 | **7.99s**     | 10.59s          | X               | X               | X             |
| 100MB-file     | **0.18s**      | 2.03s         | 0.74s           | X               | X             |
| 500MB-file     | **0.13s**      | 2.87s         | 2.50s           | 0.14s           | X             |
| 1GB-file       | **0.17s**      | 6.67s         | 3.92s           | 21.21s          | 0.21s         |
| 5GB-file       | **0.12s**      | 25s           | 21s             | 20.74s          | TIMEOUT       |

So i think in different bucket use 100MB as default part size via partCopy api, if less or in same bucket via copyObject api.

In addition, when it comes to copying a large number of small files in an intranet, using the original method(download & upload) is faster. Perhaps further optimization can be considered.


davies:
Why not copy the whole object?

zxh326:
> Why not copy the whole object?

I didn't notice, sorry. 

Some cloud vendors have size limit for copying objects, so the copyObject API cannot be used directly. For example, OSS has a size limit of 1GB and S3 limit of 1GB.

So this is what I plan to do now:

- If the size < limit, use the copyObject API to directly copy the entire object.
- If the size > limit, use the uploadPartCopy API API to only copy one part of the entire object.

or,  only using uploadPartCopy API, but may be very slow  for copying a large number of small files because this api requires 3 request operations(init, put, complete..).

What do you think?

davies:
We should use Copy or CopyPart based on size of objects and limits.

zxh326:
> Why not copy the whole object?

I did some testing (on oss and s3), for large files (around 1GB), the API copy request is very slow. Sometimes it even times out and fails. I think it would be better to use part copying for large file, possibly 100MB per part, not based on cloud limits

zxh326:
ping

PTAL @davies

pkg/object/azure.go
```diff
@@ -106,7 +106,7 @@ func (b *wasb) Put(key string, data io.Reader) error {
 	return err
 }
 
-func (b *wasb) Copy(dst, src string) error {
+func (b *wasb) Copy(dst, src, srcBucket string) error {
```
davies, line 109: It will be wrong if the `srcBucket` is different, we could add another method to support, otherwise we should return  error if it's not supported.

pkg/sync/sync.go
```diff
@@ -576,7 +615,11 @@ func worker(tasks <-chan object.Object, src, dst object.ObjectStorage, config *C
 					logger.Errorf("copy link failed: %s", err)
 				}
 			} else {
-				err = copyData(src, dst, key, obj.Size())
+				serviceCopy := false
+				if !config.DisableServiceCopy {
+					serviceCopy = checkUseServiceCopy(src, dst)
```
SandyXSD, line 620: Is it necessary to do the check for every object?

pkg/sync/sync.go
```diff
@@ -604,6 +647,13 @@ func worker(tasks <-chan object.Object, src, dst object.ObjectStorage, config *C
 	}
 }
 
+func checkUseServiceCopy(src, dst object.ObjectStorage) bool {
+	return reflect.TypeOf(src) == reflect.TypeOf(dst) &&
```
SandyXSD, line 651: Reflection is usually slow, and the check might be false positive if the object storages are wrapped by `withPrefix`.

cmd/sync.go
```diff
@@ -174,6 +174,10 @@ func syncActionFlags() []cli.Flag {
 			Name:  "dry",
 			Usage: "don't copy file",
 		},
+		&cli.BoolFlag{
+			Name:  "disable-service-copy",
```
SandyXSD, line 178: use 'server' instead of 'service'?

PR #4083, opened 2023-10-08
try to support OFD lock

davies:


pkg/vfs/vfs.go
```diff
@@ -522,14 +522,18 @@ func (v *VFS) Release(ctx Context, ino Ino, fh uint64) {
 				}
 			}
 			locks := f.locks
-			owner := f.flockOwner
+			fowner := f.flockOwner
+			powner := f.ofdOwner
```
SandyXSD, line 526: record the owners in the access log?

davies: The owner come from previous setlk, we don't need to put it in accesslog.

PR #4093, opened 2023-10-12
cmd/mount: `mount.juicefs` support `-d`

zhijian-pro:
`mount.juicefs` supports both foreground and background running without the `JFS_FOREGROUND` variables


SandyXSD:
Why we need this?

PR #4097, opened 2023-10-13
fix sync: only compare ModePerm

Hexilee:
Some object storage does not support other bits (like the sticky bit) in file mode.


Hexilee:
closed

PR #4104, opened 2023-10-17
cmd/sync: supports in-place put to avoid calling rename

zhijian-pro:


pkg/object/hdfs.go
```diff
@@ -121,29 +122,29 @@ func (h *hdfsclient) Get(key string, off, limit int64) (io.ReadCloser, error) {
...
+	if !PutInplace {
+		defer func() { _ = h.c.Remove(tmp) }()
+	}
 	if err != nil {
 		if pe, ok := err.(*os.PathError); ok && pe.Err == os.ErrNotExist {
 			_ = h.c.MkdirAll(filepath.Dir(path), 0777&^h.umask)
 			f, err = h.c.CreateFile(tmp, h.dfsReplication, 128<<20, 0666&^h.umask)
 		}
-		if pe, ok := err.(*os.PathError); ok {
-			if remoteErr, ok := pe.Err.(hdfs.Error); ok && remoteErr.Exception() == abcException {
```
davies, line 140: ？

cmd/object.go
```diff
@@ -133,13 +138,15 @@ func (j *juiceFS) Put(key string, in io.Reader) (err error) {
 	if eno != 0 {
```
davies, line 138: eno could be syscall.EEXISTS

PR #4138, opened 2023-11-06
fix path in Windows

zhijian-pro:
image


zhijian-pro:
no problem with the test

PR #4144, opened 2023-11-07
cmd/mount: add `all-squash` option to squash all users

SandyXSD:


cmd/mount_unix.go
```diff
@@ -141,6 +141,10 @@ func fuseFlags() []cli.Flag {
 			Name:  "root-squash",
 			Usage: "mapping local root user (uid = 0) to another one specified as <uid>:<gid>",
 		},
+		&cli.StringFlag{
```
Hexilee, line 144: We should make flag `non-default-permission` visible or enable it when squash all users. Otherwise, the kernel may deny users to modify files they created.

```
$ touch hello
touch: setting times of 'hello': Permission denied
```

SandyXSD: We probably have to always disable default permission to make the `all-squash` work

davies: No, there will be lots of access() and much slower.

Hexilee (approved):

cmd/mount_unix.go
Hexilee: But `all-squash` can hardly work with `default-permission`.

pkg/vfs/vfs.go
```diff
@@ -62,13 +62,14 @@ type Config struct {
 	AccessLog            string `json:",omitempty"`
 	PrefixInternal       bool
 	HideInternal         bool
-	RootSquash           *RootSquash `json:",omitempty"`
+	UserSquash           *UserSquash `json:",omitempty"`
```
davies, line 65: Will it break anything (loading a JSON)?

cmd/mount_unix.go
davies: When defaultPermission is disabled, we should make sure that the permission is checked in all the places, or access is called for them.

pkg/vfs/vfs.go
SandyXSD: Usually it's dynamically loaded

PR #4158, opened 2023-11-10
quota: atomically set limitation

Hexilee:


pkg/meta/redis.go
```diff
@@ -3471,19 +3471,18 @@ func (m *redisMeta) doSetQuota(ctx Context, inode Ino, quota *Quota) (bool, erro
...
+		} else if e == redis.Nil {
+			created = true
+		} else {
 			return e
 		}
 		if created && quota.MaxSpace < 0 && quota.MaxInodes < 0 {
-			// limit is deleted, skip
-			created = false
-			return nil
+			return errors.Errorf("limitation not set or deleted")
```
davies, line 3485: move this into the case for e is redis.Nil

PR #4172, opened 2023-11-16
CI: add random tests for meta

davies:


pkg/meta/base_test.go
```diff
@@ -542,7 +543,7 @@ func testMetaClient(t *testing.T, m Meta) {
 	if st := m.SetXattr(ctx, inode, "a", []byte("v4"), XattrReplace); st != 0 {
 		t.Fatalf("setxattr: %s", st)
 	}
-	if st := m.SetXattr(ctx, inode, "a", []byte("v5"), 5); st != 0 { // unknown flag is ignored
+	if st := m.SetXattr(ctx, inode, "a", []byte("v5"), 5); st != syscall.EINVAL { // unknown flag is ignored
```
SandyXSD, line 546: remove the comment

pkg/meta/tkv.go
```diff
@@ -1492,6 +1492,10 @@ func (m *kvMeta) doRename(ctx Context, parentSrc Ino, nameSrc string, parentDst
 			}
 			return nil
 		}
+		// todo: check dst parent is subdir of parent
+		if ino == parentDst {
```
SandyXSD, line 1496: check it as well for the other two engines?

PR #4179, opened 2023-11-20
convert current write lock into read lock

davies:
flock should convert current exclusive lock into a shared lock.


SandyXSD (approved):

SandyXSD:
https://man7.org/linux/man-pages/man2/flock.2.html

> A process may hold only one type of lock (shared or exclusive) on
       a file.  Subsequent flock() calls on an already locked file will
       convert an existing lock to the new lock mode.

Issue #4178, opened 2023-11-20
refreshUidAndGrouping function always throws an error


zhijian-pro:
### Discussed in https://github.com/juicedata/juicefs/discussions/4175

<div type='discussions-op-text'>

<sup>Originally posted by **XLPE** November 17, 2023</sup>
The hadoop core-site.xml configuration is as follows:
```    
<property>
        <name>juicefs.users</name>
        <value>jfs://bigfs/etc/users</value>
    </property>
    <property>
        <name>juicefs.groups</name>
        <value>jfs://bigfs/etc/groups</value>
</property>
```

The content of users is as follows:
`
hdfs:1110
`
The content of groups is as follows:
`
supergroup:1:hdfs
`

After starting hive server2, I always see the following error every 1 minute interval:

```2023-11-17T20:59:58,696  WARN [Background Task] juicefs.JuiceFileSystemImpl: read jfs://bigfs/etc/users failed
org.apache.hadoop.fs.InvalidRequestException: Invalid parameter
	at io.juicefs.JuiceFileSystemImpl.error(JuiceFileSystemImpl.java:267) ~[juicefs-hadoop-1.1.0.jar:?]
	at io.juicefs.JuiceFileSystemImpl.getFileStatusInternal(JuiceFileSystemImpl.java:1543) ~[juicefs-hadoop-1.1.0.jar:?]
	at io.juicefs.JuiceFileSystemImpl.getFileStatus(JuiceFileSystemImpl.java:1527) ~[juicefs-hadoop-1.1.0.jar:?]
	at io.juicefs.JuiceFileSystemImpl.readFile(JuiceFileSystemImpl.java:479) ~[juicefs-hadoop-1.1.0.jar:?]
	at io.juicefs.JuiceFileSystemImpl.updateUidAndGrouping(JuiceFileSystemImpl.java:498) ~[juicefs-hadoop-1.1.0.jar:?]
	at io.juicefs.JuiceFileSystemImpl.lambda$refreshUidAndGrouping$0(JuiceFileSystemImpl.java:510) ~[juicefs-hadoop-1.1.0.jar:?]
	at io.juicefs.utils.BgTaskUtil.lambda$startScheduleTask$1(BgTaskUtil.java:53) ~[juicefs-hadoop-1.1.0.jar:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [?:1.8.0_311]
[...]

PR #4180, opened 2023-11-20
Hadoop: fix guid refresh by avoid use closed fs

tangyoupeng:


tangyoupeng:
#4178

sdk/java/src/main/java/io/juicefs/JuiceFileSystemImpl.java
```diff
@@ -472,7 +472,7 @@ private String readFile(String file) {
               (name != null && name.equals(uri.getAuthority()))) {
         fs = this;
       } else {
-        fs = path.getFileSystem(getConf());
```
davies, line 475: Is it too slow to create a filesystem to open single file?

If the fs is closed, can we capture the exception and try again?

PR #4188, opened 2023-11-23
meta: increase inodeBatch & sliceIdBatch

SandyXSD:


davies:
This patch can reduce the number transaction needed and reduce the conflict.

PR #4189, opened 2023-11-23
cmd/mount: add option `skip-dir-mtime` to eliminate burst mtime updates of a directory

SandyXSD:


davies:
The argument should be explicit about updating mtime of parent dir.

PR #4197, opened 2023-11-24
same owner should be able to convert an EXCLUSIVE lock to a SHARED lock by requesting the SHARED lock (redis meta, flock, posix)

dbotwinick:
I moved the call to delete lkey from owners in meta/redis_lock.go so that flock would support converting an EXCLUSIVE lock to a SHARED lock for the same owner. As originally implemented, if there is any WRITE lock, then the lock operation should return EAGAIN which is problematic for calls that are not explicitly non-blocking intended to convert an EXCLUSIVE lock to a SHARED lock. I've been coming across blocking lock "downgrade" scenarios for database software. 

TODO: 
- Does this need to be applied to other meta implementations?
- Does this change have any other implications?
- Does this change only make sense in the ofd_lock branch?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,10 @@ juicefs: Makefile cmd/*.go pkg/*/*.go go.*
 	go version
 	go build -ldflags="$(LDFLAGS)"  -o juicefs .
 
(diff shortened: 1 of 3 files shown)


davies:
This is already fixed in main branch by #4179

PR #4218, opened 2023-11-30
object/ceph: Speedup list for ceph

davies:
This PR will call stat from in parallel to speedup List.

### DIFF
--- a/pkg/object/ceph.go
+++ b/pkg/object/ceph.go
@@ -230,24 +230,76 @@ func (c *ceph) ListAll(prefix, marker string, followLink bool) (<-chan Object, e
 	}
 	// the keys are not ordered, sort them first
 	sort.Strings(keys)
+	c.release(ctx)
 
 	var objs = make(chan Object, 1000)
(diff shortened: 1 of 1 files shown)


SandyXSD:
Time cost for listing 800k objects reduced from 128s to 8.7s.

SandyXSD (approved):

pkg/object/ceph.go
```diff
@@ -230,24 +230,76 @@ func (c *ceph) ListAll(prefix, marker string, followLink bool) (<-chan Object, e
...
+		conds[j] = sync.NewCond(&ms[j])
+		if j < len(keys) {
+			go func(j int) {
+				ctx, err := c.newContext()
+				if err != nil {
+					logger.Errorf("new context: %s", err)
+					errs[j] = err
+					return
+				}
+				defer ctx.Destroy()
```
SandyXSD, line 252: use `c.release`?

davies: we release the context when there is on error

PR #4221, opened 2023-11-30
docs: update wording for juicefs_vs_s3fs.md

CaitinChen:


docs/en/introduction/comparison/juicefs_vs_s3fs.md
```diff
@@ -1,30 +1,31 @@
...
 
-JuiceFS uses a technical architecture that separates data and metadata, where any file is first split into data blocks according to specific rules before being uploaded to the object storage, and the corresponding metadata is stored in a separated database. The advantage of this is that retrieval of files and modification of metadata such as file names can directly interact with the database with a faster response, bypassing the network latency impact of interacting with the object store.
+JuiceFS: Uses a unique architecture that separates data and metadata. Files are split into data blocks before being uploaded to object storage, and metadata is stored in a separate database. This architecture allows faster retrieval and metadata modification, reducing the impact of network latency.
```
CaitinChen, line 22:
```suggestion
JuiceFS uses a architecture that separates data and metadata. Files are split into data blocks according to specific rules before being uploaded to object storage, and the corresponding metadata is stored in a separate database. The advantage of this is that retrieval of files and modification of metadata such as file names can directly interact with the database with a faster response, bypassing the network latency impact of interacting with the object store.
```

PR #4253, opened 2023-12-12
meta: slowdown dump process to reduce memory usage

### LINKED ISSUE #4252: High memory usage (almost OOM) on dump with Badger
_From https://juicefs.slack.com/archives/C01HXAD7LDQ/p1702284763003769_

**What happened**:

When performing juicefs dump using Badger, the overall memory consumption is very high.

Not consuming almost 80% of both RAM and SWAP on a low-end box. (4Gb RAM + 3Gb SWAP).
INodes used are 2.6M. Badger dir is 300M and produced dump file is 1Gb.

Prefer slower dump but lower memory usage.

`juicefs dump badger://"${BADGER}" "${DUMP_FILE}" --keep-secret-key`

**Environment**:
- JuiceFS version : `juicefs version 1.1.1+2023-11-28.437f4e6`
- OS:  `Ubuntu 22.04.3 LTS`
- Kernel: `5.15.0-91-generic`
- Metadata engine: `Badger`

davies:
close #4252

### DIFF
--- a/pkg/meta/sql.go
+++ b/pkg/meta/sql.go
@@ -3524,7 +3524,7 @@ func (m *dbMeta) makeSnap(ses *xorm.Session, bar *utils.Bar) error {
 	return nil
 }
 
-func (m *dbMeta) DumpMeta(w io.Writer, root Ino, keepSecret bool) (err error) {
+func (m *dbMeta) DumpMeta(w io.Writer, root Ino, keepSecret, fast bool) (err error) {
 	defer func() {
 		if p := recover(); p != nil {
 			if e, ok := p.(error); ok {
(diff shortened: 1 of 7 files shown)


pkg/meta/sql.go
```diff
@@ -3552,6 +3552,9 @@ func (m *dbMeta) DumpMeta(w io.Writer, root Ino, keepSecret bool) (err error) {
 			if tree, err = m.dumpEntry(s, root, TypeDirectory); err != nil {
 				return err
 			}
+			if trash, err = m.dumpEntry(s, TrashInode, TypeDirectory); err != nil {
```
SandyXSD, line 3555: dump trash only if `root == RootInode`

PR #4255, opened 2023-12-12
cmd/mount: add `cache-expire` option to set expired time for cached blocks

SandyXSD:
This PR takes over #3966 .

Note with the new option set, cached blocks will be expired even if the `cache-eviction` is "none".


pkg/chunk/disk_cache.go
```diff
@@ -467,12 +505,17 @@ func (cache *cacheStore) cleanup() {
...
 	var now = uint32(time.Now().Unix())
+	var cutoff = now - uint32(cache.cacheExpire/time.Second)
 	// for each two random keys, then compare the access time, evict the older one
 	for k, value := range cache.keys {
 		if value.size < 0 {
 			continue // staging
 		}
-		if cnt == 0 || lastValue.atime > value.atime {
+		if cache.cacheExpire > 0 && value.atime < cutoff {
+			lastK = k
```
davies, line 515: evict it immediately

pkg/chunk/disk_cache.go
```diff
@@ -149,6 +154,39 @@ func (cache *cacheStore) checkFreeSpace() {
...
+		}
+		if len(todel) > 0 {
+			logger.Debugf("cleanup expired cache (%s): %d blocks (%d MB), expired %d blocks (%d MB)", cache.dir, len(cache.keys), cache.used>>20, len(todel), freed>>20)
+		}
+		cache.Unlock()
+		for _, k := range todel {
+			_ = os.Remove(cache.cachePath(cache.getPathFromKey(k)))
+		}
+		todel = todel[:0]
+		time.Sleep(time.Minute * time.Duration(1-deleted/(cnt+1)))
```
davies, line 186: `deleted/(cnt+1)` is 0

davies: time.Sleep(cache.cacheExpire * (cnt-deleted)/(cnt+1))

davies (approved):

PR #4302, opened 2023-12-28
feat: dragonfly as read-only object storage

gaius-qi:
Dragonfly read requests via peer proxy and write requests via the default client of object storage. Backend storage now supports s3, oss and obs.

English document: https://github.com/juicedata/juicefs/files/13787601/Fluid.JuiceFS.Dragonfly-EN.V2.pdf
Chinese document: https://github.com/juicedata/juicefs/files/13787604/Fluid.JuiceFS.Dargonfly-ZH.V2.pdf

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,6 @@ require (
 	github.com/dgraph-io/badger/v3 v3.2103.5
 	github.com/dustin/go-humanize v1.0.1
 	github.com/erikdubbelboer/gspt v0.0.0-20210805194459-ce36a5128377
-	github.com/go-http-utils/headers v0.0.0-20181008091004-fed159eddc2a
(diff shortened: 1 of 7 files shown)


gaius-qi:
Advantages of using Dragonfly to proxy only read requests:
- Historical storage data can be switched to use Dragonfly as cache without compatibility problem.
- If JuiceFS write requests are supported, a dragonfly cluster needs to be deployed corresponding to an object storage. If it only supports read requests, Dragonfly can be deployed only one corresponding to many object storages as a public cache.

gaius-qi:
Performance testing:
https://www.chenshaowen.com/blog/performance-testing-and-comparison-of-juicefs-ce-ee-and-dragonfly.html#6-%E6%80%BB%E7%BB%93

PR #4319, opened 2024-01-08
cmd/load: fix TrimSuffix filePath

zhijian-pro:


cmd/load.go
```diff
@@ -102,7 +102,7 @@ func load(ctx *cli.Context) error {
 			if err != nil {
 				return fmt.Errorf("failed to get absolute path of %s: %s", src, err)
 			}
-			fileBlob, err := object.CreateStorage("file", strings.TrimRight(src, filepath.Base(srcAbsPath)), "", "", "")
+			fileBlob, err := object.CreateStorage("file", strings.TrimSuffix(src, filepath.Base(srcAbsPath)), "", "", "")
```
davies, line 105: The filename of src does not contain `/`, so the result is the same.

PR #4325, opened 2024-01-09
feat: Add Bunny.net object storage support

### LINKED ISSUE #4317: Support Bunny Object Storage

https://github.com/jamesponddotco/bunnystorage-go
https://bunny.net/storage/

Happy to provide a test Endpoint if needed.

l0wl3vel:
This PR implements support for Bunny.net object storage.

Fixes #4317


solracsf:
Code shoud go trough a [go fmt](https://go.dev/blog/gofmt).

zhijian-pro:
@l0wl3vel  Can you provide the format of the endpoint ?

zhijian-pro:
image
In fact, the S3 API is better suited for juicefs, and if the S3 api is released in the future, we may need to reimplement it with the S3 API.
@l0wl3vel @solracsf

l0wl3vel:
I know about the pending implementation of S3 for bunny storage. I would say this is a similar situation to B2 with their native API and their S3 compatible API.

I think it might be useful to have their native API supported for two reasons:

- It is really simple compared to S3. Four API calls that can be done via curl, if you desire

- Performance. Bunny Storage seems to perform much better than their competitors in a Multi-Cloud scenario. (I tested B2 and Cloudflare R2)

They advertise extremely low TTFB latency. I think it might be a good idea to have their native API implemented to compare to S3 when they offer it. Bunny's performance advantage could be caused partially by their own API which would justify having a bespoke implementation.

solracsf:
S3 is "coming soon" since 2022... 💭 
SFTP is broken (not using standards).

So API is the better option (also because there are no egreess fees using the API).

l0wl3vel:
Just you know, I noticed that the bunnystore-go library is not hosted on github and is using some homebrew dependencies for error handling and the http requests. I am going to fork it and rebase my PR on it.

solracsf:
> It is really simple compared to S3. Four API calls that can be done via curl, if you desire.

@l0wl3vel regarding this, if this can simplify things, is the external library `git.sr.ht/~jamesponddotco/bunnystorage-go` really necessary? Could the necessary code to make the API calls be integrated directly into JuiceFS codebase, instead of relying on an external lib?

l0wl3vel:
Maybe, but I do not think that is a good idea to basically put an API SDK inside of a specialized codebase the size of JuiceFS.

I thought about forking/redoing the `git.sr.ht/~jamesponddotco/bunnystorage-go` library, since I  think that it features some weird abstractions and it is using some homebrew libraries written by the author, some of which (e.g. git.sr.ht/~jamesponddotco/httpx-go) are deprecated.

@solracsf Would you be interested in building new bunnystorage-go library using more standard dependencies with me?

If the JuiceFS maintainers are fine with pulling in the jamesponddotco/bunnystorage-go dependency for now I would prefer it to just get it merged and working on replacing the API SDK with another implementation later on.

solracsf:
@l0wl3vel I'm not a Go developer, I'm afraid i can't help so much here 😿

zhijian-pro:
We want to merge a stable code. And one that is useful to users.

l0wl3vel:
Would you consider merging the code in its current form? And if not, what changes would be required to consider it for merging?

What kind of stability are we talking about? Upgrade Stability (Users can upgrade JuiceFS without need for migration), functional stability, dependency stability? Are there guidelines on what you would consider stable and what assurances are necessary to proof a certain kind of stability?

jamesponddotco:
@l0wl3vel Let me know what you need changed in the base module and I can see if I can find the time next week to work on it. I'll remove the need for 'httpx-go' this week, since I deprecated that.

l0wl3vel:
Hey, really appreciating you chiming in here. Thanks for writing bunnystorage-go. Here are a few things I think would improve it.
1. Combine the functionality of `Config` and `Client`. I do not see the reason to keep them separate.
2. Initialize the new `Client` with a single `New` function or with a builder-style semantic
3. Replace the `xstrings` logic for building URLs with `net/url`. I think it is much more idiomatic and easier to read.
4. Change the Enum-Style Abstraction for the endpoints. It requires updating the library every time a new endpoint is added by Bunny.net. Replace the field in the `Client` struct with a `net/url` and providing some constants to quickly fill in the endpoint with a known endpoint.
5. Remove the dependency on `xerrors`
6. Replace `UserAgent` with a string without any custom datatype.
7. Move the code to Github for visibility and better collaboration.

I think this is getting a bit off-topic here, but I already started converting bunnystorage-go to resty [on my fork](https://github.com/l0wl3vel/bunnystorage-go). Uploading does not work properly, unfortunately. Do you think you could move the main bunnystorage-go repo to github? 

No offense, I would really like to remove the dependency on sr.ht and your xstd-go entirely to make the module it more idiomatic and accessible.

jamesponddotco:
@l0wl3vel, thank you for the feedback, it's invaluable!

> Combine the functionality of Config and Client. I do not see the reason to keep them separate.
> Initialize the new Client with a single New function or with a builder-style semantic

I mostly like to keep configuration in its own struct, hence why they're separate. This is personal preference, really, so I don't see it changing. If you look at other clients I wrote, like the `cloudcraft-go` one for Datadog, you'll see the same style.

> Replace the xstrings logic for building URLs with net/url. I think it is much more idiomatic and easier to read.

`xstrings` is used to avoid allocations, but since I didn't really took precautions anywhere else, I could definitely remove it.

> Change the Enum-Style Abstraction for the endpoints. It requires updating the library every time a new endpoint is added by Bunny.net. Replace the field in the Client struct with a net/url and providing some constants to quickly fill in the endpoint with a known endpoint.

That was a mistake indeed, but fixing it would require a breaking change that I'm not sure I'm willing to make right now. I'll consider it, though.

> Remove the dependency on xerrors.

Using errors as constants is also a personal preference, but there is some reason behind it:
https://dave.cheney.net/2016/04/07/constant-errors

> Replace UserAgent with a string without any custom datatype.

Another mistake that would require a breaking change to undo, one that I make on every client, really. Not sure if this will be undone, though.

> Move the code to Github for visibility and better collaboration.

Not happening, sorry. I much prefer the email workflow for working with code and git in general, and prefer SourceHut to GitHub as a platform for several reasons. I still use GitHub as a read-only mirror and would probably accept PRs over here since I can't disable them, but that is as far as I'll go with the platform.

---

As for the usage of `xstd`, you'll see that module in pretty much all code I write, because I wrote it to decrease duplication. If I need to write the same function for two different projects, it goes into `xstd` instead.

I'll look into those changes this week, but I assume you'll use your fork instead for this PR. Still, glad my code could be of some help, haha.

l0wl3vel:
Ok, as I see it, the plan is the following:
- @jamesponddotco moves `bunnystorage-go` away from using `httpx`
- @zhijian-pro We move away from using the deprecated httpx dependency. Would the JuiceData team be willing to merge this PR? If not, what are your concerns and preferred methods for dealing with them?
- The only thing that concerns me about this PR is the format of the endpoint passed via CLI: I think something like `https://<endpoint>/<bucketname>` makes more sense than `bunny://<bucketname>.<endpoint>`. We stay closer to how the bunny API works internally and opens the path to easily moving away from `bunnystorage-go`, if the JuiceFS maintainers see a problem with using  it, since it is just the base URL for all API calls. I will implement the endpoint change this week.

l0wl3vel:
I built my own Bunnystorage library and rebuilt my JuiceFS implementation on it.
You can find the library here: https://github.com/l0wl3vel/bunny-storage-go-sdk

And the branch using my library can be found here: https://github.com/l0wl3vel/juicefs/tree/bunny-storage-sdk
I would really appreciate some external testing.

@zhijian-pro Do you still see any blockers on getting this PR merged?

l0wl3vel:
Found an undocumented API function, which allowed me to implement the Head functionality without the massive overhead, that the workaround would have incurred otherwise (You would have to list the directory of the file, parse the JSON response, filter out only the object and return it). That means, that things like `juicefs fsck` now work :tada:

pkg/object/bunny.go
```diff
@@ -0,0 +1,132 @@
...
+func (b *bunnyClient) Limits() Limits {
+	return Limits{
+		IsSupportMultipartUpload: false,
+		IsSupportUploadPartCopy:  false,
+	}
+}
+
+// Get the data for the given object specified by key.
+func (b *bunnyClient) Get(key string, off int64, limit int64) (io.ReadCloser, error) {
+	body, err := b.client.Download(key)
```
davies, line 49: Can it do range get? When is it will be closed?

go.mod
```diff
@@ -54,14 +55,14 @@ require (
...
-	github.com/sirupsen/logrus v1.9.0
+	github.com/sirupsen/logrus v1.9.3
 	github.com/smartystreets/goconvey v1.7.2
 	github.com/stretchr/testify v1.8.4
 	github.com/studio-b12/gowebdav v0.0.0-20230203202212-3282f94193f2
 	github.com/tencentyun/cos-go-sdk-v5 v0.7.45
 	github.com/tikv/client-go/v2 v2.0.4
 	github.com/upyun/go-sdk/v3 v3.0.4
-	github.com/urfave/cli/v2 v2.19.3
+	github.com/urfave/cli/v2 v2.25.7
```
davies, line 65: Is this necessary? If not, we can upgrade it with another PR, so we can track the changes easier.

davies: Same to other dependencies.

zhijian-pro:
> Found an undocumented API function, which allowed me to implement the Head functionality without the massive overhead, that the workaround would have incurred otherwise (You would have to list the directory of the file, parse the JSON response, filter out only the object and return it). That means, that things like `juicefs fsck` now work 🎉

Where did you find this one, I couldn't find documentation on it ?

pkg/object/bunny.go
l0wl3vel: I just took a look and the GET endpoint indicates support for `accept-ranges: bytes`. I am going to implement it into my library and remove the range read emulation with the section reader.

l0wl3vel: I implemented the download using ranges.

go.mod
l0wl3vel: How would I do that? These changes are introduced when running `go get github.com/l0wl3vel/bunny-storage-go-sdk@v0.0.9`.

davies: Never mind if you did not upgrade them manually.

zhijian-pro:
@l0wl3vel 

```
func TestBunny(t *testing.T) { //skip mutate
	if os.Getenv("BUNNY_ENDPOINT") == "" {
		t.SkipNow()
	}
	bunny, err := newBunny(os.Getenv("BUNNY_ENDPOINT"), "", os.Getenv("BUNNY_SECRET_KEY"), "")
	if err != nil {
		t.Fatalf("create: %s", err)
	}
	testStorage(t, bunny)
}
```
add this to `object_storage_test.go` and make sure it can pass the test.

l0wl3vel:
@zhijian-pro Is there something I can help with to make the tests pass? Do you require an account to test with?

zhijian-pro:
> @zhijian-pro Is there something I can help with to make the tests pass? Do you require a bunny.net account to run the tests on?

I've tested it before and it didn't pass, have you tested it manually? I don't think the current code is able to pass the TestBunny function.

l0wl3vel:
```
=== RUN   TestBunny
2024/02/05 10:48:42.058504 WARN RESTY parse "unit-test/测试编码文件{\"name\":\"juicefs\"}\x1f%uFF081%uFF09.jpg": net/url: invalid control character in URL, Attempt 1
2024/02/05 10:48:42.058546 ERROR RESTY parse "unit-test/测试编码文件{\"name\":\"juicefs\"}\x1f%uFF081%uFF09.jpg": net/url: invalid control character in URL
time="2024-02-05T10:48:42+01:00" level=error msg="Put Request Failed: parse \"unit-test/测试编码文件{\\\"name\\\":\\\"juicefs\\\"}\\x1f%uFF081%uFF09.jpg\": net/url: invalid control character in URL"
    object_storage_test.go:131: PUT testEncodeFile failed: parse "unit-test/测试编码文件{\"name\":\"juicefs\"}\x1f%uFF081%uFF09.jpg": net/url: invalid control character in URL
2024/02/05 10:48:42.058571 WARN RESTY parse "unit-test/测试编码文件{\"name\":\"juicefs\"}\x1f%uFF081%uFF09.jpg": net/url: invalid control character in URL, Attempt 1
2024/02/05 10:48:42.058573 ERROR RESTY parse "unit-test/测试编码文件{\"name\":\"juicefs\"}\x1f%uFF081%uFF09.jpg": net/url: invalid control character in URL
time="2024-02-05T10:48:42+01:00" level=error msg="Delete Request Failed: parse \"unit-test/测试编码文件{\\\"name\\\":\\\"juicefs\\\"}\\x1f%uFF081%uFF09.jpg\": net/url: invalid control character in URL"
    object_storage_test.go:180: out-of-range get: '', but got 0, error: 416 Requested Range Not Satisfiable
    object_storage_test.go:239: list3 should not return anything, but got 1
--- FAIL: TestBunny (0.95s)
FAIL
FAIL    github.com/juicedata/juicefs/pkg/object 1.001s
FAIL
```

Ok, so I have some questions here. 

1.  `net/url` from the standard library is not able to parse the URL you are trying to input. Throwing an error is apparently not ok, since the test case failed. Other object stores (e.g. Backblaze B2) exhibit the same behavior. What is the expected behavior here if the Object Store just does not support the input? Solutions like masking the invalid characters sound like a terrible idea, since you would risk file collisions.
2. When requesting a range that is not available the Endpoint throws a 416, as expected. But JuiceFS expects this masked as an empty byte slice. Again, B2 behaves exactly the same. I am not sure if changes are necessary here.
3. The list returning objects that are not expected is an error in my code. After looking into it, it looks like the test case passes a continuation marker into the listAll function and has some expectations about the output of the list enpoint being in lexical order. I will try to emulate this behavior, since the Bunny API does not work with continuation tokens and might not return elements in lexicographic order.


B2 Tests: https://github.com/juicedata/juicefs/issues/4317#issuecomment-1902367114

solracsf:
> net/url from the standard library is not able to parse the URL you are trying to input. Throwing an error is apparently not ok, since the test case failed.

Can https://github.com/go-resty/resty/issues/533 help here?

l0wl3vel:
> > net/url from the standard library is not able to parse the URL you are trying to input. Throwing an error is apparently not ok, since the test case failed.
> 
> Can [go-resty/resty#533](https://github.com/go-resty/resty/issues/533) help here?

IMO, masking the error won't resolve the issue. The bunny API requires you to encode the filename as an URL. If that is not possible because your URL contains control characters and invalid UTF8, you throw an error, like other Object storage implementations do.

I am not sure what to do here, since other Object Storage implementations have the same "issue" but work fine, since JuiceFS will not ever upload or request an object with this kind of name. It is a test case that tests a "feature" that is not used in JuiceFS.

As far as I can tell, the test is wrong, other upstream implementations fail the test as well and work fine.

@zhijian-pro @jamesponddotco Are there any tests that test file system functionality, instead of testing if the object store is accepting the same requests as S3? Would you also be willing to hop on a call to discuss what is required to get this upstream?

zhijian-pro:
@l0wl3vel We use `Fatalf` for fatal errors, and `Logf` for those that don't affect the majority of cases, so you only need to fix `Fatalf` errors.

zhijian-pro:
> since JuiceFS will not ever upload or request an object with this kind of name.

No, juicefs sync maybe will, juicefs sync synchronizes objects, and object names are user-defined

l0wl3vel:
> @l0wl3vel We use `Fatalf` for fatal errors, and `Logf` for those that don't affect the majority of cases, so you only need to fix `Fatalf` errors.
> 
> ```
> object_storage_test.go:239: list3 should not return anything, but got 1
> ```
> 
> About list api We need to support the marker parameter，If bunny doesn't support it natively, you need to support it manually. The bunny list is similar to the linux ls command, and what we want is to return all objects in the directory including objects in subdirectories. Since the bunny list method is a hierarchical list, it may be better to implement the list interface, see [sftp.go](https://github.com/juicedata/juicefs/blob/fc89521434b68718ede2ac3dbebcaf7411b359e8/pkg/object/sftp.go#L363C21-L363C25) for an example.

Thank you for clarifying. I will implemente the marker functionality and the recursive listing.

l0wl3vel:
`object_storage_test.go` is passing now :tada:

zhijian-pro:
> `object_storage_test.go` is passing now 🎉

I will review this pr again after the Spring Festival.

l0wl3vel:
Updated the bunny-storage-go-sdk to 0.0.10, which uses fasthttp instead of net/http and broke 1GiB/s in write speed and got some nice boost in read speed as well:

```
BlockSize: 1 MiB, BigFileSize: 1024 MiB, SmallFileSize: 128 KiB, SmallFileCount: 100, NumThreads: 1
Time used: 0.0 s, CPU: NaN%, Memory: 0.0 MiB
+------------------+----------------+---------------+
|       ITEM       |      VALUE     |      COST     |
+------------------+----------------+---------------+
|   Write big file |  1001.05 MiB/s |   1.02 s/file |
|    Read big file |   706.92 MiB/s |   1.45 s/file |
| Write small file |   24.6 files/s | 40.63 ms/file |
|  Read small file | 1540.2 files/s |  0.65 ms/file |
|        Stat file | 6314.6 files/s |  0.16 ms/file |
|   FUSE operation |   0 operations |    0.00 ms/op |
|      Update meta |   0 operations |    0.00 ms/op |
|       Put object |   0 operations |    0.00 ms/op |
|       Get object |   0 operations |    0.00 ms/op |
|    Delete object |   0 operations |    0.00 ms/op |
| Write into cache |   0 operations |    0.00 ms/op |
|  Read from cache |   0 operations |    0.00 ms/op |
+------------------+----------------+---------------+
```

pkg/object/bunny.go
```diff
@@ -0,0 +1,193 @@
...
+		normalizedObjectNameWithinZone(object),
+		int64(object.Length),
+		lastChanged,
+		object.IsDirectory,
+		"",
+	}
+}
+
+func (b *bunnyClient) Head(key string) (Object, error) {
+	logger.Debug(key)
```
davies, line 167: too much logging in debug mode, can you remove it or change it to trace level.

pkg/object/bunny.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+// Delete a object.
+// Requires a conditional retry, since deleting a directory or file called foo/bar requires two different calls to the Bunny API, which JuiceFS does not do
+// Deleting a directory requires a trailing slash in the key to delete, which JuiceFS does not add to the path, leading to test case failures.
+// We implement a conditional retry here to try deleting the directory if the delete for a key of the passed name fails.
+// Deleting keys that do not exist are expected to not throw an error
+func (b *bunnyClient) Delete(key string) error {
+	if err := b.client.Delete(key, false); err != nil {
+		if err.Error() == "Not Found" {
```
davies, line 79: This error message is confusing, we should only retry it when the key is ending with `/`.

davies: Or call `b.client.Delete(key, strings.HasSuffix("/"))`

pkg/object/bunny.go
```diff
@@ -0,0 +1,193 @@
...
+		b.walkObjects(prefix, marker, out)
+	}()
+
+	return out, nil
+}
+
+func (b *bunnyClient) walkObjects(prefix string, marker string, out chan<- Object) {
+	objects, err := b.client.List(prefix)
+	if err != nil {
+		return
```
davies, line 109: We should log this err and send `nil` into `out`, so the consumer know that something is wrong.

pkg/object/bunny.go
```diff
@@ -0,0 +1,193 @@
...
+			}
+			if markerEncountered {
+				if o.IsDirectory {
+					objects, err := b.client.List(normalizedPath)
+					if err != nil {
+						return
+					}
+					objectsToProcess = append(objectsToProcess, objects...)
+				}
+				out <- parseObjectMetadata(o)
```
davies, line 139: The objects should be ordered by key, for example,

a/
a/a
a/b
b/
b/a
b/b

pkg/object/bunny.go
```diff
@@ -0,0 +1,193 @@
...
+
+type bunnyClient struct {
+	DefaultObjectStorage
+	client   *bunnystorage.Client
+	endpoint string
+}
+
+// Description of the object storage.
+func (b *bunnyClient) String() string {
+	return b.endpoint
```
davies, line 40: it should be `bunny://endpoint`

pkg/object/bunny.go
l0wl3vel: This workaround is necessary to pass the tests.  The boolean flag on `b.client.Delete()` only appends a / character to the path, which means that setting it at all should not be necessary.

Tests fail on repeated test runs with the following implementation:

```
func (b *bunnyClient) Delete(key string) error {
	if err := b.client.Delete(key, false); err != nil {
		if err.Error() == "Not Found" {
			return nil
		} else {
			return err
		}
	}
	return nil
}
```

Well, apparently some tests do not consistently pass paths with a trailing /. The first run works fine, but on any subsequent run a list test fails because there are empty directories left over from the first test run.

### First run
```
=== RUN   TestBunny
    object_storage_test.go:131: PUT testEncodeFile failed: Unauthorized
    object_storage_test.go:180: out-of-range get: '', but got 0, error: Requested Range Not Satisfiable
    object_storage_test.go:284: list with delimiter is not supported
    object_storage_test.go:326: list with delimiter is not supported
    object_storage_test.go:527: https://storage.bunnycdn.com/juicefs-testunit-test/ does not support multipart upload: not supported
--- PASS: TestBunny (10.97s)
PASS
```
### Object storage directory after the first run
image

### Second Run
```
go1.20.13 test -v -tags gluster -run TestBunny github.com/juicedata/juicefs/pkg/object      
=== RUN   TestBunny
    object_storage_test.go:131: PUT testEncodeFile failed: Unauthorized
    object_storage_test.go:180: out-of-range get: '', but got 0, error: Requested Range Not Satisfiable
    object_storage_test.go:222: First key should be test, but got a/
--- FAIL: TestBunny (0.65s)
FAIL
FAIL    github.com/juicedata/juicefs/pkg/object 1.020s
FAIL
```

This workaround seems to be necessary, since it is not guaranteed that JuiceFS passes in the trailing /.

pkg/object/bunny.go
l0wl3vel: Do you think it would be necessary that `ListAll` also returns an error if a `b.client.List` calls fails?

pkg/object/bunny.go
l0wl3vel: Ok, I am going to implement the list function instead of the listAll method. That should fix this behavior.

pkg/object/bunny.go
davies: lgtm

l0wl3vel: I retested it after replacing ListAll with List and the tests fail again. With or without my deletion workaround there are objects left in the unit-test directory. Subsequent runs will always fail again. Is it possible that testStorage does not handle the cleanup of the unit-test directory on object stores, that allow prefixes containing only empty directories? I would propose cleaning up the root directory of the storage prefix like this:

```
diff --git a/pkg/object/object_storage_test.go b/pkg/object/object_storage_test.go
index cfab5b97..a806c4ef 100644
--- a/pkg/object/object_storage_test.go
+++ b/pkg/object/object_storage_test.go
@@ -120,6 +120,7 @@ func testStorage(t *testing.T, s ObjectStorage) {
        }
        prefix := "unit-test/"
        s = WithPrefix(s, prefix)
+       defer s.Delete("/")
        defer func() {
                if err := s.Delete("test"); err != nil {
                        t.Fatalf("delete failed: %s", err)
```

pkg/object/bunny.go
l0wl3vel: ListAll was replaced with List

pkg/object/bunny.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func (b *bunnyClient) List(prefix, marker, delimiter string, limit int64, followLink bool) ([]Object, error) {
+	if delimiter != "/" {
+		return nil, notSupported
+	}
+
+	//listInParentDirectory := !strings.HasSuffix(prefix, "/")
+
+	listedObjects, err := b.client.List(path.Dir(prefix))
```
davies, line 102: Please take the file.go as an example

pkg/object/bunny.go
l0wl3vel: It seems like this problem only appears on Object Storage providers that do support the concept of an empty directory, which S3 style APIs do not support. We would need to adjust the tests to clean up the unit-test directory properly.

SandyXSD:
@l0wl3vel Hi, it seems there's still an issue relating to the 'list' method that is not solved yet. Are you still working on this?

l0wl3vel:
@SandyXSD I still want to finish it. I already have the changes implemented, but I can´t make sure it works, since Bunny Object Storage has the concept of empty directories, which trips up the test suite when running it multiple times, because there are empty directories left from the previous run. I just pushed my changes.

pkg/object/bunny.go
```diff
@@ -0,0 +1,186 @@
...
+	var output []Object
+	var dir = prefix
+	if !strings.HasSuffix(dir, dirSuffix) { // If no Directory list in parent directory
+		dir = path.Dir(dir)
+		if !strings.HasSuffix(dir, dirSuffix) {
+			dir += dirSuffix
+		}
+	} else if marker == "" { // If Directory && no marker: Return prefix directory as well
+		parentPath := path.Dir(path.Dir(prefix))
+		objects, err := b.client.List(parentPath+dirSuffix)
```
davies, line 103: We just need to add the directory itself here, will list it later. Please follow the example in file.go

l0wl3vel: I am not sure what changes you would like to see. The code here is based on sftp.go, like you suggested before. Could you elaborate please on what about it is incorrect?

davies: Here is how sftp implements List:
```
	var objs []Object
	dir := f.path(prefix)
	if !strings.HasSuffix(dir, "/") {
		dir = filepath.Dir(dir)
		if !strings.HasSuffix(dir, dirSuffix) {
			dir += dirSuffix
		}
	} else if marker == "" {
		obj, err := f.Head(prefix)
		if err != nil {
			if os.IsNotExist(err) {
				return nil, nil
			}
			return nil, err
		}
		objs = append(objs, obj)
	}
	infos, err := c.sftpClient.ReadDir(dir)
	if err != nil {
		if os.IsPermission(err) {
			logger.Warnf("skip %s: %s", dir, err)
			return nil, nil
		}
		if os.IsNotExist(err) {
			return nil, nil
		}
		return nil, err
	}

	entries := f.sortByName(c.sftpClient, dir, infos, followLink)
	for _, o := range entries {
		key := o.Key()
		if !strings.HasPrefix(key, prefix) || (marker != "" && key <= marker) {
			continue
		}
		objs = append(objs, o)
		if len(objs) == int(limit) {
			break
		}
	}
```
When the marker is `''`, it try to return the prefix self (could be a directory). So instead of list the parent of prefix, can we just call Head() instead?

davies:
@l0wl3vel please rebase to main branch and fix the conflict, I think we are very close to merge this one.

pkg/object/bunny.go
l0wl3vel: I am not sure, but listing a prefix in S3 does not only return the virtual prefix "directory", rather all keys that start with that prefix. This would mean that I implemented the correct behavior. Working on this is also really difficult, since the tests for object storage are very incomplete and miss a lot of corner cases around listing objects.

davies:
This one is took over by #4719, thanks very much!

PR #4331, opened 2024-01-10
object/obs: don't decode the object key in list result

zhijian-pro:


davies:
@zhijian-pro please update the unit test to cover this issue

PR #4333, opened 2024-01-10
cmd/compact: add new compact cmd, used for specific path

### LINKED ISSUE #4303: Support triggering compaction of all files under a given directory
Support triggering compaction of all files under a given directory

**Why is this needed**:
An alternative light way of `gc --compact` for large clusters.

jiefenghuang:
close #4303

### DIFF
--- a/cmd/compact_test.go
+++ b/cmd/compact_test.go
@@ -1,0 +1,111 @@
+/*
+ * JuiceFS, Copyright 2020 Juicedata, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func createTestFile(path string, size int) error {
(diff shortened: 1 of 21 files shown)


cmd/compact_test.go
```diff
@@ -0,0 +1,111 @@
+/*
+ * JuiceFS, Copyright 2020 Juicedata, Inc.
```
davies, line 2: 2024

pkg/utils/file.go
```diff
@@ -0,0 +1,35 @@
...
+
+package utils
+
+import (
+	"fmt"
+	"path/filepath"
+)
+
+func GetInode(path string) (uint64, error) {
+	dpath, err := filepath.Abs(path)
```
davies, line 25: Is this necessary?

pkg/meta/sql_lock.go
```diff
@@ -249,9 +249,9 @@ func (m *dbMeta) Setlk(ctx Context, inode Ino, owner_ uint64, block bool, ltype
 	return err
 }
 
-func (r *dbMeta) ListLocks(ctx context.Context, inode Ino) ([]PLockItem, []FLockItem, error) {
+func (m *dbMeta) ListLocks(ctx context.Context, inode Ino) ([]PLockItem, []FLockItem, error) {
```
davies, line 252: We can move these into another PR.

cmd/compact_unix.go
```diff
@@ -0,0 +1,199 @@
...
+	compactConcurrency := ctx.Int("compact-concurrency")
+	if compactConcurrency <= 0 {
+		logger.Warn("thread number should be > 0")
+		compactConcurrency = 1
+	}
+
+	// new meta client
+	metaConf := meta.DefaultConf()
+	metaConf.MaxDeletes = int(deleteConcurrency)
+	client := meta.NewClient(metaUri, metaConf)
```
SandyXSD, line 105: Don't use meta client, instead write control message to a mount point. Check these codes:

- https://github.com/juicedata/juicefs/blob/7b85645cac16c8442b3b7d35365a99dc0e339cdd/cmd/info.go#L116
- https://github.com/juicedata/juicefs/blob/7b85645cac16c8442b3b7d35365a99dc0e339cdd/pkg/vfs/internal.go#L297

PR #4342, opened 2024-01-15
doc: S3Gateway enable virtual-host-style request

zhijian-pro:


davies:
We can add `--domain` for gateway to enable this.

PR #4346, opened 2024-01-15
cmd: refactor os based code

### LINKED ISSUE #4343: CMD: Refactor Operating System-Specific Code in cmd
Refactor Operating System-Specific Code in cmd

**Why is this needed**:
Currently, many commands have conditional checks in the code to differentiate behavior based on the operating system.
To address this, let's introduce a new file called 'cmd_list' to classify the supported command sets for different operating systems.

jiefenghuang:
close #4343

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -26,6 +26,8 @@ import (
 	"sync"
 	"syscall"
 
+	"go.uber.org/automaxprocs/maxprocs"
+
 	"github.com/erikdubbelboer/gspt"
 	"github.com/google/uuid"
 	"github.com/juicedata/juicefs/pkg/utils"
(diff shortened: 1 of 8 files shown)


cmd/main.go
```diff
@@ -55,36 +56,7 @@ func Main(args []string) error {
...
-			cmdMdtest(),
-			cmdWarmup(),
-			cmdRmr(),
-			cmdSync(),
-			cmdDebug(),
-			cmdClone(),
-			cmdSummary(),
-			cmdCompact(),
-		},
+		Commands:             GetCmds(),
```
SandyXSD, line 59:
```suggestion
		Commands:             getCmds(),
```

cmd/rmr.go
```diff
@@ -59,10 +58,6 @@ func openController(dpath string) (*os.File, error) {
 
 func rmr(ctx *cli.Context) error {
 	setup(ctx, 1)
-	if runtime.GOOS == "windows" {
```
davies, line 62: I think the current implementation is better, because we have not find a way to communicate with running client, could be fixed in future.

PR #4370, opened 2024-01-19
Feat: cmd-warmup add option "check" and "evict"

### LINKED ISSUE #4359: cmd/warmup: support `--check` and `--evict` options
Refs: https://juicefs.com/docs/cloud/reference/command_reference#options-15

**Why is this needed**:

jiefenghuang:
close #4359

### DIFF
--- a/cmd/warmup.go
+++ b/cmd/warmup.go
@@ -19,6 +19,7 @@ package cmd
 import (
 	"bufio"
 	"encoding/binary"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
@@ -28,8 +29,10 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/dustin/go-humanize"
 	"github.com/juicedata/juicefs/pkg/meta"
 	"github.com/juicedata/juicefs/pkg/utils"
+	"github.com/juicedata/juicefs/pkg/vfs"
 	"github.com/urfave/cli/v2"
 )
 
@@ -70,6 +73,14 @@ $ juicefs warmup -f /tmp/filelist`,
 				Name:    "background",
 				Aliases: []string{"b"},
 				Usage:   "run in background",
+			},
+			&cli.BoolFlag{
+				Name:  "evict",
+				Usage: "evict cached blocks",
+			},
+			&cli.BoolFlag{
+				Name:  "check",
+				Usage: "check whether the data blocks are cached or not",
 			},
 		},
 	}
@@ -128,7 +139,7 @@ END:
 }
 
 // send fill-cache command to controller file
-func sendCommand(cf *os.File, batch []string, threads uint, background bool, dspin *utils.DoubleSpinner) {
+func sendCommand(cf *os.File, action vfs.CacheAction, batch []string, threads uint, background bool, dspin *utils.DoubleSpinner) vfs.CacheResponse {
 	paths := strings.Join(batch, "\n")
 	var back uint8
 	if background {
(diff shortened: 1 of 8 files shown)


pkg/vfs/fill.go
```diff
@@ -184,23 +246,85 @@ func (v *VFS) walkDir(ctx meta.Context, inode Ino, todo chan _file) {
...
+	iter.nextSliceIndex++
+	return s
+}
+
+func (iter *sliceIterator) Iterate(handler sliceHandler) error {
+	for iter.hasNext() {
+		s := iter.next()
+		iter.stat.count++
+		iter.stat.bytes += uint64(s.Size)
+		if handler == nil {
```
SandyXSD, line 304: check it before the `for` loop

pkg/vfs/fill.go
```diff
@@ -34,36 +34,88 @@ type _file struct {
...
 					}
-					_ = v.Meta.Close(ctx, f.ino)
 				}
-				if count != nil {
-					atomic.AddUint64(count, 1)
+
+				// log and skip error
+				err := iter.Iterate(handler)
+				if err != nil {
+					logger.Error(fmt.Errorf("%s error: %w", action, err))
```
SandyXSD, line 110: `logger.Errorf`

pkg/vfs/fill.go
```diff
@@ -184,23 +246,85 @@ func (v *VFS) walkDir(ctx meta.Context, inode Ino, todo chan _file) {
...
+type sliceIterStat struct {
+	count uint64
+	bytes uint64
+}
+
+type sliceIterator struct {
+	ctx      meta.Context
+	mClient  meta.Meta
+	ino      Ino
+	size     uint64
```
SandyXSD, line 258: `size` is not needed.

pkg/vfs/fill.go
```diff
@@ -34,36 +34,88 @@ type _file struct {
...
+						_ = v.Meta.Close(ctx, f.ino)
+					}
+				case EvictCache:
+					handler = func(s meta.Slice) error {
+						return v.Store.EvictCache(s.Id, s.Size)
+					}
+				case CheckCache:
+					handler = func(s meta.Slice) error {
+						missBytes, err := v.Store.CheckCache(s.Id, s.Size)
+						if err != nil {
```
SandyXSD, line 97: for now err can be ignored.

cmd/warmup.go
```diff
@@ -214,11 +249,20 @@ func warmup(ctx *cli.Context) error {
...
+	} else if check {
+		action = vfs.CheckCache
+	}
+
 	background := ctx.Bool("background")
 	start := len(mp)
 	batch := make([]string, 0, batchMax)
 	progress := utils.NewProgress(background)
-	dspin := progress.AddDoubleSpinner("Warming up")
+	dspin := progress.AddDoubleSpinner(string(action))
```
SandyXSD, line 264: use `action.String()`

pkg/vfs/fill.go
jiefenghuang: This error handling is kept here, based on the check API.

pkg/vfs/fill.go
```diff
@@ -34,36 +34,88 @@ type _file struct {
...
+			defer wg.Done()
+			for f := range todo {
+				if ctx.Canceled() {
+					return
 				}
-				if err := v.fillInode(ctx, f.ino, f.size, bytes); err != nil {
-					logger.Errorf("Inode %d could be corrupted: %s", f.ino, err)
+
+				if f.ino == 0 {
+					continue
```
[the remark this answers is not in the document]
jiefenghuang: if ino == 0 happen, that means all the other progress should be stopped, maybe we should call ctx.Cancel()
``` go
ctx.Cancel()
return
```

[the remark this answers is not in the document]
SandyXSD: Usually `cancel` is cased by user interruption, like Ctrl+C, while `ino == 0` implies internal error.

PR #4381, opened 2024-01-25
feat: add Rename test

jiefenghuang:


pkg/meta/random_test.go
```diff
@@ -932,7 +934,7 @@ func (m *fsMachine) listxattr(inode Ino) ([]byte, syscall.Errno) {
 
 func (m *fsMachine) Mkdir(t *rapid.T) {
 	parent := m.pickNode(t)
-	name := rapid.StringN(1, 200, 255).Draw(t, "name")
+	name := rapid.StringN(3, 200, 255).Draw(t, "name")
```
SandyXSD, line 937: why we need this change?

pkg/meta/random_test.go
```diff
@@ -691,12 +685,23 @@ func (m *fsMachine) rename(srcparent Ino, srcname string, dstparent Ino, dstname
...
+	srcnode := src.children[srcname]
+	if srcnode == nil {
+		return syscall.ENOENT
+	}
+
+	if !src.stickyAccess(srcnode, m.ctx.Uid()) {
 		return syscall.EACCES
 	}
+
+	// 目录拥有者不能rename其他用户子目录
```
SandyXSD, line 698: write comment in English

pkg/meta/random_test.go
jiefenghuang: already skip . and .. in the next judge, forget to convert it

PR #4390, opened 2024-01-31
cmd/sync: fix the dst password is not removed

zhijian-pro:


cmd/main.go
```diff
@@ -344,15 +344,19 @@ func setup(c *cli.Context, n int) {
...
-	if uri2 != uri {
-		for i, a := range os.Args {
-			if a == uri {
-				args[i] = uri2
-				break
+	var idx int
+	for _, uri := range uris {
+		uri2 := utils.RemovePassword(uri)
+		if uri2 != uri {
+			for i := idx; i < len(os.Args); i++ {
```
davies, line 354: args

Issue #4409, opened 2024-02-18
POSIX ACL


jiefenghuang:
**Why is this needed**:
support [POSIX ACL](https://www.usenix.org/legacy/publications/library/proceedings/usenix03/tech/freenix03/full_papers/gruenbacher/gruenbacher_html/main.html)

- link: #2559 
- link: #512 

---
**What would you like to be added**:
# basic ACL functionality
## reference
- https://github.com/torvalds/linux/commit/60bcc88ad185d512f5718f2f8dcccb483ea8fb73

## what to be added
- add enable-acl flag (also implies default_permissions) to check access acl
  - TODO: For SDK users, user-space ACL checking is required.
- update acl: in setxattr
- default acl: inheritance of default ACLs when new filesystem nodes are
created
- sync between file mode and acl: setattr and setxattr

# compatibility
### why is this needed
- Allowing client mount without ACL will disrupt the synchronization of file mode and ACL.
- Disabling ACL requires a time-consuming conversion of the group class part of file modes for all files (from mask to owner group), so currently, disabling ACL is not supported. 

### what to be added
- add enable-acl flag to meta format
- set fuse.MountOptions
  - DisableXAttrs = !(enable-acl || enable-xattr)
  - EnableAcl = enable-acl
  - IgnoreSecurityLabels = !enable-acl

### why is this needed
If an older version client mounts a newer version of meta format, it will disrupt the synchronization of mode and ACL.

### what to be added
Setting the matching versions for the client and meta.

### cases
1. old client <-> new fs without enable-acl, supported
2. old client <-> new fs with enable-acl, not supported
- logic of "setattr, getattr, mknod" is not compatible
- we need to change format's MinClientVersion when config "enable-acl" to true.
4. new client <-> old fs, supported.  
- two new fields will be added to attr. SQL will sync the table, and KV storage is in []byte format, will ensuring compatibility in encode and decode.
- also support "enable-acl" for old fs.

# storage
## what to be added
- Implement an ACL value sharing mechanism
- We assume that the number of ACLs should be very limited. Having too many ACLs may indicate issues in the way they are being used, and it may be necessary to optimize them through group settings.

1. limited acls (ex. 65536)
- Adding two uint32 variables in the attr to store the index of access ACL and default ACL, enabling quick determination of whether a file/dir has ACL set.
- ACL value is stored in meta with the key 'acl_[index]'.
  - index will keep increasing
  - unused index will not be removed.
- Cache all ACL value in memory

# test
[...]

PR #4421, opened 2024-02-23
feat/acl: add enable-acl flag and version checking

jiefenghuang:
#4409 
- add enable-acl flag
- add version checking

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -242,6 +242,11 @@ func config(ctx *cli.Context) error {
 				format.MaxClientVersion = new
 				clientVer = true
 			}
+		case "enable-acl":
+			// cannot disable
+			if ctx.Bool(flag) {
+				format.EnableACL = true
+			}
 		}
 	}
 	if msg.Len() == 0 {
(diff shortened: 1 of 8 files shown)


cmd/format.go
```diff
@@ -435,7 +440,8 @@ func format(c *cli.Context) error {
 			TrashDays:        c.Int("trash-days"),
 			DirStats:         true,
 			MetaVersion:      meta.MaxVersion,
-			MinClientVersion: "1.1.0-A",
```
davies, line 438: Don't enable ACL by default

cmd/config.go
```diff
@@ -242,6 +242,11 @@ func config(ctx *cli.Context) error {
 				format.MaxClientVersion = new
 				clientVer = true
 			}
+		case "enable-acl":
+			// cannot disable
+			if ctx.Bool(flag) {
+				format.EnableACL = true
```
davies, line 248: We should update the MinClientVersion once enable ACL

cmd/format.go
davies: We should set `MinClientVersion` to `1.2` when enable-acl is true.

cmd/config.go
davies: By default, new volume created by 1.2 should be accessible by 1.2-, then when ACL is enabled, we should update `MinClientVersion` to `1.2`

PR #4437, opened 2024-02-29
feat/acl: add acl rule and cache

jiefenghuang:
#4409
add acl rule and acl cache which will cache all acl rules

### DIFF
--- a/pkg/acl/acl.go
+++ b/pkg/acl/acl.go
@@ -63,7 +63,7 @@ func (es *Entries) String() string {
 }
 
 func (es *Entries) Encode() []byte {
-	w := utils.NewBuffer(uint32(es.Len() * 6))
+	w := utils.NewBuffer(4 + uint32(es.Len()*6))
 	for _, e := range *es {
 		w.Put32(e.Id)
 		w.Put16(e.Perm)
@@ -96,20 +96,47 @@ func (r *Rule) String() string {
 }
 
 func (r *Rule) Encode() []byte {
-	w := utils.NewBuffer(uint32(8 + (len(r.NamedUsers)+len(r.NamedGroups))*6))
+	w := utils.NewBuffer(uint32(16 + (len(r.NamedUsers)+len(r.NamedGroups))*6))
 	w.Put16(r.Owner)
 	w.Put16(r.Group)
 	w.Put16(r.Mask)
 	w.Put16(r.Other)
+	w.Put32(uint32(len(r.NamedUsers)))
 	for _, entry := range r.NamedUsers {
 		w.Put32(entry.Id)
 		w.Put16(entry.Perm)
 	}
+	w.Put32(uint32(len(r.NamedGroups)))
 	for _, entry := range r.NamedGroups {
 		w.Put32(entry.Id)
 		w.Put16(entry.Perm)
 	}
 	return w.Bytes()
+}
+
+func (r *Rule) Decode(buf []byte) {
+	rb := utils.ReadBuffer(buf)
+	r.Owner = rb.Get16()
+	r.Group = rb.Get16()
+	r.Mask = rb.Get16()
+	r.Other = rb.Get16()
+	uCnt := rb.Get32()
+	if uCnt != 0 {
+		r.NamedUsers = make([]Entry, uCnt)
+		for i := 0; i < int(uCnt); i++ {
+			r.NamedUsers[i].Id = rb.Get32()
+			r.NamedUsers[i].Perm = rb.Get16()
+		}
+	}
+
+	gCnt := rb.Get32()
+	if gCnt != 0 {
+		r.NamedGroups = make([]Entry, gCnt)
+		for i := 0; i < int(gCnt); i++ {
+			r.NamedGroups[i].Id = rb.Get32()
+			r.NamedGroups[i].Perm = rb.Get16()
+		}
+	}
 }
 
 func EmptyRule() *Rule {
--- a/pkg/acl/acl.go
+++ b/pkg/acl/acl.go
@@ -19,8 +19,6 @@ package acl
 import (
 	"fmt"
 	"hash/crc32"
-	"sort"
-	"strings"
 
 	"github.com/juicedata/juicefs/pkg/utils"
 )
@@ -42,8 +40,6 @@ func (es *Entries) IsEqual(other *Entries) bool {
 	if es.Len() != other.Len() {
 		return false
 	}
-	sort.Sort(es)
-	sort.Sort(other)
 	for i := 0; i < es.Len(); i++ {
 		if (*es)[i].Id != (*other)[i].Id || (*es)[i].Perm != (*other)[i].Perm {
 			return false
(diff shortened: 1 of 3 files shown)


jiefenghuang:
commit 860f4b86d90713d092c897d04ed2ea4e624deda3 is for code review only, not the final version.

pkg/acl/acl.go
```diff
@@ -0,0 +1,230 @@
...
+	builder.WriteString("{")
+	for _, e := range *es {
+		builder.WriteString(fmt.Sprintf("{id: %d, perm: %o}, ", e.Id, e.Perm))
+	}
+	builder.WriteString("}")
+	return builder.String()
+}
+
+func (es *Entries) Encode() []byte {
+	w := utils.NewBuffer(4 + uint32(es.Len()*6))
```
davies, line 66: why 4?

pkg/acl/acl.go
```diff
@@ -0,0 +1,230 @@
...
+
+func (es *Entries) Len() int           { return len(*es) }
+func (es *Entries) Less(i, j int) bool { return (*es)[i].Id < (*es)[j].Id }
+func (es *Entries) Swap(i, j int)      { (*es)[i], (*es)[j] = (*es)[j], (*es)[i] }
+
+func (es *Entries) IsEqual(other *Entries) bool {
+	if es.Len() != other.Len() {
+		return false
+	}
+	sort.Sort(es)
```
davies, line 45: We can assume it's ordered, call sort for newly created entries.

pkg/acl/acl.go
```diff
@@ -0,0 +1,230 @@
...
+	Group       uint16
+	Mask        uint16
+	Other       uint16
+	NamedUsers  Entries
+	NamedGroups Entries
+}
+
+func (r *Rule) String() string {
+	return fmt.Sprintf("owner %o, group %o, mask %o, other %o, named users: %s, named group %s",
+		r.Owner, r.Group, r.Mask, r.Other, r.NamedUsers.String(), r.NamedGroups.String())
```
davies, line 95: `%s` will call String()

pkg/acl/acl.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+func (r *Rule) Decode(buf []byte) {
+	rb := utils.ReadBuffer(buf)
+	r.Owner = rb.Get16()
+	r.Group = rb.Get16()
+	r.Mask = rb.Get16()
+	r.Other = rb.Get16()
+	uCnt := rb.Get32()
+	if uCnt != 0 {
```
davies, line 124: it's OK to `make([]Entry, 0)`

pkg/acl/acl.go
```diff
@@ -0,0 +1,230 @@
...
+	if r.IsMinimal() {
+		return ((r.Owner & 7) << 6) | ((r.Group & 7) << 3) | (r.Other & 7)
+	}
+	return ((r.Owner & 7) << 6) | ((r.Mask & 7) << 3) | (r.Other & 7)
+}
+
+func (r *Rule) ChildAccessACL(mode uint16) *Rule {
+	cRule := &Rule{}
+	cRule.Owner = (mode >> 6) & 7
+	cRule.Owner &= r.Owner
```
davies, line 206: r.Owner & (mode >> 6) & 7

pkg/acl/acl.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+func (r *Rule) GetMode() uint16 {
+	if r.IsMinimal() {
+		return ((r.Owner & 7) << 6) | ((r.Group & 7) << 3) | (r.Other & 7)
+	}
+	return ((r.Owner & 7) << 6) | ((r.Mask & 7) << 3) | (r.Other & 7)
+}
+
+func (r *Rule) ChildAccessACL(mode uint16) *Rule {
```
davies, line 203: Add doc for these public API

pkg/acl/cache.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	if ids, ok := c.cksum2Id[r.Checksum()]; ok {
+		for _, id := range ids {
+			if r.IsEqual(c.id2Rule[id]) {
+				return id
+			}
+		}
+	}
+	return None
```
davies, line 139: return new id？

pkg/acl/cache.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+type cache struct {
+	lock     sync.RWMutex
+	maxId    uint32
+	id2Rule  map[uint32]*Rule
+	cksum2Id map[uint32][]uint32
+}
+
+func (c *cache) GetMissIds(maxId uint32) []uint32 {
```
davies, line 54: Assume all the IDs are continuous, should we scan the database with prefix or table, rather than generating missing ids?

pkg/acl/cache.go
```diff
@@ -0,0 +1,140 @@
...
+
+	cksum := r.Checksum()
+	if _, ok := c.cksum2Id[cksum]; ok {
+		c.cksum2Id[cksum] = append(c.cksum2Id[cksum], id)
+	} else {
+		c.cksum2Id[r.Checksum()] = []uint32{id}
+	}
+}
+
+func (c *cache) GetId(r *Rule) uint32 {
```
davies, line 124: When will we need this one?

davies: I'd like to combine GetID and Put(*Rule) together, `Put()` will return the generated id.

pkg/acl/acl.go
jiefenghuang: forget to remove it.

pkg/acl/acl.go
jiefenghuang: default is nil, no need to allocate an empty slice

pkg/acl/cache.go
jiefenghuang: case: In a multi-client concurrent situation, Client A initializes with IDs 1-10. Client B sets ACLs for IDs 11-20. Next, Client A needs to fetch the missing ACLs for IDs 11-20.

pkg/acl/cache.go
jiefenghuang: In a multi-client concurrent environment, ID increments cannot be local; they must be managed in the meta store to ensure atomicity.

GetId is needed, for client B may have id 20, but not in client A.

davies: I see.

pkg/acl/cache.go
davies: Client can fetch missed one by one when need

pkg/acl/acl.go
davies: It's better to remove the `if`, less branches is better.

pkg/acl/acl.go
davies: It we don't sort them for checksum, then it's useless here.

It's OK to don't sort them at all.

pkg/acl/cache.go
jiefenghuang: During ACL insertion into metadata, the client can only rely on local cache to determine the existence of ACL values, which may result in duplicate ACL values. Therefore, I currently batch read to the local cache during ACL insertion if the ACL is missing, reducing the likelihood of duplicate ACLs in the future. Since insertion should be a low-frequency operation, this overhead is acceptable

pkg/acl/acl.go
```diff
@@ -0,0 +1,211 @@
...
+package acl
+
+import (
+	"fmt"
+	"hash/crc32"
+
+	"github.com/juicedata/juicefs/pkg/utils"
+)
+
+const Version uint8 = 2
```
davies, line 26: Where is it used?

PR #4440, opened 2024-03-01
cmd/sync: print skipped bytes

menwenjun:
When transferring files with the `juicefs sync` command, failures often occur, so the `--update` option is specified for the retransfer. When this option is used, a large number of files are skipped, and the number of bytes skipped is printed out. This allows the caller to estimate the total number of bytes transferred by calculating the `skipped_bytes` and the `copied_bytes` when the transfer is successful.


davies (approved):

davies:
@menwenjun thanks for the PR, please sign the CLA, then we can merge it.

menwenjun:
> @menwenjun thanks for the PR, please sign the CLA, then we can merge it.

done

PR #4461, opened 2024-03-07
`mount.juicefs` support graceful upgrade

zhijian-pro:
- [x] supprot `mount.juicefs` 
- [x] supprot `mount.juicefs` is a symlink eg: `ln -s /usr/local/bin/juicefs /sbin/mount.juicefs`

### DIFF
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -529,7 +529,6 @@ func updateFstab(c *cli.Context) error {
 func mount(c *cli.Context) error {
 	setup(c, 2)
 	addr := c.Args().Get(0)
-	removePassword(addr)
 	mp := c.Args().Get(1)
 
 	// __DAEMON_STAGE env is set by the godaemon.MakeDaemon function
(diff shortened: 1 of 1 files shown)


cmd/main.go
```diff
@@ -98,15 +97,18 @@ func Main(args []string) error {
...
+	err := app.Run(options)
 	if errno, ok := err.(syscall.Errno); ok && errno == 0 {
 		err = nil
 	}
 	return err
 }
 
 func calledViaMount(args []string) bool {
-	return strings.HasSuffix(args[0], "/mount.juicefs")
+	return strings.HasSuffix(args[0], "/mount.juicefs") && args[1] != "mount"
```
davies, line 111: What if the volume is called `mount`?

zhijian-pro: `handleSysMountArgs`  cannot handle correctly `/mount.juicefs mount -d redis://127.0.0.1 /tmp/jfs`
We previously thought that `mount.juicefs` should be used without the `mount` subcommand.

zhijian-pro:
Finish with another pr #4487

PR #4481, opened 2024-03-11
dump: reduce memory when dump large directories from redis

### LINKED ISSUE #4455: Reduce memory consumption of the meta backup procedure

Refine code of meta backup procedure, so it won't load everything into memory and marshal all at once into json.

**Why is this needed**:

OOM happens when backing up ~20 million inodes in a 5GiB memory mount pod. 

We can setup a dedicated large container to do the backup, but it's not a good solution - still hoping the code can do some streaming.

**Version:** 1.1
**Meta Engine:** redis

davies:
close #4455

### DIFF
--- a/pkg/meta/redis.go
+++ b/pkg/meta/redis.go
@@ -3646,8 +3646,6 @@ var entryPool = sync.Pool{
 	},
 }
 
-const scanEntryBatch = 1000
-
 func (m *redisMeta) dumpEntries(es ...*DumpedEntry) error {
 	ctx := Background
 	var keys []string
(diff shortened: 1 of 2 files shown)


SandyXSD (approved):

pkg/meta/redis.go
```diff
@@ -3807,12 +3802,32 @@ func (m *redisMeta) dumpDir(inode Ino, tree *DumpedEntry, bw *bufio.Writer, dept
...
+		err := m.hscan(Background, m.entryKey(inode), func(keys []string) error {
+			tree.keys = append(tree.keys, keys...)
+			return nil
+		})
+		if err != nil {
+			return err
+		}
+	}
+	entries := make([]*DumpedEntry, len(tree.keys)/2)
+	tree.Entries = make(map[string]*DumpedEntry)
```
SandyXSD, line 3818: not needed any more?

PR #4491, opened 2024-03-12
meta: reduce the memory cost of open file cache

SandyXSD:


davies (approved):

pkg/meta/openfile.go
```diff
@@ -10,14 +10,37 @@ const (
...
 	refs      int
-	lastCheck time.Time
+	lastCheck int64
+	first     []Slice
 	chunks    map[uint32][]Slice
 }
 
+func (o *openFile) invalidateChunk() {
+	o.first = nil
+	for c := range o.chunks {
```
jiefenghuang, line 30: use o.chunks = nil?

jiefenghuang: map will shrink after runtime.GC(), same as o.chunks = nil.

jiefenghuang: Correction: Earlier, I directly tested openFiles, the reduced memory should be from other objects; validating the map directly, map won't shrink after clear by deleting all elements. 

The risk here might be the map will only grow larger. Assuming a large file was opened in the past, the map will not decrease in size even if reused by smaller files.

```golang
func printMemStats(mag string) {
	var m runtime.MemStats
	runtime.ReadMemStats(&m)
	fmt.Printf("%v：memory = %vKB, GC Times = %v\n", mag, m.Alloc/1024, m.NumGC)
}

func TestMap(t *testing.T) {
	v := struct{}{}

	a := make(map[int]struct{})
	printMemStats("before")

	nums := 1000000
	for i := 0; i < nums; i++ {
		a[i] = v
	}

	runtime.GC()
	printMemStats("After Map Add 100000")
	t.Log("len of map: ", len(a))

	for i := 0; i < nums; i++ {
		delete(a, i)
	}

	runtime.GC()
	printMemStats("After Map Delete 9999")
	t.Log("len of map: ", len(a))

	for i := 0; i < 1000; i++ {
		a[i] = v
	}

	runtime.GC()
	printMemStats("After Map Add 9999 again")
	t.Log("len of map: ", len(a))
}
```

davies: Yes, go map can't shrink.

PR #4506, opened 2024-03-14
log duration of flush buffered data

zhijian-pro:


pkg/vfs/vfs.go
```diff
@@ -1214,8 +1214,12 @@ func (v *VFS) cleanupModified() {
 	}
 }
 
-func (v *VFS) FlushAll(path string) error {
-	err := v.writer.FlushAll()
+func (v *VFS) FlushAll(path string) (err error) {
+	now := time.Now()
+	defer func() {
+		logger.Infof("flush buffered data to %s duration %s error %s", path, time.Since(now), err)
```
davies, line 1220: two much logging

zhijian-pro: It will only print when upgrading, and upgrading is a low-frequency operation.
Knowing the duration of flush is very helpful for troubleshooting  issues.

PR #4507, opened 2024-03-14
gateway: add open flag when downloading an object

SandyXSD:


pkg/gateway/gateway.go
```diff
@@ -479,7 +479,7 @@ func (n *jfsObjects) GetObjectNInfo(ctx context.Context, bucket, object string,
 	if err != nil {
 		return
 	}
-	f, eno := n.fs.Open(mctx, n.path(bucket, object), 0)
+	f, eno := n.fs.Open(mctx, n.path(bucket, object), vfs.MODE_MASK_R)
```
davies, line 482: also cmd/object.go

PR #4522, opened 2024-03-15
fix: config acl

jiefenghuang:


cmd/config.go
```diff
@@ -243,10 +243,15 @@ func config(ctx *cli.Context) error {
...
-				format.EnableACL = true
-				format.MinClientVersion = "1.2.0-A"
+			if enableACL := ctx.Bool(flag); enableACL != format.EnableACL {
+				if enableACL {
+					msg.WriteString(fmt.Sprintf("%s: %v -> %v\n", flag, format.EnableACL, true))
+					msg.WriteString(fmt.Sprintf("%s: %s -> %v\n", "min-client-version", format.MinClientVersion, "1.2.0-A"))
+					format.EnableACL = true
+					format.MinClientVersion = "1.2.0-A"
+				} else {
+					fmt.Println("cannot disable acl")
```
SandyXSD, line 253: better return an error.

cmd/config.go
```diff
@@ -243,10 +243,15 @@ func config(ctx *cli.Context) error {
...
 			}
 		case "enable-acl":
-			// cannot disable
-			if ctx.Bool(flag) {
-				format.EnableACL = true
-				format.MinClientVersion = "1.2.0-A"
+			if enableACL := ctx.Bool(flag); enableACL != format.EnableACL {
+				if enableACL {
+					msg.WriteString(fmt.Sprintf("%s: %v -> %v\n", flag, format.EnableACL, true))
+					msg.WriteString(fmt.Sprintf("%s: %s -> %v\n", "min-client-version", format.MinClientVersion, "1.2.0-A"))
```
SandyXSD, line 249:
```suggestion
					msg.WriteString(fmt.Sprintf("%s: %s -> %s\n", "min-client-version", format.MinClientVersion, "1.2.0-A"))
```

jiefenghuang: done

cmd/config.go
jiefenghuang: i think this is enough, print msg and nothing changed.

image

zhoucheng361 (approved):

PR #4525, opened 2024-03-15
meta/redis: change scan threshold based on number of files

SandyXSD:


pkg/meta/redis.go
```diff
@@ -1002,7 +1002,7 @@ func (m *redisMeta) Truncate(ctx Context, inode Ino, flags uint8, length uint64,
 		if left > right {
 			right, left = left, right
 		}
-		if (right-left)/ChunkSize >= 10000 {
```
davies, line 1005: 30 -> 1000

davies: How about redis cluster? 

number of keys is better than number of inodes

SandyXSD: From: https://redis.io/docs/reference/cluster-spec/
> This means that if all keys that can match a pattern must belong to a specific slot, only this slot is searched for keys matching the pattern. The pattern slot optimization is introduced in Redis 8.0.

So since Redis 8.0 the scan will search only one slot, while in older versions it will scan all slots.

SandyXSD:
No simple way to make the decision, just leave it for now.

PR #4526, opened 2024-03-15
doc: add acl user guide

### LINKED ISSUE #4409: POSIX ACL

jiefenghuang:
close #4409


docs/en/security/posix_acl.md
```diff
@@ -0,0 +1,52 @@
...
+:::
+
+## Others
+
+- Enabling ACL increases the minimum client version requirement to v1.2.
+- Enabling ACL may have additional performance implications.
+For scenarios with infrequent ACL changes,
+the impact is minimal with memory cache optimization.
+- Enabling ACL will activate extended attributes (xattr) functionality.
+- Enabling ACL is recommended for using ["Sync Accounts between Multiple Hosts"](../administration/sync_accounts_between_multiple_hosts.md)
```
davies, line 52: 这两个不相关

docs/en/security/posix_acl.md
```diff
@@ -0,0 +1,52 @@
...
+---
+
+# POSIX ACL
+
+Version 1.2 supports POSIX ACL. For detailed rules, please refer to:
+
+- [POSIX Access Control Lists on Linux](https://www.usenix.org/legacy/publications/library/proceedings/usenix03/tech/freenix03/full_papers/gruenbacher/gruenbacher_html/main.html#:~:text=Access%20Check%20Algorithm&text=The%20ACL%20entries%20are%20looked,matching%20entry%20contains%20sufficient%20permissions.)
+- [setfacl](https://linux.die.net/man/1/setfacl)
+
+## Usage
```
davies, line 12: Add an example to enable ACL for volume created by old versions.

how to enable it:
1. upgrade all client to 1.2 and remount it
2. juicefs config --enable cli

docs/zh_cn/security/posix_acl.md
```diff
@@ -0,0 +1,44 @@
+---
+sidebar_position: 3
+---
+# POSIX ACL
+
+1.2 版本支持 POSIX ACL, 详细规则参考
+
+- [POSIX Access Control Lists on Linux](https://www.usenix.org/legacy/publications/library/proceedings/usenix03/tech/freenix03/full_papers/gruenbacher/gruenbacher_html/main.html#:~:text=Access%20Check%20Algorithm&text=The%20ACL%20entries%20are%20looked,matching%20entry%20contains%20sufficient%20permissions.)
+- [setfacl](https://linux.die.net/man/1/setfacl)
```
davies, line 9: 有没有这个文档的中文版？

docs/en/security/posix_acl.md
jiefenghuang: removed

docs/zh_cn/security/posix_acl.md
jiefenghuang: 这个没有

docs/en/security/posix_acl.md
jiefenghuang: done

PR #4543, opened 2024-03-20
feat: disk cache switch to consistent hash and add bad disk detection

### LINKED ISSUE #4540: Enhance disk cache: consistent hash
- add consistent hash support
- add disk cache uuid
- add bad disk detection
  - set free ratio to no less than 20% if in root volume

**Why is this needed**:
- consistent hashing supports the addition and removal of disk cache.
- disk cache's uuid allows for disk cache migration without affecting the results of consistent hashing.

jiefenghuang:
close #4540

### DIFF
--- a/pkg/chunk/disk_cache.go
+++ b/pkg/chunk/disk_cache.go
@@ -26,21 +26,29 @@ import (
 	"math"
 	"os"
 	"path/filepath"
+	"reflect"
 	"regexp"
 	"runtime"
 	"sort"
 	"strings"
 	"sync"
 	"sync/atomic"
+	"syscall"
 	"time"
 
+	"github.com/davies/groupcache/consistenthash"
+	"github.com/google/uuid"
 	"github.com/juicedata/juicefs/pkg/utils"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/twmb/murmur3"
 )
 
 var (
-	stagingDir = "rawstaging"
-	cacheDir   = "raw"
+	stagingDir   = "rawstaging"
+	cacheDir     = "raw"
+	maxIODur     = time.Second * 5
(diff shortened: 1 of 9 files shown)


pkg/chunk/disk_cache.go
```diff
@@ -26,21 +26,29 @@ import (
...
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/twmb/murmur3"
 )
 
 var (
-	stagingDir = "rawstaging"
-	cacheDir   = "raw"
+	stagingDir   = "rawstaging"
+	cacheDir     = "raw"
+	maxIODur     = time.Second * 5
```
davies, line 49: 60 seconds

pkg/chunk/disk_cache.go
```diff
@@ -797,46 +990,121 @@ func newCacheManager(config *Config, reg prometheus.Registerer, uploader func(ke
...
+		dir = s.dir
+	}
+	delete(m.storeMap, id)
+	for i, c := range m.stores {
+		if c != nil && c.id == id {
+			m.stores[i] = nil
+		}
+	}
+	m.Unlock()
+	logger.Infof("cache dir `%s`(%s) is unavailable, removed", dir, id)
```
davies, line 1056: error

pkg/chunk/disk_cache.go
jiefenghuang: ok, read/write on hdd bad sector may hang for 30s.

may be we should retry for 3 ~ 5 times ?

jiefenghuang: change io timeout to 60s

pkg/chunk/disk_cache.go
jiefenghuang: done

PR #4558, opened 2024-03-22
cmd: support human friendly values for options

SandyXSD:


cmd/flags.go
```diff
@@ -372,3 +373,40 @@ func duration(s string) time.Duration {
...
 	return d + time.Hour*time.Duration(v*24)
 }
+
+func parseBytes(ctx *cli.Context, key string, unit byte) uint64 {
+	str := ctx.String(key)
+	if len(str) == 0 {
+		return 0
+	}
+
+	s := str
```
jiefenghuang, line 383: use humanize.ParseBytes(ctc.String(key)) ?

SandyXSD: For simplicity, I want to use a single char as the unit, like 'M' or 'G'. While in the humanize package, you have to write 'mi' or 'gi'.

jiefenghuang: docker/go-units RAMInBytes support that
https://github.com/docker/go-units/blob/16e18b2861ca6fd622e7042ffeb9a3ebe8a9dff9/size.go#L86

Issue #4570, opened 2024-03-25
nlink not correct after dump->load->dump


zhoucheng361:
**What happened**:
```
     META_URL=sqlit3://test.db
    ./juicefs format $META_URL myjfs --trash-days 1 --enable-acl
    ./juicefs mount -d $META_URL /jfs --enable-xattr
    SEED=$SEED MAX_EXAMPLE=100 STEP_COUNT=50 PROFILE=generate ROOT_DIR1=/jfs/fsrand ROOT_DIR2=/tmp/fsrand python3 .github/scripts/hypo/fsrand2.py || true
    ./juicefs dump $META_URL dump.json --fast
    ./juicefs load sqlite3://test2.db dump.json
    ./juicefs dump sqlite3://test2.db dump2.json --fast

root@bench-01:~/juicefs# diff dump.json dump2.json
1020c1020
<     "attr": {"inode":0,"type":"directory","mode":365,"uid":0,"gid":0,"atime":1711344962,"mtime":1711344962,"ctime":1711344962,"atimensec":617063975,"mtimensec":617063975,"ctimensec":617063975,"nlink":2,"length":0},
---
>     "attr": {"inode":0,"type":"directory","mode":365,"uid":0,"gid":0,"atime":1711344962,"mtime":1711344962,"ctime":1711344962,"atimensec":617063975,"mtimensec":617063975,"ctimensec":617063975,"nlink":3,"length":0},
```

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**

**Environment**:
- JuiceFS version (use `juicefs --version`) or Hadoop Java SDK version:
- Cloud provider or hardware configuration running JuiceFS:
- OS (e.g `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Object storage (cloud provider and region, or self maintained):
- Metadata engine info (version, cloud provider managed or self maintained):
- Network connectivity (JuiceFS to metadata engine, JuiceFS to object storage):
- Others:

davies:
Why the inode is 0?

zhoucheng361:
> Why the inode is 0?

inode is set to 0 manually.

zhoucheng361:
fixed by https://github.com/juicedata/juicefs/pull/4572

PR #4572, opened 2024-03-25
fix nlink of loaded trash directory

zhijian-pro:


zhoucheng361:
fix https://github.com/juicedata/juicefs/issues/4570

PR #4577, opened 2024-03-26
assign default values to the trash node

zhijian-pro:
If the trash node does not exist, give it a default value.


davies:
also sql and redis?

pkg/meta/tkv.go
```diff
@@ -3323,6 +3323,16 @@ func (m *kvMeta) DumpMeta(w io.Writer, root Ino, keepSecret, fast, skipTrash boo
 		tree = m.snap[root]
 		if !skipTrash {
 			trash = m.snap[TrashInode]
+			if trash == nil {
```
davies, line 3326: When this happen? It's OK to have no trash.

zhijian-pro: Uniform behavior is for easier testing.

zhijian-pro:
> also sql and redis?

Other engines don't have this problem.

PR #4582, opened 2024-03-26
add timeout for range get and delete

davies:


pkg/chunk/cached_store.go
```diff
@@ -665,33 +668,38 @@ func (store *cachedStore) load(key string, page *Page, cache bool, forceCache bo
...
+		defer page.Release()
+		// it will be retried outside
+		for err != nil && tried < 2 {
+			time.Sleep(time.Second * time.Duration(tried*tried))
+			if tried > 0 {
+				logger.Warnf("GET %s: %s; retrying", key, err)
+				store.objectReqErrors.Add(1)
+				start = time.Now()
+			}
+			in, err = store.storage.Get(key, 0, -1)
```
SandyXSD, line 684: only wrap the `Get` method with `GetTimeout`?

pkg/chunk/cached_store.go
```diff
@@ -155,11 +155,16 @@ func (s *rSlice) ReadAt(ctx context.Context, page *Page, off int) (n int, err er
...
 		// partial read
 		st := time.Now()
-		in, err := s.store.storage.Get(key, int64(boff), int64(len(p)))
-		if err == nil {
-			n, err = io.ReadFull(in, p)
-			_ = in.Close()
-		}
+		page.Acquire()
+		err := utils.WithTimeout(func() error {
+			defer page.Release()
```
polyrabbit, line 160: Why we need acquire/release, since this is a synchronized call?

SandyXSD: Without the acquire the function may access released page, since it's not stopped even after timeout.

pkg/chunk/cached_store.go
davies: Get() return reader, it may hang inside Read()

pkg/chunk/cached_store.go
polyrabbit: OK I see, `utils.WithTimeout` calls func asynchronously.

pkg/chunk/cached_store.go
```diff
@@ -665,33 +668,38 @@ func (store *cachedStore) load(key string, page *Page, cache bool, forceCache bo
...
-		_ = in.Close()
-	}
-	if compressed && err == io.ErrUnexpectedEOF {
-		err = nil
-	}
+		if err == nil {
+			if compressed {
+				c := NewOffPage(needed)
+				defer c.Release()
+				buf = c.Data
```
SandyXSD, line 691: The `buf` will be released before used.

davies: fixed

SandyXSD (approved):

pkg/chunk/cached_store.go
```diff
@@ -155,11 +155,16 @@ func (s *rSlice) ReadAt(ctx context.Context, page *Page, off int) (n int, err er
...
 		st := time.Now()
-		in, err := s.store.storage.Get(key, int64(boff), int64(len(p)))
-		if err == nil {
-			n, err = io.ReadFull(in, p)
-			_ = in.Close()
-		}
+		page.Acquire()
+		err := utils.WithTimeout(func() error {
+			defer page.Release()
+			in, err := s.store.storage.Get(key, int64(boff), int64(len(p)))
```
jiefenghuang, line 161: goroutines that started by utils.WithTimeout will not be interrupted and continue to execute or block (which is not the desired outcome). and may cause the number of goroutines to keep increasing? maybe we need pass context to object storage interface and cancel it, like:
```golang
type ObjectStorage interface {
	Get(ctx context, key string, off, limit int64, getters ...AttrGetter) (io.ReadCloser, error)
}

// i.e. pass the context instead of context.Background()
func (s *s3client) Get(ctx context, key string, off, limit int64, getters ...AttrGetter) (io.ReadCloser, error) {
     ....
     resp, err := s.s3.GetObjectWithContext(ctx, params, request.WithGetResponseHeader(s3RequestIDKey, &reqID))
}
```

davies: we can do this later

polyrabbit: Agreed, we can do our best effort to support object clients that accept a ctx.

jiefenghuang: issue #4642

polyrabbit: When ctx parameter is supported, I think we can pass `getters ...AttrGetter` via `context.WithValue`

PR #4590, opened 2024-03-27
refactor: add version check when refresh metadata

jiefenghuang:


pkg/meta/error.go
```diff
@@ -0,0 +1,9 @@
+package meta
```
SandyXSD, line 1: move it to `utils/errors`.

pkg/meta/base.go
```diff
@@ -573,15 +573,16 @@ func (m *baseMeta) refresh() {
...
-		if format, err := m.Load(false); err != nil {
+		if format, err := m.Load(true); err != nil {
+			logger.Warnf("reload setting: %s", err)
+
+			if errors.Is(err, ErrMetadataVer) {
+				os.Exit(UmountCode)
+			}
+
 			if strings.HasPrefix(err.Error(), "database is not formatted") {
-				logger.Errorf("reload setting: %s", err)
```
SandyXSD, line 578: We do need an outstanding error message to tell users why the mount process exits.

davies (approved):

PR #4605, opened 2024-03-28
Add `fuse_written_file_size_bytes` metric to better understand user's usage pattern

polyrabbit:
Together with `fuse_read_size_bytes` and `fuse_written_size_bytes` metrics, we can better understand user's IO pattern.

### DIFF
--- a/pkg/vfs/writer.go
+++ b/pkg/vfs/writer.go
@@ -404,7 +404,11 @@ func (f *fileWriter) Flush(ctx meta.Context) syscall.Errno {
 
 func (f *fileWriter) Close(ctx meta.Context) syscall.Errno {
 	defer f.w.free(f)
-	return f.Flush(ctx)
+	eno := f.Flush(ctx)
+	if eno == 0 {
+		fileSizeHistogram.Observe(float64(f.GetLength()))
+	}
+	return eno
 }
 
 func (f *fileWriter) GetLength() uint64 {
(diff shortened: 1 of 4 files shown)


pkg/vfs/writer.go
```diff
@@ -404,7 +404,11 @@ func (f *fileWriter) Flush(ctx meta.Context) syscall.Errno {
 
 func (f *fileWriter) Close(ctx meta.Context) syscall.Errno {
 	defer f.w.free(f)
-	return f.Flush(ctx)
+	eno := f.Flush(ctx)
+	if eno == 0 {
+		fileSizeHistogram.Observe(float64(f.GetLength()))
```
davies, line 409: The size of written file will be difficult to understand, it also increase the number of metrics a lot (41), I'd like to not have these.

polyrabbit: * For the bucket number - it's OK to cut it to a smaller number (eg. 24).
* For the metric - it tells us size distribution of user's files, allowing us to understand whether we are dealing with large or small files. Or any other suggestions to get this distribution?
image

davies: you can know the avg size by total usage and inodes. For a given time period, you can know added files and added space. You can also find out the written data and write ops of metadata to find out the average size of chunk.

polyrabbit: OK, got it, will close.

PR #4609, opened 2024-03-29
Expose the storage-class to the metrics of throughput and latency

polyrabbit:
Some object storage supports configuring lifecycles to automatically move objects into different storage classes. This label allows us to observe the throughput and latency of different storage classes.

### DIFF
--- a/pkg/object/storage_class.go
+++ b/pkg/object/storage_class.go
@@ -1,0 +1,53 @@
+/*
+ * JuiceFS, Copyright 2024 Juicedata, Inc.
+ *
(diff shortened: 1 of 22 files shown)


pkg/object/storage_class.go
```diff
@@ -0,0 +1,37 @@
+package object
```
davies, line 1: license header

davies:
lgtm, please add unit test for it

polyrabbit:
@davies Done, please take a look

Issue #4611, opened 2024-03-29
``--custom-lables`` does not work with the juicefs csi driver


kyungwan-nam:
**What happened**:

``--custom-lables`` introduced in https://github.com/juicedata/juicefs/pull/4312 does not work with the juicefs csi driver

storageclass

```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
...
mountOptions:
- ${.PVC.annotations.csi.juicefs.com/additional-mount-options}
```

pvc

```yaml
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: myclaim
  namespace: kwnam
  annotations:
    csi.juicefs.com/additional-mount-options: "custom-labels=namespace:kwnam,name:myclaim"
```

the mount-pod failed as below.

```
2024/03/27 20:55:55.683602 juicefs[57] <INFO>: Prometheus metrics listening on [::]:9567 [mount.go:135]
2024/03/27 20:55:55.683943 juicefs[57] <INFO>: Mounting volume jfs at /jfs/pvc-xxxxxxxxxxxxxxx ... [mount_unix.go:763]
/usr/bin/fusermount: unknown option 'name:myclaim'
2024-03-27 20:55:55.685112 I | mount: fusermount exited with code 256
```

``,`` is the delimiter for each option. https://github.com/juicedata/juicefs/blob/main/cmd/main.go#L151
It should be changed to another character instead of ``,``

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**

**Environment**:
- JuiceFS version (use `juicefs --version`) or Hadoop Java SDK version:
- Cloud provider or hardware configuration running JuiceFS:
- OS (e.g `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Object storage (cloud provider and region, or self maintained):
- Metadata engine info (version, cloud provider managed or self maintained):
- Network connectivity (JuiceFS to metadata engine, JuiceFS to object storage):
- Others:

kyungwan-nam:
PR: https://github.com/juicedata/juicefs/pull/4612

zhijian-pro:
done #4612

kyungwan-nam:
@zhijian-pro Thanks!

PR #4612, opened 2024-03-29
Delimiter for --custom-labels need to change

kyungwan-nam:
changed the delimiter to ``;`` instead of ``,``


zhijian-pro:
@kyungwan-nam  What is the reason for doing this?

kyungwan-nam:
@zhijian-pro 
https://github.com/juicedata/juicefs/issues/4611 is the reason why It should be fixed.
Thanks!

PR #4613, opened 2024-03-29
update grafana dashboard template

yuhr123:
~Please pay attention to the variables and names in the template.~
Please help review the variable names in the template.

### DIFF
--- a/docs/en/grafana_template.json
+++ b/docs/en/grafana_template.json
@@ -1,9 +1,43 @@
 {
+  "__inputs": [
+    {
+      "name": "DS_GRAFANACLOUD-JUICEFS-PROM",
+      "label": "grafanacloud-juicefs-prom",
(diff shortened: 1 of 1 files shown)


davies:
What's the requirement for Grafana?

yuhr123:
> What's the requirement for Grafana?

image

As shown in the image, when uploading a template on the Grafana Dashboard platform, it displays this error message. I didn't find a clear syntax explanation based on the link provided on the page. 

However, I was able to upload the template to the public repository after rebuilding and exporting it on the latest version of Grafana.

It is possible that the reason is that our current template was exported based on an older version of Grafana.

davies:
User may still use older version of Grafana, they should be able to use this template.

yuhr123:
For compatibility considerations, we will not overwrite the existing template files for now.

PR #4619, opened 2024-03-30
Refactor: unified way to get request_id and storage_class from different object requests

polyrabbit:
### Limitation of current implementation to get request id:
* The periodical cleaner may delete request id before it's logged, especially when we are having a slow request. More often than not, we see logs with request id missing:
  > GET chunks/0/4/4099_8_4194304 (, 1.204s) [cached_store.go:651]

* Lock is needed when getting and setting a request id. Synchronization should be avoided in the critical IO path.

### New implementation:
Provides a unified way to get request_id and [storage_class](https://github.com/juicedata/juicefs/pull/4609) from different object clients, as they are all attributes of a request. It is also extensible to other attributes on future needs.

#### Disadvantages:
Changed interface of `Get/Put/Delete` methods by adding an optional parameter. But it shouldn't break existing callers with the optional pattern.


pkg/object/response_attrs_test.go
```diff
@@ -0,0 +1,28 @@
+package object
```
davies, line 1: license header

polyrabbit: Fixed

PR #4628, opened 2024-04-01
Aggressively flush slice when buffer is almost full

polyrabbit:
There are two conditions that can trigger a write stall:
1. When there are too many slices: `f.totalSlices() >= 1000`
2. When memory usage is too high: `f.w.usedBufferSize() > f.w.bufferSize`

However, the `flushAll` method currently only considers the number of slices when deciding to trigger an aggressive flush. It does not take into account the memory usage. This can lead to a situation where write stalls occur frequently due to high memory usage, even though the backend uploading threads are idle. This is because the `flushAll` method does not trigger a flush based on memory usage, resulting in a lack of progress and frequent write stalls.

image

After applying this patch, we seldom see write p99 exceeds 10ms:
image

### DIFF
--- a/pkg/vfs/writer.go
+++ b/pkg/vfs/writer.go
@@ -451,21 +451,35 @@ func (w *dataWriter) flushAll() {
 	for {
 		w.Lock()
 		now := time.Now()
+		var timeTriggered, sliceTriggered, bufferTriggered int
 		for _, f := range w.files {
 			f.refs++
 			w.Unlock()
-			tooMany := f.totalSlices() > 800
-			f.Lock()
-
+			tooManySlices := f.totalSlices() > 800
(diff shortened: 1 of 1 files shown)


pkg/vfs/writer.go
```diff
@@ -294,14 +294,17 @@ func (f *fileWriter) Write(ctx meta.Context, off uint64, data []byte) syscall.Er
 		if f.totalSlices() < 1000 {
 			break
 		}
+		logger.Warnf("Too many pending slices of inode %d, pendings: %d, will wait for 1ms", f.inode, f.totalSlices())
```
davies, line 297: there will be too much logging

polyrabbit: OK, removed both logging. My point was - we need to tell user a write stall is happening now.

PR #4640, opened 2024-04-03
refactor on ACL

davies:


pkg/acl/cache.go
```diff
@@ -148,29 +125,18 @@ func (c *cache) put(id uint32, r *Rule) {
 		return
 	}
 
-	sort.Sort(&c.id2Rule[id].NamedUsers)
```
jiefenghuang, line 151: sorting is needed to ensure equality comparison, such as with IsEqual or calculating a checksum.

acls come from dumped json or kernel maybe unsorted.

davies: It's better to preserve the order from user. It's OK to have some duplicated record when the order is different

Since we use List in dumped JSON, that will be OK.

pkg/meta/sql.go
```diff
@@ -982,18 +982,18 @@ func (m *dbMeta) doGetAttr(ctx Context, inode Ino, attr *Attr) syscall.Errno {
...
-			if err != nil {
-				return err
-			}
-			attr.Mode = (rule.GetMode() & 0777) | (attr.Mode & 07000)
-		}
 		return nil
 	}))
 }
 
+func dupAcl(acl *aclAPI.Rule) *aclAPI.Rule {
```
jiefenghuang, line 989: move to acl.go, func (r *Rule) Dup() *Rule {...}?
and deep copy named entries
```golang
dup := &Rule{
		Owner:       acl.Owner,
		Group:       acl.Group,
		Mask:        acl.Mask,
		Other:       acl.Other,
		NamedUsers:  append(Entries{}, acl.NamedUsers...),
		NamedGroups: append(Entries{}, acl.NamedGroups...),
	}
```

jiefenghuang (approved):

PR #4655, opened 2024-04-07
Hadoop: support conf val with unit

tangyoupeng:


pkg/utils/humanize.go
```diff
@@ -88,11 +113,15 @@ func ParseMbps(ctx *cli.Context, key string) int64 {
...
 		}
 	}
+	return int64(val), err
+}
+
+func ParseMbpsStr(str string) int64 {
+	val, err := parseMbpsStr(str)
 	if err != nil {
-		logger.Fatalf("Invalid value \"%s\" for option \"--%s\"", str, key)
+		logger.Fatalf("Invalid value \"%s\"", str)
```
davies, line 122: it's better to have the option name

PR #4656, opened 2024-04-07
Add errno label to fuse ops metric

polyrabbit:
Senario:
1. Better understanding users' pattern
2. Set up alarm rules on abnormal error rate
image

### DIFF
--- a/pkg/vfs/accesslog.go
+++ b/pkg/vfs/accesslog.go
@@ -19,6 +19,7 @@ package vfs
 import (
 	"fmt"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/juicedata/juicefs/pkg/utils"
@@ -39,6 +40,10 @@ var (
 		Name: "fuse_ops_durations_seconds",
 		Help: "Operations latency in seconds.",
 	}, []string{"method"})
+	opsIOErrors = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "fuse_ops_io_errors",
+		Help: "Number of IO errors.",
+	}, []string{"errno"})
 )
 
 type logReader struct {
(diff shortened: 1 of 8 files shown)


pkg/vfs/accesslog.go
```diff
@@ -30,7 +32,7 @@ var (
 		Name:    "fuse_ops_durations_histogram_seconds",
 		Help:    "Operations latency distributions.",
 		Buckets: prometheus.ExponentialBuckets(0.0001, 1.5, 30),
-	}, []string{"method"})
+	}, []string{"method", "errno"})
```
davies, line 35: there will be too many metrics, especially for hisogram

davies: Maybe we can have a single metric for all errors.

polyrabbit: By `too many metrics`, do you mean too many metrics on the client side or tsdb side?

What do you think I move the statistic to a counter, such as:
```suggestion
	opsIOErrors = prometheus.NewCounterVec(prometheus.CounterOpts{
		Name: "fuse_ops_io_errors",
		Help: "Number of IO errors.",
	}, []string{"method", "errno"})
```

davies: both.

The combination of method and errno could be hundreds. A single counter or with single label should be enough for monitoring. For further investing, we should collect the accesslog to find out the details.

polyrabbit: Updated - added a new metric for io errors

davies: The new metric has too much labels, please pick one

polyrabbit: Done

pkg/vfs/accesslog.go
```diff
@@ -31,6 +33,10 @@ var (
 		Help:    "Operations latency distributions.",
 		Buckets: prometheus.ExponentialBuckets(0.0001, 1.5, 30),
 	}, []string{"method"})
+	opsIOErrors = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "fuse_ops_io_errors",
+		Help: "Number of IO errors.",
+	}, []string{"errno"})
```
davies, line 39: We should add the `errno` into the name of metric in `.stats`, see internal.go

polyrabbit: Do you mean change `juicefs_fuse_ops_io_errors` to `juicefs_fuse_ops_io_errors_errno` in .stats file?

davies: yes

polyrabbit: Added

PR #4658, opened 2024-04-08
Only prefetch on successful partial read

polyrabbit:
When s3 storage is unhealthy, we see lots of failed read requests for the same key - half of them are triggered by prefetch. Actually only a successful partial read needs a prefetch.

### DIFF
--- a/pkg/chunk/cached_store.go
+++ b/pkg/chunk/cached_store.go
@@ -173,11 +173,12 @@ func (s *rSlice) ReadAt(ctx context.Context, page *Page, off int) (n int, err er
 		logRequest("GET", key, fmt.Sprintf("RANGE(%d,%d) ", boff, len(p)), reqID, err, used)
 		s.store.objectDataBytes.WithLabelValues("GET", sc).Add(float64(n))
 		s.store.objectReqsHistogram.WithLabelValues("GET", sc).Observe(used.Seconds())
-		s.store.fetcher.fetch(key)
 		if err == nil {
+			s.store.fetcher.fetch(key)
 			return n, nil
 		} else {
 			s.store.objectReqErrors.Add(1)
+			// fall back to full read
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/chunk/cached_store.go
```diff
@@ -173,11 +173,12 @@ func (s *rSlice) ReadAt(ctx context.Context, page *Page, off int) (n int, err er
 		logRequest("GET", key, fmt.Sprintf("RANGE(%d,%d) ", boff, len(p)), reqID, err, used)
 		s.store.objectDataBytes.WithLabelValues("GET", sc).Add(float64(n))
 		s.store.objectReqsHistogram.WithLabelValues("GET", sc).Observe(used.Seconds())
-		s.store.fetcher.fetch(key)
 		if err == nil {
+			s.store.fetcher.fetch(key)
 			return n, nil
 		} else {
 			s.store.objectReqErrors.Add(1)
+			return n, err
```
SandyXSD, line 181: we want to try everything we can do before failure, don't return here.

pkg/vfs/reader.go
```diff
@@ -224,7 +224,7 @@ func (s *sliceReader) run() {
 		_ = f.r.m.InvalidateChunkCache(meta.Background, inode, indx)
 		if f.tried > f.r.maxRetries {
 			s.done(err, 0)
-		} else {
+		} else { // Why retry here since we already have retried in `cachedStore.load`?
```
SandyXSD, line 227: This is another level of retry. The slices obtained from the `meta.Read` method may be changed.

polyrabbit: Got it, updated

pkg/chunk/cached_store.go
polyrabbit: We need a way to distinguish retriable errors, otherwise we will have a large read amplification.

PR #4695, opened 2024-04-14
Fix: backup always runs in fast mode, consuming more than 5G memory on a large volume

polyrabbit:
It turns out the backup process always runs in fast mode. Which causes OOM

### DIFF
--- a/pkg/vfs/backup.go
+++ b/pkg/vfs/backup.go
@@ -63,8 +63,8 @@ func Backup(m meta.Meta, blob object.ObjectStorage, interval time.Duration, skip
 		}
 		if now := time.Now(); now.Sub(last) >= interval {
 			var iused, dummy uint64
+			_ = m.StatFS(ctx, meta.RootInode, &dummy, &dummy, &iused, &dummy)
 			if interval <= time.Hour {
-				_ = m.StatFS(ctx, meta.RootInode, &dummy, &dummy, &iused, &dummy)
 				if iused > 1e6 {
 					logger.Warnf("backup metadata skipped because of too many inodes: %d %s; "+
 						"you may increase `--backup-meta` to enable it again", iused, interval)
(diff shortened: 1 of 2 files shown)


pkg/vfs/backup.go
```diff
@@ -63,23 +63,23 @@ func Backup(m meta.Meta, blob object.ObjectStorage, interval time.Duration, skip
...
-				_ = m.StatFS(ctx, meta.RootInode, &dummy, &dummy, &iused, &dummy)
 				if iused > 1e6 {
 					logger.Warnf("backup metadata skipped because of too many inodes: %d %s; "+
 						"you may increase `--backup-meta` to enable it again", iused, interval)
 					continue
 				}
 			}
 			if st := m.SetXattr(ctx, 0, key, []byte(now.Format(time.RFC3339)), meta.XattrCreateOrReplace); st != 0 {
-				logger.Warnf("setxattr inode 1 key %s: %s", key, st)
+				logger.Warnf("setxattr inode 0 key %s: %s", key, st)
```
SandyXSD, line 75: It will actually set xattr on inode 1 (the root inode), see `baseMeta.checkRoot`.

polyrabbit: OK, updated. I thought it was a typo

polyrabbit:
Around 30TiB. We have larger volumes that the backup procedure runs OK, so I think inodes is a good estimation for now.

BTW, how should we use the `LastBackupDurationG` metric in a multi-client environment? It's a gauge, that once one client sets, it will always emit the same value.
In an environment with hundreds of pods, all pods are emitting their last backup duration, even most of them are obsolete. How do I get the latest backup duration?

davies:
Have you run into OOM with current setting (fast with 100K- inodes, slow with 100K-1000K inodes)?

We will fix the LastBackupDurationG to reset in every cycle, so you can plot the max of last_backup_duration

polyrabbit:
No, it runs OK on a volume of 4.9M inodes (and even faster than the `fast` mode).

Thanks!

PR #4707, opened 2024-04-16
object/ceph: add Shutdown method

SandyXSD:


pkg/object/interface.go
```diff
@@ -76,6 +76,8 @@ type Limits struct {
 type ObjectStorage interface {
 	// Description of the object storage.
 	String() string
+	// Shutdown connections with the object storage.
+	Shutdown() error
```
davies, line 80: We can have a special interface for it

PR #4710, opened 2024-04-16
Fix panic when reading and writing s.writer concurrently

polyrabbit:
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1ed3c80]

goroutine 3559509 [running]:
github.com/juicedata/juicefs/pkg/chunk.(*wSlice).ID(0x10100c00f1f1c20?)
        /home/miaochangxin/workspace/juicefs/pkg/chunk/cached_store.go:472
github.com/juicedata/juicefs/pkg/vfs.(*sliceWriter).prepareID(0xc00f330680, {0x3fbda18, 0xc0110dcf00}, 0x0)
        /home/miaochangxin/workspace/juicefs/pkg/vfs/writer.go:91 +0x282
created by github.com/juicedata/juicefs/pkg/vfs.(*sliceWriter).write in goroutine 1
        /home/miaochangxin/workspace/juicefs/pkg/vfs/writer.go:139 +0x149
```

### DIFF
--- a/pkg/vfs/writer.go
+++ b/pkg/vfs/writer.go
@@ -120,9 +120,6 @@ func (s *sliceWriter) flushData() {
 		s.writer.Abort()
 		s.err = syscall.EIO
 	}
-	s.chunk.file.Lock()
-	s.writer = nil
-	s.chunk.file.Unlock()
 }
 
 // protected by s.chunk.file
--- a/pkg/vfs/writer.go
+++ b/pkg/vfs/writer.go
@@ -120,7 +120,9 @@ func (s *sliceWriter) flushData() {
 		s.writer.Abort()
 		s.err = syscall.EIO
 	}
+	s.chunk.file.Lock()
 	s.writer = nil
+	s.chunk.file.Unlock()
 }
 
 // protected by s.chunk.file


pkg/vfs/writer.go
```diff
@@ -120,6 +120,8 @@ func (s *sliceWriter) flushData() {
 		s.writer.Abort()
 		s.err = syscall.EIO
 	}
+	s.chunk.file.Lock()
+	defer s.chunk.file.Unlock()
 	s.writer = nil
```
jiefenghuang, line 125:
```suggestion
	s.chunk.file.Lock()
	s.writer = nil
	s.chunk.file.Unlock()
```

polyrabbit: OK, updated

SandyXSD (approved):

pkg/vfs/writer.go
```diff
@@ -120,7 +120,9 @@ func (s *sliceWriter) flushData() {
 		s.writer.Abort()
 		s.err = syscall.EIO
 	}
+	s.chunk.file.Lock()
 	s.writer = nil
```
davies, line 124: it's better to remove this line

davies:
This could be introduced by https://github.com/juicedata/juicefs/pull/4694, which will start multiple gorotuine to prepare id, we should fix that also.

polyrabbit:
Before #4694, I suppose it was also possible to spawn multiple goroutines to prepare one ID?

davies:
it may spawn 2 goroutines when it crosss the 4M border.

PR #4713, opened 2024-04-17
fix: file length after CopyFileRange and add checks for meta CopyFileRange

jiefenghuang:


davies:
Please add a test to cover it

Issue #4723, opened 2024-04-17
Support ``;`` separated values for juicefs.users, juicefs.group


kyungwan-nam:
**What would you like to be added**:

``juicefs.users``, ``juicefs.group`` requires a file.
https://juicefs.com/docs/community/hadoop_java_sdk#other-configurations
I'd like to set it as ``;`` separated values instead of a file.

**Why is this needed**:

I want to let users set their ``juicefs.users``, ``juicefs.group`` in our multi-tenant Hadoop cluster.
setting ``;`` separated values looks easier for users.

kyungwan-nam:
PR: https://github.com/juicedata/juicefs/pull/4724

PR #4724, opened 2024-04-17
Support ; separated values for juicefs.users, juicefs.group

kyungwan-nam:
Support ``;`` separated values for juicefs.users, juicefs.group https://github.com/juicedata/juicefs/issues/4723


tangyoupeng:
Hi, why do we need support ";" sperated values? We design this file value according the local /etc/group @kyungwan-nam

tangyoupeng:
Could you add some doc for it in this file "hadoop_java_sdk.md"

PR #4741, opened 2024-04-19
Fix error message in quota always shows inode 0

polyrabbit:
We see lots of error messages on quota, which displays a wrong inode:

```
Get directory parent of inode 0: no such file or directory [quota.go:344]
```


jiefenghuang:
I think the problem is that the call to GetAttr fails in getDirParent, because there are other places that call getDirParent, it's best to handle it in getDirParent.

polyrabbit:
OK, also updated those two

PR #4743, opened 2024-04-21
Enable concurrent write to both stage files and object storage to get 5GB/s write throughput

polyrabbit:
Currently, the writeback mode works by writing the block to the staging files first, and then uploading them to the object storage. So the throughput of the writeback mode is limited by the speed of disk. And the max write concurrency is `buffer-size / block-size` which could be overwhelming for some disks.

By limiting concurrent writes to staged files and spill others to object backend, we can have concurrent uploads to two backends, and theoretically the write throughput equals to **sum of the two** (see following experiment).

> By setting `--max-stage-write=10`, we got a write throughput of **5GB/s**, almost double the speed of original write. And it lasts for a long time, as long as we have enough space for stage files

image

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -128,6 +128,11 @@ func storageFlags() []cli.Flag {
 			Usage: "number of connections to upload",
 		},
 		&cli.IntFlag{
+			Name:  "max-stage-write",
+			Value: 0, // Enable this to have concurrent uploads to two backends, and get write bandwidth equals to sum of the two
+			Usage: "number of threads allowed to write staged files, other requests will be uploaded directly (this option is only effective when 'writeback' mode is enabled)",
+		},
+		&cli.IntFlag{
 			Name:  "max-deletes",
 			Value: 10,
 			Usage: "number of threads to delete objects",
--- a/pkg/chunk/cached_store.go
+++ b/pkg/chunk/cached_store.go
@@ -438,8 +438,10 @@ func (s *wSlice) upload(indx int) {
 		if s.store.conf.Writeback {
 			stagingPath, err := s.store.bcache.stage(key, block.Data, s.store.shouldCache(blen))
 			if err != nil {
-				s.store.stageBlockErrors.Add(1)
-				logger.Warnf("write %s to disk: %s, upload it directly", stagingPath, err)
+				if !errors.Is(err, errStageConcurrency) {
+					s.store.stageBlockErrors.Add(1)
+					logger.Warnf("write %s to disk: %s, upload it directly", stagingPath, err)
+				}
 			} else {
 				s.errors <- nil
 				if s.store.conf.UploadDelay == 0 && s.store.canUpload() {
@@ -538,6 +540,7 @@ type Config struct {
 	AutoCreate        bool
 	Compress          string
 	MaxUpload         int
+	MaxStageWrite     int
 	MaxRetries        int
 	UploadLimit       int64 // bytes per second
 	DownloadLimit     int64 // bytes per second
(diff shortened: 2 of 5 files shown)


jiefenghuang:
This modification may impact the performance in small file scenarios, which are the main focus of write-back optimization.

polyrabbit:
Then, we expanded `write-back` scenarios :)

cmd/flags.go
```diff
@@ -127,6 +127,11 @@ func storageFlags() []cli.Flag {
 			Value: 20,
 			Usage: "number of connections to upload",
 		},
+		&cli.IntFlag{
+			Name:  "max-stage-write",
+			Value: 0, // Enable this to have concurrent uploads to two backends, and get write bandwidth equals to sum of the two
```
SandyXSD, line 132: Maybe we can enable it by default, with a value 10 or 20.

cmd/flags.go
```diff
@@ -127,6 +127,11 @@ func storageFlags() []cli.Flag {
...
+			Usage: "number of threads allowed to write staged files, other requests will be uploaded directly",
```
SandyXSD, line 133: Better to emphasize that this option is only effective in writeback mode.

pkg/chunk/metrics.go
```diff
@@ -25,6 +29,7 @@ type cacheManagerMetrics struct {
 	cacheEvicts     prometheus.Counter
 	cacheWriteBytes prometheus.Counter
 	cacheWriteHist  prometheus.Histogram
+	stagingBlocks   atomic.Int64
```
SandyXSD, line 32: Better use a type in prometheus as well.

pkg/chunk/cached_store.go
```diff
@@ -434,8 +435,12 @@ func (s *wSlice) upload(indx int) {
 		if s.store.conf.Writeback {
 			stagingPath, err := s.store.bcache.stage(key, block.Data, s.store.shouldCache(blen))
 			if err != nil {
-				s.store.stageBlockErrors.Add(1)
-				logger.Warnf("write %s to disk: %s, upload it directly", stagingPath, err)
+				logLevel := logrus.DebugLevel
```
SandyXSD, line 438: Don't adjust log level, instead use `Warnf` and `Debugf` in different if blocks.

pkg/chunk/cached_store.go
```diff
@@ -559,10 +565,16 @@ func (c *Config) SelfCheck(uuid string) {
...
-		c.UploadDelay = 0
-		c.UploadHours = ""
+	if !c.Writeback {
+		if c.UploadDelay > 0 || c.UploadHours != "" {
+			logger.Warnf("delayed upload is disabled in non-writeback mode")
+			c.UploadDelay = 0
+			c.UploadHours = ""
+		}
+		if c.MaxStageWrite > 0 {
+			logger.Warnf("max-stage-write is disabled in non-writeback mode")
```
SandyXSD, line 575: `Debugf`, since we want the default value > 0

cmd/flags.go
polyrabbit: I'm afraid enabling this by default will break current expectation for `write-back`, is that a concern to you?

Also, it's hard to set a default value without observing `staging_writing_blocks` metric first. User may have multiple disks for staging, or only have HDD for staging, etc. Different environments require different optimal value.

pkg/chunk/metrics.go
polyrabbit: Yes, any suggestions where to put this metric? I didn't find other good places in the codebase, other than a global variable.

cmd/flags.go
polyrabbit: Updated description.

pkg/chunk/cached_store.go
polyrabbit: Updated.

pkg/chunk/cached_store.go
```diff
@@ -433,7 +433,7 @@ func (s *wSlice) upload(indx int) {
 		}
 		if s.store.conf.Writeback {
 			stagingPath, err := s.store.bcache.stage(key, block.Data, s.store.shouldCache(blen))
-			if err != nil {
+			if err != nil && !errors.Is(err, errStageConcurrency) {
```
SandyXSD, line 436: this is not correct now when err is errStageConcurrency

jiefenghuang:
> Then, we expanded `write-back` scenarios :)

ok, but I think if the write performance of large files is not satisfactory, it's a matter to consider for object storage. Allocating the cost of increasing disk cache on this side towards improving object storage bandwidth might be a more direct approach.

pkg/chunk/metrics.go
SandyXSD: yeah, looks like we have to use a global variable.

polyrabbit: If you prefer that, I can modify the code

pkg/chunk/cached_store.go
polyrabbit: Right. Sorry just had too many context switches this afternoon and lost track of the code.

pkg/chunk/metrics.go
SandyXSD: do it please.

polyrabbit: OK, updated and tested.

PR #4748, opened 2024-04-22
Delete leaked objects if it's already deleted by other goroutines

polyrabbit:
When gc and uploading happens simultaneously on the same stage file, the uploaded object will be leaked - never got a second chance to be deleted.

We observed several cases of leaded objects when `write-back` is enabled and `trash-days` set to 0.

### DIFF
--- a/pkg/chunk/cached_store.go
+++ b/pkg/chunk/cached_store.go
@@ -204,23 +204,7 @@ func (s *rSlice) ReadAt(ctx context.Context, page *Page, off int) (n int, err er
 
 func (s *rSlice) delete(indx int) error {
 	key := s.key(indx)
-	st := time.Now()
-	var reqID string
-	err := utils.WithTimeout(func() error {
-		return s.store.storage.Delete(key, object.WithRequestID(&reqID))
-	}, s.store.conf.PutTimeout)
-	used := time.Since(st)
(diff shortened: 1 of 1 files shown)


pkg/chunk/cached_store.go
```diff
@@ -965,10 +966,19 @@ func (store *cachedStore) uploadStagingFile(key string, stagingPath string) {
...
+	if !store.isPendingValid(key) {
+		block.Release()
+		logger.Debugf("Key %s is not needed, drop it", key)
+		return
+	}
 
 	store.stageBlockDelay.Add(time.Since(item.ts).Seconds())
 	if err = store.upload(key, block, nil); err == nil {
 		store.bcache.uploaded(key, blen)
+		if !store.isPendingValid(key) { // Delete leaked objects if it's already deleted by other goroutines
```
SandyXSD, line 978: Skip other functions (`uploaded`, `removePending`, `removeStage`) when isPendingValid = false.

polyrabbit: Is it safe? I suppose at least we still need `removePending`?

SandyXSD: pending not valid means it's already removed from the pending dict

polyrabbit: Right, updated

PR #4781, opened 2024-04-25
Actively flush dirty pages when under memory pressure to avoid wrong OOM killer from cgroupv1

polyrabbit:
Purpose:
1. Ensure staged/cached files is persisted in disk, close doesnot guarantee this - see https://linux.die.net/man/2/close
2. Avoid massive dirty pages in memory, which may cause OOM - see https://github.com/kubernetes/kubernetes/issues/43916
img_v3_02a9_f5296c41-d455-4796-8a79-7f8191b350ag

### DIFF
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -373,6 +373,7 @@ func initBackgroundTasks(c *cli.Context, vfsConf *vfs.Config, metaConf *meta.Con
 	if !c.Bool("no-usage-report") {
 		go usage.ReportUsage(m, version.Version())
 	}
+	go utils.FlushDirtyPages()
 }
 
 type storageHolder struct {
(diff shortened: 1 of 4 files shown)


davies:
This will slow down writeback especially with HDD or virtual disk.

polyrabbit:
Right, this will hurt HDD performance a lot. How about using an environment variable as a switch? for some cases, user can trade off performance for stability/reliability.

jiefenghuang:
"Avoid massive dirty pages in memory, which may cause OOM"
It might be more suitable to place this functionality in a separate script dedicated to operations platform.
The operation differ in OS distributions,  [for example](https://docs.redhat.com/en/documentation/Red_Hat_Enterprise_Linux/5/html/tuning_and_optimizing_red_hat_enterprise_linux_for_oracle_9i_and_10g_databases/sect-oracle_9i_and_10g_tuning_guide-memory_usage_and_page_cache-tuning_the_page_cache), you can directly set the page cache ratio.

PR #4783, opened 2024-04-25
Short-circuit fileReader.visit function if visitor returns false

polyrabbit:
Avoid unnecessary loop


davies:
Have your saw actual performance difference?

polyrabbit:
Not yet. I didn't intentionally construct test cases for this scenario.
Just suppose it's a common practice in visitor pattern when reading the code.

PR #4790, opened 2024-04-26
style(acl): change config name to Acl

jiefenghuang:
use same config name
```shell
root@bench-01:~/juicefs# grep -i acl /tmp/jfs/.config
  "EnableACL": true
  "EnableAcl": true,
```

### DIFF
--- a/pkg/meta/config.go
+++ b/pkg/meta/config.go
@@ -94,7 +94,7 @@ type Format struct {
 	MinClientVersion string `json:",omitempty"`
 	MaxClientVersion string `json:",omitempty"`
 	DirStats         bool   `json:",omitempty"`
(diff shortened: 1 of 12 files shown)


pkg/meta/config.go
```diff
@@ -94,7 +94,7 @@ type Format struct {
 	MinClientVersion string `json:",omitempty"`
 	MaxClientVersion string `json:",omitempty"`
 	DirStats         bool   `json:",omitempty"`
-	EnableACL        bool
+	EnableAcl        bool
```
davies, line 97: this is a broken change to beta1

davies:
this is a broken change, may not worth it.

PR #4794, opened 2024-04-26
fix: fix the concurrency issue with trash cleanup

CodeRusher:
The modification of the changed value must depend on the result of the tx.Set() method. 
This could lead to issues with multiple cleanup tasks running.

### DIFF
--- a/pkg/meta/redis.go
+++ b/pkg/meta/redis.go
@@ -419,8 +419,11 @@ func (m *redisMeta) setIfSmall(name string, value, diff int64) (bool, error) {
 		if old > value-diff {
 			return nil
 		} else {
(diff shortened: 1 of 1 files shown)


SandyXSD:
The `changed` will be reset at the start of each transaction.

CodeRusher:
> The `changed` will be reset at the start of each transaction.

yes，i got it！
However, this change ensures that tx.Set() must execute successfully before returning true to the caller. Concurrency does indeed occur, and it is unclear whether this is related to the error handling logic.

SandyXSD:
We'll always check the error first, so the value of `changed` has no impact if the `err != nil`.

PR #4802, opened 2024-04-28
fix: close /proc/1/mountinfo

testwill:


cmd/mount.go
```diff
@@ -458,6 +458,7 @@ func insideContainer() bool {
 	if os.IsNotExist(err) {
 		return false
 	}
+	defer mountinfo.Close()
```
davies, line 461: mountinfo is be nil

davies: yes

PR #4808, opened 2024-05-04
Remove uploaded slices if failed to check into meta

polyrabbit:
Another case to remove leaked objects automatically.

### DIFF
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -2091,6 +2091,7 @@ func (m *baseMeta) compactChunk(inode Ino, indx uint32, once, force bool) {
 		m.of.InvalidateChunk(inode, indx)
 	} else {
 		logger.Warnf("compact %d %d: %s", inode, indx, err)
+		// TODO: should also delete slice `id`?
 	}
 
 	if force {
--- a/pkg/vfs/writer.go
+++ b/pkg/vfs/writer.go
@@ -199,6 +199,9 @@ func (c *chunkWriter) commitThread() {
 			var ss = meta.Slice{Id: s.id, Size: s.length, Off: s.soff, Len: s.slen}
 			err = f.w.m.Write(meta.Background, f.inode, c.indx, s.off, ss, s.lastMod)
 			f.w.reader.Invalidate(f.inode, uint64(c.indx)*meta.ChunkSize+uint64(s.off), uint64(ss.Len))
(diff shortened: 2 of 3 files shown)


pkg/meta/base.go
```diff
@@ -2091,6 +2091,7 @@ func (m *baseMeta) compactChunk(inode Ino, indx uint32, once, force bool) {
 		m.of.InvalidateChunk(inode, indx)
 	} else {
 		logger.Warnf("compact %d %d: %s", inode, indx, err)
+		// TODO: should also delete slice `id`?
```
SandyXSD, line 2094: No, we have no idea about the real status of the chunk at this stage.

pkg/vfs/writer.go
```diff
@@ -199,6 +199,9 @@ func (c *chunkWriter) commitThread() {
 			var ss = meta.Slice{Id: s.id, Size: s.length, Off: s.soff, Len: s.slen}
 			err = f.w.m.Write(meta.Background, f.inode, c.indx, s.off, ss, s.lastMod)
 			f.w.reader.Invalidate(f.inode, uint64(c.indx)*meta.ChunkSize+uint64(s.off), uint64(ss.Len))
+			if err != 0 { // Remove uploaded slices if failed to check into meta
```
SandyXSD, line 202: The state is not determined for some errors like "timed out".

polyrabbit: Whatever the state is, we no longer track it in the meta engine. So we should delete those leaked objects?

SandyXSD: It's possible that the `Write` actually succeeded, in which case deleting the objects will corrupt the file.

polyrabbit: I suppose we should reject write requests at the beginning of `fileWriter.Write` if `f.err` is already marked. Some user will always retry regardless of the error. Failing fast can protect us from those useless retries.

SandyXSD: This is indeed worth discussing. Currently, JuiceFS tries its best to preserve data as meticulously as possible, as the consequences of data loss are often far more severe than those of data leakage.

polyrabbit: Failing fast is only a way to self-protection. I suppose it has no effect on data persistence?

PR #4810, opened 2024-05-05
dumpEntry should ignore non-exist inode

polyrabbit:


davies:
Right now, we dump it as empty one (could be deleted or corrupted)

pkg/meta/tkv.go
```diff
@@ -2744,6 +2744,7 @@ func (m *kvMeta) dumpEntry(inode Ino, e *DumpedEntry, showProgress func(totalInc
 		a := tx.get(m.inodeKey(inode))
 		if a == nil {
 			logger.Warnf("inode %d not found", inode)
+			return nil
```
SandyXSD, line 2747: If this is a directory, we still want all its children.

polyrabbit: OK, so a directory will also trigger a warning message? I see it a lot in the log

davies: so we can improved the case that file are delete after we saw the entry (during the dump)

polyrabbit:
Closing as it's by design

PR #4845, opened 2024-05-13
gateway: add rw locks to lock multiple requests within a single machine

zhijian-pro:


pkg/gateway/gateway.go
```diff
@@ -1183,19 +1184,46 @@ func (j *jfsFLock) getFlockWithTimeOut(ctx context.Context, ltype uint32, timeou
 	start := time.Now()
 	deadline := start.Add(timeout.Timeout())
 	lockStr := "write"
+
+	var getLockFunc func() bool
```
davies, line 1188: please add a unit test for jfsFlock

PR #4879, opened 2024-05-21
cmd/sync: return list objects error

zhijian-pro:


davies:
what's the change of behavior and why?

PR #4882, opened 2024-05-21
Write should fail fast if file is already in error state

polyrabbit:
Write should fail fast if file is already in error state, avoiding resource consumption caused by infinite retries from bad clients.

If this change poses a risk, we may consider it in 1.3


jiefenghuang:
There are many types of errors, and we want to try writing if possible, so this PR is temporarily closed.

Issue #4754, opened 2024-04-23
background tasks should exit when summary/clone/compact/info killed.


zhoucheng361:
**What would you like to be added**:

when user interrupted summary/clone by Ctrl+C, the corresponded tasks run in the mount process should exit. 

**Why is this needed**:
otherwise the memory will increase to a large amount until the task finish.

jiefenghuang:
related: 
#6033 for compact

PR #4888, opened 2024-05-22
enhance(cmd): quit background task when tool exit

jiefenghuang:
#4754


zhijian-pro:
Does this enhancement work for the clone command?

jiefenghuang:
> Does this enhancement work for the clone command?

meta.Clone call GetSummary too, so it work too.

zhijian-pro:
We'd better cancel the `cloneEntry`

zhijian-pro (approved):

PR #4893, opened 2024-05-23
fix: JfsObjects.CompleteMultipartUpload

prOOrc:
## Description
Increased max size of part to 5Gb in CompleteMultipartUpload

## Motivation and Context
I've caught bug when tested server side copy with rclone. Rclone has default copy cutoff [4.656Gi](https://rclone.org/s3/#s3-copy-cutoff). During CompleteMultipartUpload gateway ignores size of parts and doesn't consider globalMaxPartSize (5GiB) and copies 1Gb max of each part.

### DIFF
--- a/pkg/gateway/gateway.go
+++ b/pkg/gateway/gateway.go
@@ -1052,7 +1052,7 @@ func (n *jfsObjects) CompleteMultipartUpload(ctx context.Context, bucket, object
 	var total uint64
 	for _, part := range parts {
 		p := n.ppath(bucket, uploadID, strconv.Itoa(part.PartNumber))
-		copied, eno := n.fs.CopyFileRange(mctx, p, 0, tmp, total, 1<<30)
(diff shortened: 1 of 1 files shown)


zhijian-pro (approved):

pkg/gateway/gateway.go
```diff
@@ -1052,7 +1052,7 @@ func (n *jfsObjects) CompleteMultipartUpload(ctx context.Context, bucket, object
 	var total uint64
 	for _, part := range parts {
 		p := n.ppath(bucket, uploadID, strconv.Itoa(part.PartNumber))
-		copied, eno := n.fs.CopyFileRange(mctx, p, 0, tmp, total, 1<<30)
+		copied, eno := n.fs.CopyFileRange(mctx, p, 0, tmp, total, 1<<30*5)
```
zhijian-pro, line 1055: `5<<30`  is better.

PR #4918, opened 2024-06-04
refactor: reorderOptions support subcommands

CodingPoeta:
- keep reorder options for all flags
- if HelpFlag is set, it will be moved after the last command's flags
- allow subcommands with flags
- if deeper subcommand's flags overlap with it's ancestors, it have higher priority to catch the flag


davies:
We don't use subcommands, close it for now.

PR #4921, opened 2024-06-04
cmd/dump: support zstd compression, and adjust the compression level

SandyXSD:


cmd/dump.go
```diff
@@ -95,14 +94,15 @@ func dumpMeta(m meta.Meta, dst string, threads int, keepSecret, fast, skipTrash
 		}()
 
 		if strings.HasSuffix(dst, ".gz") {
-			zw := gzip.NewWriter(fp)
+			w, _ = gzip.NewWriterLevel(fp, gzip.BestSpeed)
+			defer func() {
+				err = errors.Join(err, w.Close())
```
jiefenghuang, line 99: w may be nil

cmd/dump.go
```diff
@@ -95,14 +94,15 @@ func dumpMeta(m meta.Meta, dst string, threads int, keepSecret, fast, skipTrash
...
+			}()
+		} else if strings.HasSuffix(dst, ".zstd") {
```
jiefenghuang, line 101: cmd/load should support zstd too?

cmd/dump.go
SandyXSD: It's always non-nil as long as the 'level' is valid.

jiefenghuang (approved):

PR #4925, opened 2024-06-05
fix(sync): Handling the case where local file paths contain a colon(:)

CodeRusher:
The issue here is determining whether a given URI is for a local file or SFTP based on whether the URI contains a colon. However, it's important to note that local file paths can indeed contain a colon.

Solution:
When the URI contains a colon, first check if the local file path exists.

### DIFF
--- a/pkg/meta/config.go
+++ b/pkg/meta/config.go
@@ -64,7 +64,7 @@ func (c *Config) SelfCheck() {
 		c.Heartbeat = time.Second
 	}
 	if c.Heartbeat > time.Minute*10 {
-		logger.Warnf("heartbeat shouldd not be greater than 10 minutes")
+		logger.Warnf("heartbeat should not be greater than 10 minutes")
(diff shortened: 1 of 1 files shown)


davies:
This is still confusing, for example, you can't use `ip:port` when there is a file or directory has the same name.

It's better to use full uri to address the confusion, for example,   file:///path/with/:

CodeRusher:
> This is still confusing, for example, you can't use `ip:port` when there is a file or directory has the same name.
> 
> It's better to use full uri to address the confusion, for example, file:///path/with/:
Yes, that is still confusing，but this modification is slightly better. Situations where local files exist in paths similar to SFTP are rare, whereas it's more common for local files to contain colons (":").

Additionally, is it possible to discontinue support for abbreviated usage of the SFTP/SSH protocol, and retain only the support for the abbreviated usage of local files?

PR #4931, opened 2024-06-07
fuse: remove MaxPages, and increase MAX_KERNEL_WRITE to 1 MiB

SandyXSD:


SandyXSD:
We hard-coded the `opt.MaxWrite = 1 << 20`, so with this PR the max write will be set to 1 MiB directly and can't be changed.

PR #4947, opened 2024-06-17
cache: dedup prefetch requests to avoid read amplification

polyrabbit:
We observed a large read amplification when user is doing an incremental but non-continuous sequential read on large amount of files. After labeling object GET requests with purpose, we found prefetching is the main cause of the read amplification.

It turns out the deduplication in `prefetcher::do` method does not work as expected especially when there is only one prefetcher (the default behavior), which caused duplicated prefetching. Instead, it's better to do deduplication before inserting into the `pending` queue.

image

### DIFF
--- a/pkg/chunk/prefetch.go
+++ b/pkg/chunk/prefetch.go
@@ -39,23 +39,23 @@ func newPrefetcher(parallel int, fetch func(string)) *prefetcher {
 
 func (p *prefetcher) do() {
 	for key := range p.pending {
+		p.op(key)
+
 		p.Lock()
-		if _, ok := p.busy[key]; !ok {
-			p.busy[key] = true
-			p.Unlock()
-
-			p.op(key)
-
-			p.Lock()
-			delete(p.busy, key)
-		}
+		delete(p.busy, key)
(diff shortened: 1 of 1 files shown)


pkg/chunk/prefetch.go
```diff
@@ -39,23 +39,23 @@ func newPrefetcher(parallel int, fetch func(string)) *prefetcher {
...
-			delete(p.busy, key)
-		}
+		p.Lock()
+		delete(p.busy, key)
 		p.Unlock()
 	}
 }
 
 func (p *prefetcher) fetch(key string) {
+	p.Lock()
```
jiefenghuang, line 51: this may be blocked and fetch should not block the read process

polyrabbit: Which line it will be blocked on?

jiefenghuang:
The unit tests in "prefetch_test.go" also need to be changed.

polyrabbit:
@jiefenghuang Sorry I missed your last message. Which part in prefetch_test.go do you think needs to be changed?

PR #4948, opened 2024-06-18
fix(meta): keep open&&mknod permission check behavior consistent with linux

201341:


pkg/meta/base.go
```diff
@@ -1354,6 +1354,10 @@ func (m *baseMeta) Open(ctx Context, inode Ino, flags uint32, attr *Attr) (st sy
 	switch flags & (syscall.O_RDONLY | syscall.O_WRONLY | syscall.O_RDWR) {
 	case syscall.O_RDONLY:
 		mmask = MODE_MASK_R
+		// 0x20 means O_FMODE_EXEC
```
jiefenghuang, line 1357: Can you explain the reason for this change or provide documentation related to O_FMODE_EXEC? I haven't found much relevant information, is it some special logic in certain distributions?

201341: hi, 0x20 comes from the linux kernel, But it doesn't expose as the ACCESS MODE. you can find this O_FMODE_EXEC

https://stackoverflow.com/questions/3657003/unknown-open-flag-passed-by-execve
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/linux/fs.h?id=HEAD#l75
https://github.com/gluster/glusterfs/blob/devel/libglusterfs/src/glusterfs/glusterfs.h#L36

jiefenghuang: Ok, the call originating from execve will include the O_FMODE_EXEC flag.

jiefenghuang (approved):
LGTM

Xuyuchao-juice:
Unit test has covered

PR #4965, opened 2024-06-21
Flush staging files on close as much as possible

polyrabbit:
Flush staging files as much as possible, making writeback safer to turn on.


jiefenghuang:
I think this is not an essential requirement for all writeback scenarios
It would be more appropriate to place it in a separate command as a tool.

jiefenghuang:
It would be better to make this into a separate command, where flushALL does not force uploading stage files.

PR #4979, opened 2024-06-28
fix: flush volume stats on session close

polyrabbit:
We observed several volumes having much larger usage than the actual data size. It's confirmed when two of them are emptied but still have a large usage, and `juicefs summary` shows there are no files.
Here fix one possible reason: the usage is not flushed on close - thus we have different volume stats than dir stats.


jiefenghuang (approved):
LGTM

PR #4981, opened 2024-07-01
chunk: add an env "JFS_DROP_OSCACHE" to drop page cache for block cache

SandyXSD:


polyrabbit:
Which scenario it is used for, and why not dropping cache after flushing stage blocks?

jiefenghuang:
> Which scenario it is used for, and why not dropping cache after flushing stage blocks?

1. reduce page cache usage
2. upcoming uploads requirement

polyrabbit:
I know it reduces page cache. The question is in which scenario user needs to turn on this flag, like avoiding OOM caused by page cache we encountered [before](https://github.com/juicedata/juicefs/pull/4781)?

lidaohang:
mark

PR #5001, opened 2024-07-17
fix(cmd/fsck): inode is considered broken only if ObjectStorage.Head returns ErrNotExist

jiefenghuang:


cmd/fsck.go
```diff
@@ -186,15 +188,19 @@ func fsck(ctx *cli.Context) error {
...
+						if path, ok = brokens[inode]; !ok {
 							if ps := m.GetPaths(meta.Background, inode); len(ps) > 0 {
-								brokens[inode] = ps[0]
+								path = ps[0]
 							} else {
-								brokens[inode] = fmt.Sprintf("inode:%d", inode)
+								path = fmt.Sprintf("inode:%d", inode)
 							}
 						}
-						logger.Errorf("can't find block %s for file %s: %s", objKey, brokens[inode], err)
```
SandyXSD, line 196: For other errors we can't tell exactly whether the file is broken or not, so better mark it suspicious and log the error. Then users can do double-check of the object/file based on the error.

Issue #5010, opened 2024-07-19
object should have etag when gateway is started with --etag.


zhoucheng361:
**What happened**:
```
root@bench-01:~/juicefs# ./juicefs format sqlite3://test.db test
2024/07/19 16:05:03.596102 juicefs[2148417] <INFO>: Meta address: sqlite3://test.db [interface.go:504]
2024/07/19 16:05:03.596689 juicefs[2148417] <INFO>: Data use file:///var/jfs/test/ [format.go:484]
2024/07/19 16:05:03.663798 juicefs[2148417] <INFO>: Volume is formatted as {
  "Name": "test",
  "UUID": "05bdf5ac-c267-4703-b8d6-d4cdc88c10f2",
  "Storage": "file",
  "Bucket": "/var/jfs/",
  "BlockSize": 4096,
  "Compression": "none",
  "EncryptAlgo": "aes256gcm-rsa",
  "TrashDays": 1,
  "MetaVersion": 1,
  "MinClientVersion": "1.1.0-A",
  "DirStats": true,
  "EnableACL": false
} [format.go:521]
root@bench-01:~/juicefs# MINIO_ROOT_USER=minioadmin MINIO_ROOT_PASSWORD=minioadmin ./juicefs gateway  sqlite3://test.db localhost:9006 --keep-etag -d --multi-buckets
2024/07/19 16:06:41.292124 juicefs[2148505] <INFO>: Meta address: sqlite3://test.db [interface.go:504]
2024/07/19 16:06:41.292617 juicefs[2148505] <INFO>: open log file /var/log/juicefs-gateway.log: %!s(<nil>) [mount_unix.go:248]
2024/07/19 16:06:41.793750 juicefs[2148505] <INFO>: OK, service is ready on localhost:9006 [mount_unix.go:222]
root@bench-01:~/juicefs# mc alias set myminio http://localhost:9006 minioadmin minioadmin
Added `myminio` successfully.
root@bench-01:~/jfs# mc mb myminio/abc
Bucket created successfully `myminio/abc`.
root@bench-01:~/jfs# mc cp README.md myminio/abc/
[...]

PR #5013, opened 2024-07-22
fix: list object should return etag information when needed

zhijian-pro:
fix https://github.com/juicedata/juicefs/pull/4969


zhoucheng361:
fix https://github.com/juicedata/juicefs/issues/5010

PR #5019, opened 2024-07-23
feat(cmd/fsck): refactor and add path for data check

### LINKED ISSUE #4994: fsck: check data integrity of specified path

jiefenghuang:
close #4994

### DIFF
--- a/cmd/fsck.go
+++ b/cmd/fsck.go
@@ -52,12 +52,14 @@ $ juicefs fsck redis://localhost --path /d1/d2 --repair
 $ juicefs fsck redis://localhost --path /d1/d2 --recursive`,
 		Flags: []cli.Flag{
 			&cli.StringFlag{
-				Name:  "path",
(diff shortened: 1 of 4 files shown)


cmd/fsck.go
```diff
@@ -51,172 +51,200 @@ $ juicefs fsck redis://localhost --path /d1/d2 --repair
...
-		parts := strings.Split(obj.Key(), "/")
-		if len(parts) != 3 {
-			continue
+		blob, err := createStorage(*format)
+		if err != nil {
+			logger.Fatalf("object storage: %s", err)
+		}
+		logger.Infof("Data use %s", blob)
+		blob = object.WithPrefix(blob, "chunks/")
+		objs, err := osync.ListAll(blob, "", "", "", true)
```
davies, line 182: it's slow to list all objects when we only need to check a single file.
We should use Head to check each object instead.

jiefenghuang: fixed

PR #5020, opened 2024-07-25
object/gs: use clients pool to optimize gs performance

zhijian-pro:


pkg/object/gs.go
```diff
@@ -117,9 +128,17 @@ func (g *gs) Get(key string, off, limit int64, getters ...AttrGetter) (io.ReadCl
...
 func (g *gs) Put(key string, data io.Reader, getters ...AttrGetter) error {
-	writer := g.client.Bucket(g.bucket).Object(key).NewWriter(ctx)
+	writer := g.getClient().Bucket(g.bucket).Object(key).NewWriter(ctx)
 	writer.StorageClass = g.sc
-	_, err := io.Copy(writer, data)
+
+	// If you upload small objects (< 16MiB), you should set ChunkSize
+	// to a value slightly larger than the objects' sizes to avoid memory bloat.
+	// This is especially important if you are uploading many small objects concurrently.
+	writer.ChunkSize = 5 << 20
```
jiefenghuang, line 137: maybe set ChunkSize to 1.1 * BlockSize, upper limit 5MiB.

zhijian-pro: The blocksize is not easy to pass in.This can be improved later

jiefenghuang (approved):

pkg/object/gs.go
SandyXSD: Add a FIXME

SandyXSD (approved):

PR #5083, opened 2024-08-13
fix invalid file matching regular expressions

zhijian-pro:
fix with hash prefix path `chunks/5D/6/6174/6174813_0_8388608`

### DIFF
--- a/pkg/chunk/disk_cache.go
+++ b/pkg/chunk/disk_cache.go
@@ -895,7 +895,7 @@ func (cache *cacheStore) scanCached() {
 	cache.Unlock()
 }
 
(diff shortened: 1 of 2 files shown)


pkg/chunk/disk_cache.go
```diff
@@ -895,7 +895,7 @@ func (cache *cacheStore) scanCached() {
 	cache.Unlock()
 }
 
-var pathReg, _ = regexp.Compile(`^chunks/\d+/\d+/\d+_\d+_\d+$`)
+var pathReg, _ = regexp.Compile(`^chunks/((\d+)|(..))/\d+/\d+_\d+_\d+$`)
```
jiefenghuang, line 898: maybe the regex could match more precisely, like [0-9a-fA-F]{2} for hash prefix.

jiefenghuang (approved):

PR #5084, opened 2024-08-13
set the upload flag to false if the upload fails

zhijian-pro:


pkg/chunk/cached_store.go
```diff
@@ -945,44 +945,44 @@ func parseObjOrigSize(key string) int {
...
 	}()
 
 	if !store.canUpload() {
-		return
+		return nil
 	}
 	store.pendingMutex.Lock()
 	item, ok := store.pendingKeys[key]
 	store.pendingMutex.Unlock()
 	if !ok {
```
jiefenghuang, line 960: maybe `defer item.uploading = false` is enough, no matter success or failed, we just want to filter the duplicated process.

jiefenghuang (approved):

PR #5087, opened 2024-08-15
feat(fuse): support write size config

jiefenghuang:


Xuyuchao-juice:
Test steps:
```dd if=/dev/zero of=./testfile bs=1M count=1024```
```./juicefs stats ~/mountpoint```

Test Result:
image
861Mb / 6914 = 127.5k

PR #5092, opened 2024-08-16
fix: Handle ENOENT error by reducing stageBlocks and stageBlockBytes

zhijian-pro:


zhijian-pro:
It is also reduced when clearing the read cache, so there is no need to change it if the change is too large

pkg/chunk/disk_cache.go
```diff
@@ -385,6 +385,8 @@ func (cache *cacheStore) removeStage(key string) error {
 	}
 	// ignore ENOENT error
 	if err != nil && os.IsNotExist(err) {
```
jiefenghuang, line 387: remove will call removeStage (like tryRemoveStage), may  incorrectly update stats

PR #5095, opened 2024-08-19
warmup: evict sub command may remove stage incorrectly

jiefenghuang:
Add a new interface to clear the read cache. 
Since there are already interfaces like remove and removeStage, a flag-based merge is not used to avoid too many API changes. Moreover, the read and write cache logic is actually quite independent, so it is better to keep them separate.

### DIFF
--- a/pkg/chunk/disk_cache.go
+++ b/pkg/chunk/disk_cache.go
@@ -579,6 +579,11 @@ func (cache *cacheStore) removeReadCache(key string) {
 		return
 	}
 
+	// skip stage
+	if utils.Exists(cache.stagePath(key)) {
+		return
+	}
+
 	path := cache.doRemove(key)
 	if path != "" {
 		if err := cache.removeFile(path); err != nil && !os.IsNotExist(err) {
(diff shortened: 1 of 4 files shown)


pkg/chunk/disk_cache.go
```diff
@@ -1005,6 +1025,7 @@ func expandDir(pattern string) []string {
 type CacheManager interface {
 	cache(key string, p *Page, force, dropCache bool)
 	remove(key string)
+	removeReadCache(key string) // remove read cache only
```
davies, line 1028: it's better to add a flag in `remove(key)`

polyrabbit:
Kind reminder - when file is not uploaded to S3, it's dangerous to delete it from cache folder, especially when user enabled `upload-delay` or `upload-hours` features. The read procedure only checks file existence in cache folder and S3, not checking stage folder, so user will get an `EIO` error.

I suggest the evict subcommand should just ignore staging files.

zhijian-pro (approved):

PR #5109, opened 2024-08-23
cmd/clone: cloning trash files is prohibited

zhijian-pro:


pkg/meta/base.go
```diff
@@ -2437,12 +2437,6 @@ func (m *baseMeta) ScanDeletedObject(ctx Context, tss trashSliceScan, pss pendin
 }
 
 func (m *baseMeta) Clone(ctx Context, srcIno, parent Ino, name string, cmode uint8, cumask uint16, count, total *uint64) syscall.Errno {
-	if isTrash(parent) {
```
jiefenghuang, line 2440: The conditional restrictions of the Clone should be placed here, as they may also be used elsewhere.

jiefenghuang (approved):
LGTM

Issue #2831, opened 2022-10-08
Big directory when using tikv as metadata engine 


penglaixy99:
**What happened**:
I have written 20 million files in a single directory.
ls command has stuck for 40 min, and the memory used has increased from 71MB to 10GB, but  soon  decreased slowly , and finally to 4.5GB.  But the ls command has no result. 

**What you expected to happen**:
ls command output normally.

**How to reproduce it (as minimally and precisely as possible)**:
write 2000 0000 files in a single direcotry.

**Anything else we need to know?**
When ls command is executing, 8 clients are stilling writing files to the directory.

**Environment**:
- JuiceFS version (use `juicefs --version`) or Hadoop Java SDK version: 1.0.0
- Cloud provider or hardware configuration running JuiceFS:
- OS (e.g `cat /etc/os-release`):  centos 7.4
- Kernel (e.g. `uname -a`): 
- Object storage (cloud provider and region, or self maintained):  ks3
- Metadata engine info (version, cloud provider managed or self maintained): tikv 2.0.1
- Network connectivity (JuiceFS to metadata engine, JuiceFS to object storage):
- Others:

SandyXSD:
Could you get the juicefs.log and part of the [access log](https://juicefs.com/docs/community/fault_diagnosis_and_analysis/#access-log) when running the `ls` command?

penglaixy99:
LOG :
slow operation: readdir (8302,4096,0): OK (18402593) <128.107549> [accesslog.go:65]    

entries number is correct. 

But ls command is stuck.  I will retry the command and set the fuse debug on.

SandyXSD:
The first `readdir` will retrieve all entries (20m) and cache them for the following `readdir`s, so it's normal that it takes a bit long like 2 mins.
Usually the access log should be more useful in such case.

penglaixy99:
I have a  retry.

11:44:27  juicefs begin to load entries from tikv.
2022/10/08 11:44:27 rx 27: READDIRPLUS n8302 {Fh 2 [0 +4096)  L 0 DIRECTORY,NONBLOCK,0x8000}



11:46:53   all entries are cached.
2022/10/08 11:46:53.347175 vivo_juicefs[2680] <INFO>: slow operation: readdir (8302,4096,0): OK (18593782) <145.730275> [accesslog.go:65]


then all entries are filled to the kernel fuse.

[...]

Issue #2594, opened 2022-08-24
"ls" takes a very long time when using NFS on JuiceFS


huntersman:
**What happened**:
Test `ls` performace on NFS. It takes about 10 minutes for 20000 files.
**What you expected to happen**:
Less than 1 second for 20000 files.
**How to reproduce it (as minimally and precisely as possible)**:
### JuiceFS and NFS server
```bash
juicefs format \
    --storage s3 \
    --bucket http://serverIp/bucketName \
    --access-key ak \
    --secret-key sk \
    "mysql://user:password@(127.0.0.1:3306)/jfs" \
    myjfs
```
```bash
juicefs mount -d "mysql://user:password@(127.0.0.1:3306)/jfs" /mnt/jfs
```
```bash
yum -y install nfs-utils rpcbind
```
```bash
vi /etc/exports
/mnt/jfs *(rw,insecure,async,no_subtree_check,no_root_squash,fsid=0)

service rpcbind start
service nfs start
```
### NFS client
```bash
mount -t nfs -o nfsvers=3 -o sync serverIp:/mnt/jfs /mnt/nfs
```
Add 20000 4KB files into /mnt/nfs
```bash
seq 20000|xargs -i dd if=/dev/zero of=/mnt/nfs/testfile_{}.cache bs=4k count=1
```
Test `ls` performance and it takes about 10 minutes.
```bash
time ls /mnt/nfs
real 10m36.130s
user 0m0.105s
sys  0m0.378s
```
**Anything else we need to know?**
I use `juicefs stats /mnt/jfs` and notice that cpu usage becomes really high for a long time.

BTW, I test only use JuiceFS or NFS, both performance are good.
```bash
# 20000 4KB files
time ls /mnt/jfs
real 0m0.335s
user 0m0.054s
sys  0m0.031s
```
**Environment**:
- JuiceFS version (use `juicefs --version`) or Hadoop Java SDK version: juicefs version 1.0.0+2022-08-08.cf0c269
- OS (e.g `cat /etc/os-release`): CentOS 7
- Kernel (e.g. `uname -a`): 5.4.210-1.el7.elrepo.x86_64
- Object storage (cloud provider and region, or self maintained): self maintained
- Metadata engine info (version, cloud provider managed or self maintained): self maintained, MySQL 5.7.33

Thanks for your help!

SandyXSD:
[...]

PR #5110, opened 2024-08-25
feat: support streaming readdir for kv meta

### LINKED ISSUE #5099: [Suggestion] Change VFS readdir Function Behavior from Full Data Scan to Read-Ahead
## Background

We built our JuiceFS storage in 2022, two years ago, using TiKV and Ali OSS. It used to work fine with rare mistakes. However, things have become difficult recently. More and more AI engineers are using this JuiceFS storage to store images and other datasets. Even worse, some use it as their working directory.

They store files ranging from 100k to 10 million in a single directory. I know this is not a good practice, but it has already happened. Some data volumes have really large directories.

## Suggestion for readdir

Currently, when an application like `gnu ls` calls readdir(3), FUSE will send a readdir or readdirplus request to JuiceFS. This process can be simplified as: `user process readdir(3) -> glibc getdents64 syscall -> FUSE readdirplus -> JuiceFS readdirplus`.

The readdir is a streaming request and will be called multiple times by the user application process. However, JuiceFS loads the full data of the directory when the first readdir request comes in, and then stores the entries in the file handler.

The KVMeta readdir function can be split into two parts:

1. Scanning keys: It scans all keys that start with the `A${Parent_inode}D` prefix. Scanning keys is 10 times slower than the gets operation in TiKV.
2. Getting all file attributes: It uses the scanned keys, which contain all filenames and inodes, to get the files attr.

This approach is inefficient and dangerous. I suspect that when the directory is large and file handlers leak, this big memory will also leak.

Additionally, the readdir request can only use a 4KB buffer to fill file entries limited by `go-fuse`. When the directory size reaches millions, it will hold the memory for a long time.

Therefore, it might be better to change the `readdir` function's behavior from a full data scan to read-ahead.

SonglinLife:
close #5099
This pr try to addresses the issue #5099, where reading large directories consumes massive memory. This problem has persisted for a long time without a resolution. 

#5099 already discusses the current inefficient implementation:

1. Memory Usage: The implementation scans all file entries in a directory when readdir request first come. If the directory is large (e.g., 10M entries), it results in very high peak RSS memory usage.
2. Time Efficiency: The readdir(3) function is a streaming function that calls the getdents syscall multiple times. When scanning data from the metadata engine, the FUSE server just waits. It would be better to return some data to FUSE and fetch the remaining data in the background. also, the full data scan would suck the tools like `ls`, w already see some complaint about `ls take a long time` https://github.com/juicedata/juicefs/issues/2831 https://github.com/juicedata/juicefs/issues/2594 also memtioned this problem.

This pr try to solve this problem. According to the [TiKV documentation](https://tikv.org/blog/how-tikv-reads-writes/), keys are stored in a certain order. When using an iterator, TiKV returns a stored key-value list, so we can move forward from the startKey based on the last results.

When vfs readdir is called, it invokes metaScanner.GetData. This scanner caches **4096** entries. If the read offset is out of cache, it waits for the in-flight TiKV scan to complete and then starts a new goroutine to prefetch. A flag streaming-readdir indicates whether to enable this feature.

I am not a proficient Go developer , so the code might not be well-styled. I would appreciate any guidance or hep for refine this pr code

### Why Fetch 4096 File Entries?

I tested this in our production machine. In a simple test of ls on a directory with 5 million files, setting the fetch size to 4096 entries took. it only 1 minute and used 139MB peak memory to get all entries, compared to 3.7GB with a full data scan implementation—10x lower memory usage. I believe this middle-ground number is reasonable for both large and small directories.

and there also more information:
<p align="center">
  image
  image
</p>

Now, we can use `ls -1 --sort=none` to display files immediately, as juicefs return data to FUSE as soon as it is available.

### DIFF
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -2864,7 +2864,7 @@ func (sc *baseEntrySanner) GetData(start int) ([]*Entry, syscall.Errno) {
 	return res, 0
 }
 
-func (m *baseMeta) NewBaseEntryScanner(ctx Context, inode Ino, plus uint8) (mc EntryScanner, rerr syscall.Errno) {
+func (m *baseMeta) NewBaseEntryScanner(ctx Context, inode Ino, plus uint8, InternalEntries []*Entry) (mc EntryScanner, rerr syscall.Errno) {
 	var attr Attr
 	defer func() {
 		if rerr == 0 {
@@ -2898,6 +2898,7 @@ func (m *baseMeta) NewBaseEntryScanner(ctx Context, inode Ino, plus uint8) (mc E
 		Name:  []byte(".."),
 		Attr:  &Attr{Typ: TypeDirectory},
 	})
+	entries = append(entries, InternalEntries...)
 	mc = &baseEntrySanner{
 		c:              m.en.NewMetaEntryScanner(inode, (plus == 1)),
 		specialEntries: entries,
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -1547,40 +1547,50 @@ func (m *baseMeta) Fallocate(ctx Context, inode Ino, mode uint8, off uint64, siz
 	return st
 }
 
-func (m *baseMeta) Readdir(ctx Context, inode Ino, plus uint8, entries *[]*Entry) (rerr syscall.Errno) {
+func (m *baseMeta) getDotEntries(ctx Context, inode Ino, plus uint8) ([]*Entry, syscall.Errno) {
 	var attr Attr
-	defer func() {
-		if rerr == 0 {
-			m.touchAtime(ctx, inode, &attr)
-		}
-	}()
+
 	inode = m.checkRoot(inode)
 	if err := m.GetAttr(ctx, inode, &attr); err != 0 {
-		return err
+		return nil, err
 	}
-	defer m.timeit("Readdir", time.Now())
 	var mmask uint8 = MODE_MASK_R
 	if plus != 0 {
 		mmask |= MODE_MASK_X
 	}
 	if st := m.Access(ctx, inode, mmask, &attr); st != 0 {
-		return st
+		return nil, st
 	}
 	if inode == m.root {
 		attr.Parent = m.root
 	}
-	*entries = []*Entry{
+	entries := []*Entry{
 		{
 			Inode: inode,
 			Name:  []byte("."),
 			Attr:  &Attr{Typ: TypeDirectory},
 		},
 	}
-	*entries = append(*entries, &Entry{
+	entries = append(entries, &Entry{
 		Inode: attr.Parent,
 		Name:  []byte(".."),
 		Attr:  &Attr{Typ: TypeDirectory},
 	})
+	return entries, 0
+}
+
+func (m *baseMeta) Readdir(ctx Context, inode Ino, plus uint8, entries *[]*Entry) (rerr syscall.Errno) {
+	defer m.timeit("Readdir", time.Now())
+	defer func() {
+		if rerr == 0 {
+			m.touchAtime(ctx, inode, nil)
+		}
+	}()
+	if dotEntries, err := m.getDotEntries(ctx, inode, plus); err != 0 {
+		return err
+	} else {
+		*entries = dotEntries
+	}
 	st := m.en.doReaddir(ctx, inode, plus, entries, -1)
 	if st == syscall.ENOENT && inode == TrashInode {
 		st = 0
@@ -2838,6 +2848,7 @@ func (b *MockMetaEntryScannerGen) NewMetaEntryScanner(inode Ino, plus bool) Entr
 
 type baseEntrySanner struct {
 	c              EntryScanner
+	m              *baseMeta
 	specialEntries []*Entry
 }
 
@@ -2848,6 +2859,7 @@ func (sc *baseEntrySanner) Close() {
 	sc.c.Close()
 }
 func (sc *baseEntrySanner) GetData(start int) ([]*Entry, syscall.Errno) {
+	defer sc.m.timeit("ScanEntries", time.Now())
 	if !sc.Valid() {
 		return nil, 0
 	}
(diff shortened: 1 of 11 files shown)


jiefenghuang:
@SonglinLife can you set "Allow edits from maintainers" so we can make some changes?

pkg/meta/base.go
```diff
@@ -2808,3 +2810,98 @@ func inGroup(ctx Context, gid uint32) bool {
...
+	}
+	if entries, err := sc.c.GetData(start - len(sc.specialEntries)); err != 0 {
+		return make([]*Entry, 0), err
+	} else {
+		res = append(res, entries...)
+	}
+	return res, 0
+}
+
+func (m *baseMeta) NewBaseEntryScanner(ctx Context, inode Ino, plus uint8, InternalEntries []*Entry) (mc EntryScanner, rerr syscall.Errno) {
```
jiefenghuang, line 2867: 1. maybe no need to nest a Scanner implementation.
just move kvEntryScanner.GetData to baseEntryScanner.getData

pkg/meta/base.go
```diff
@@ -121,6 +121,8 @@ type engine interface {
 	doSetFacl(ctx Context, ino Ino, aclType uint8, rule *aclAPI.Rule) syscall.Errno
 	doGetFacl(ctx Context, ino Ino, aclType uint8, aclId uint32, rule *aclAPI.Rule) syscall.Errno
 	cacheACLs(ctx Context) error
+
```
jiefenghuang, line 124: base on comment 1, remove this?
we need one new API only

pkg/meta/tkv.go
```diff
@@ -3756,3 +3758,146 @@ func (m *kvMeta) loadDumpedACLs(ctx Context) error {
...
+			})
+			cnt++
+			return cnt < limit
+		})
+		return nil
+	}, 0)
+	return entries, err
+}
+
+func (sc *kvEntryScanner) fillAttr(entries []*Entry) (err error) {
```
jiefenghuang, line 3826: change and reuse fillAttr

pkg/meta/tkv.go
```diff
@@ -3756,3 +3758,146 @@ func (m *kvMeta) loadDumpedACLs(ctx Context) error {
...
+func (sc *kvEntryScanner) Close() {
+	if sc.bgChan != nil {
+		_ = <-sc.bgChan
+		sc.cache = make([]*Entry, 0)
+		sc.cnt = 0
+		sc.bgChan = nil
+	}
+}
+
+func (sc *kvEntryScanner) scanEntries(startKey, endKey []byte, limit int) ([]*Entry, error) {
```
jiefenghuang, line 3798: change and reuse scanValues()?

pkg/vfs/vfs.go
```diff
@@ -425,6 +425,35 @@ func (v *VFS) Readdir(ctx Context, ino Ino, size uint32, off int, fh uint64, plu
 	h.Lock()
 	defer h.Unlock()
 
+	if v.Conf.Meta.StreamingReadDir && h.sc == nil {
```
jiefenghuang, line 428: change to this?
```
if v.Conf.Meta.StreamingReadDir {
  if h.sc == nil {
	  var inodes []*meta.Entry
	  if ino == rootID && !v.Conf.HideInternal {
		  inodes = internalDEntries[1:]
	  }
	  if h.sc, err = v.Meta.NewEntryScanner(ctx, ino, plus, inodes); err != 0 {
		  return
	  }
  }
  if h.sc.Valid() {
	  if entries, err = h.sc.GetData(off); err != 0 {
		  return
	  }
	  readAt = time.Now()
	  logger.Debugf("readdir: %d entries streaming, off=%d", len(entries), off)
	  return
  }
}
```

pkg/meta/base.go
SonglinLife: Thank you for your suggestion. Based on my current understanding of juicefs code structure, we cannot operate directly on the engine at the **baseMeta** level. juicefs‘s architecture requires that it only call the functions specified in the engine interface, such as `doReaddir` and `doWrite`. Direct calls to TKV functions like `scanValues` or `gets` at this level are not feasible.

To address this, my current approach is to use a baseEntry to handle the interface boundaries.

pkg/meta/tkv.go
SonglinLife: `fillAttr` is a lambda function https://github.com/juicedata/juicefs/blob/main/pkg/meta/tkv.go#L1799  , it is ok to extract it as a new function `func (m *kvMeta) fillAttres([]*Entry) error `

pkg/meta/tkv.go
SonglinLife: current, scanValues only take starKey, and use `NextKey(startKey)` as endKey, if a startKey is `fooD48` the endKey will be `fooD49`, but we want by endKey be `fooE` to retrieve any keys in `fooD48` to `fooE`.

It need to change `func (m *kvMeta) scanValues(prefix []byte, limit int, filter func(k, v []byte) bool) (map[string][]byte, error) ` to `func (m *kvMeta) scanValues(startKey, endKey []byte, limit int, filter func(k, v []byte) bool) (map[string][]byte, error) `

or Just add a new func  `func (m *kvMeta) scanValues2(startKey, endKey []byte, limit int, filter func(k, v []byte) bool) (map[string][]byte, error) `

SonglinLife:
> @SonglinLife can you set "Allow edits from maintainers" so we can make some changes?

I apologize that I cannot enable this feature. GitHub does not support "Allow edits from maintainers" when the PR comes from an organization.

pkg/vfs/vfs.go
SonglinLife: yes, It looks more clear.

pkg/meta/tkv.go
jiefenghuang: yes

pkg/meta/tkv.go
jiefenghuang: maybe like this
```
func (m *kvMeta) scanValues(prefix []byte, limit int, filter func(k, v []byte) bool) (map[string][]byte, error) {
	return m.doScanValues(prefix, nextKey(prefix), limit, filter)
}

func (m *kvMeta) doScanValues(startKey, endKey []byte, limit int, filter func(k, v []byte) bool) (map[string][]byte, error) {
	if limit == 0 {
		return nil, nil
	}
	values := make(map[string][]byte)
	err := m.client.txn(func(tx *kvTxn) error {
		var c int
		tx.scan(startKey, endKey, false, func(k, v []byte) bool {
			if filter == nil || filter(k, v) {
				values[string(k)] = v
				c++
			}
			return limit < 0 || c < limit
		})
		return nil
	}, 0)
	return values, err
}
```

pkg/meta/tkv.go
```diff
@@ -3756,3 +3758,146 @@ func (m *kvMeta) loadDumpedACLs(ctx Context) error {
...
+			sc.m.parseAttr(re, entries[j].Attr)
+		}
+	}
+	return err
+}
+
+func (sc *kvEntryScanner) GetData(start int) ([]*Entry, syscall.Errno) {
+	if sc.bgChan == nil {
+		sc.bgChan = make(chan []*Entry, 1)
+		go sc.fetchEntries()
```
jiefenghuang, line 3853: Maybe we should conduct a performance benchmark test to confirm if there is a need to control the number of goroutines. 
In user scenarios, the creation of an excessive number of goroutines in this section could potentially cause a decline in performance, which may also affect other APIs within the system.

SonglinLife: Sorry for late reply. I have already rewritten this piece of code using the waitgroup to synchronize. Now, the background goroutines will not remain alive once they have completed their data-fetching tasks. Could you please confirm if this modification is acceptable?

pkg/meta/base.go
SonglinLife: still a bit confused. I believe that this interface is essential. Based on the current code structure, we can't handle all kvmeta operations within baseMeta. Could you provide some specific suggestions or guidance on how to approach this? I'm having trouble figuring it out.

jiefenghuang: My apologies for not being clear enough.
- move engine API NewMetaEntryScanner
- keep Meta API NewBaseEntryScanner (maybe change the name to NewEntryScanner , would be clearly enough)
- move the implement in baseMeta to kvMeta, merge them

SonglinLife: thanks for your hints, now i remove this unnecessary **BaseEntryScanner**.

jiefenghuang:
After our internal discussion, we will reconsider a plan to adapt to multiple engines later, and for the TiKV part, we will refer to your implementation.

jiefenghuang:
@SonglinLife Hi, I have submitted a new pr that supports other meta engines based on this pr. #5162

SonglinLife:
thx for the awesome work on PR #5162 , learn a lot from it!

PR #5119, opened 2024-08-29
feat: WebDAV no need to call MkdirAll

myml:
WebDAV clients automatically create directories when writing files, so there's no need to call MkdirAll in advance.

https://github.com/studio-b12/gowebdav/blob/3b85fdfd9eece3af01775fb566e633a2065bc534/client.go#L421


zhijian-pro:
@myml  The behavior of `MkdirAll` is necessary, for example when `key=c/`, the expected behavior is to create the directory rather than doing nothing.
Our unit test found this case. You can run the [TestWebDAV](https://github.com/juicedata/juicefs/blob/6c5f5e91782ce1548151465fb26aa6c62ef455fb/pkg/object/object_storage_test.go#L856) test to debug this behavior

Issue #5134, opened 2024-09-03
random read cause repeat prefetch same blocks


sjp00556:
**What happened**:
random read will cause repeat prefetch same blocks，which will case bandwidth amplification

**How to reproduce it (as minimally and precisely as possible)**:
fio -filename=./f1 -direct=1 -iodepth 1 -thread -rw=randread -ioengine=psync -bs=128K -size=100m -numjobs=1  -group_reporting -name=mytest

the same key may be prefetched repeatedly

![screenshot-20240903-000450](https://github.com/user-attachments/assets/9a685208-9ab6-4e2e-8d58-75453f212509)

SandyXSD:
Fixed by #4947.

PR #5135, opened 2024-09-03
enhance: cache store check key exist before prefetch

sjp00556:
random read will cause repeat prefetch same blocks，which will case bandwidth amplification

https://github.com/juicedata/juicefs/issues/5134

### DIFF
--- a/pkg/chunk/cached_store.go
+++ b/pkg/chunk/cached_store.go
@@ -819,6 +819,9 @@ func NewCachedStore(storage object.ObjectStorage, config Config, reg prometheus.
 		if size == 0 || size > store.conf.BlockSize {
 			return
 		}
+		if store.bcache.exist(key) {
+			return
+		}
 		p := NewOffPage(size)
 		defer p.Release()
 		_ = store.load(key, p, true, true)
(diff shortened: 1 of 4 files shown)


pkg/chunk/cached_store.go
```diff
@@ -819,6 +819,9 @@ func NewCachedStore(storage object.ObjectStorage, config Config, reg prometheus.
 		if size == 0 || size > store.conf.BlockSize {
 			return
 		}
+		if store.bcache.exist(key) {
```
davies, line 822: we can use bcache.load()

sjp00556: bcache.load will open cache file which is more costly. so i add bcache.exist function. if cache is scaned, only check block exist in keys map

polyrabbit:
@sjp00556 You also need to manually pick PR #4947 to get rid of duplicate prefetch.

polyrabbit:
I suppose those are two different duplications - #4947 fixes duplication between prefetches, #5135 is for duplication between prefetch and readahead.

SandyXSD:
This PR has nothing to do with the readahead, and #5134 has already been fixed by #4947.

sjp00556:
> This PR has nothing to do with the readahead, and #5134 has already been fixed by #4947.

yes, #4947 fixed this issue

PR #5142, opened 2024-09-04
cmd/sync: add 'max-failure' to exit when the transfer encounters more than a certain number of errors

### LINKED ISSUE #5140: Sync tool support fail on error mode

Add an option to allow sync tool to fail fast.

**Why is this needed**:
Currently producer has no feedback from workers. When workers encounter a fatal error, the producer will not stop until it has read all data, which is time consuming. It will be helpful to fail fast if any error happens.

zhijian-pro:
close #5140

### DIFF
--- a/pkg/sync/sync.go
+++ b/pkg/sync/sync.go
@@ -1318,11 +1318,11 @@ func Sync(src, dst object.ObjectStorage, config *Config) error {
 
 	if !config.Dry {
 		failed = progress.AddCountSpinner("Failed objects")
-		if config.FailFast {
+		if config.MaxFailure > 0 {
 			go func() {
 				for {
-					if failed.Current() > 0 {
-						logger.Infof("A transmission error occurred and the quick exit process began")
+					if failed.Current() >= config.MaxFailure {
+						logger.Infof("%d transmission error occurred and the quick exit process began", config.MaxFailure)
 						if syncExitFunc() != nil {
 							os.Exit(1)
 						}
--- a/pkg/sync/sync.go
+++ b/pkg/sync/sync.go
@@ -18,6 +18,7 @@ package sync
 
 import (
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


zhijian-pro:
> Maybe --max-errors or --max-failure? we have --max-failure for warmup

ok

pkg/sync/sync.go
```diff
@@ -1282,8 +1282,55 @@ func Sync(src, dst object.ObjectStorage, config *Config) error {
...
+		}
+		return nil
+	}
+
 	if !config.Dry {
 		failed = progress.AddCountSpinner("Failed objects")
+		if config.MaxFailure > 0 {
+			go func() {
+				for {
+					if failed.Current() >= config.MaxFailure {
```
davies, line 1324: >

pkg/sync/sync.go
```diff
@@ -1282,8 +1282,55 @@ func Sync(src, dst object.ObjectStorage, config *Config) error {
...
+						logger.Infof("the maximum error limit of %d was reached, triggering the quick exit process", config.MaxFailure)
+						if syncExitFunc() != nil {
```
davies, line 1326: will it return nil?

pkg/sync/sync.go
```diff
@@ -1282,8 +1282,55 @@ func Sync(src, dst object.ObjectStorage, config *Config) error {
...
+				msg += fmt.Sprintf(", lost: %d", total-handled.Current())
+			}
+			logger.Info(msg)
+		} else {
+			sendStats(config.Manager)
+			logger.Debugf("This worker process has already completed its task")
+		}
+		if failed != nil {
+			if n := failed.Current(); n > 0 || total > handled.Current() {
+				return fmt.Errorf("failed to handle %d objects", n+total-handled.Current())
```
davies, line 1313: When failed < maxFailure, should we exit with 0?

zhijian-pro: I think users expect quick results, but that doesn't mean results aren't important. So any failure should return a non-zero error code.

jiefenghuang (commented):
for supporting fast failed, add context to worker may be more concise and simple
```
func worker(ctx context.Context, tasks <-chan object.Object, src, dst object.ObjectStorage, config *Config) {
    for {
        select {
            case <-ctx.Done():
                // cancel
                return
             case obj := <-tasks:
                 // origin logic
        }
        if failed >= xxx {
            ctx.cancel()
            ...
        }
    }
}
```

zhijian-pro:
> for supporting fast failed, add context to worker may be more concise and simple
> 
> ```
> func worker(ctx context.Context, tasks <-chan object.Object, src, dst object.ObjectStorage, config *Config) {
>     for {
>         select {
>             case <-ctx.Done():
>                 // cancel
>                 return
>              case obj := <-tasks:
>                  // origin logic
>         }
>         if failed >= xxx {
>             ctx.cancel()
>             ...
>         }
>     }
> }
> ```

ok

jiefenghuang (approved):
LGTM

zhijian-pro:
> for supporting fast failed, add context to worker may be more concise and simple
> 
> ```
> func worker(ctx context.Context, tasks <-chan object.Object, src, dst object.ObjectStorage, config *Config) {
>     for {
>         select {
>             case <-ctx.Done():
>                 // cancel
>                 return
>              case obj := <-tasks:
>                  // origin logic
>         }
>         if failed >= xxx {
>             ctx.cancel()
>             ...
>         }
>     }
> }
> ```

Let's do it the easy way, it's a little bit more complicated

PR #5154, opened 2024-09-10
enhance: rm object when write failed

### LINKED ISSUE #5132: Object files of non-existing files may leak when we committing it to meta

jiefenghuang:
close #5132 

remove failed objects as much as possible, non-blocking.

### DIFF
--- a/pkg/chunk/cached_store.go
+++ b/pkg/chunk/cached_store.go
@@ -854,7 +854,7 @@ func NewCachedStore(storage object.ObjectStorage, config Config, reg prometheus.
 	store.gcJobs = make(chan gcJob, 10)
 	go func() {
 		for job := range store.gcJobs {
-			store.Remove(job.id, job.length)
+			_ = store.Remove(job.id, job.length)
 		}
 	}()
 
--- a/pkg/chunk/cached_store.go
+++ b/pkg/chunk/cached_store.go
@@ -1101,7 +1101,10 @@ func (store *cachedStore) Remove(id uint64, length int) error {
 }
 
 func (store *cachedStore) TryRemove(id uint64, length int) {
-	store.gcJobs <- gcJob{id, length}
+	select {
+	case store.gcJobs <- gcJob{id, length}:
+	default:
+	}
 }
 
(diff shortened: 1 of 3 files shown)


solracsf:
Superseeds https://github.com/juicedata/juicefs/pull/5133 ?

PR #5162, opened 2024-09-12
meta: support dir stream

### LINKED ISSUE #5099: [Suggestion] Change VFS readdir Function Behavior from Full Data Scan to Read-Ahead

jiefenghuang:
close #5099 
ref #5110 

- redis: unchanged
- sql: read full large directory by pagination will increase the total consumed time.

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -284,6 +284,11 @@ func metaFlags() []cli.Flag {
 			Name:  "sort-dir",
 			Usage: "sort entries within a directory by name",
 		},
+		&cli.BoolFlag{
+			Name:  "readdir-stream",
+			Value: false,
+			Usage: "enable streaming readdir",
+		},
 	})
 }
 
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -284,11 +284,6 @@ func metaFlags() []cli.Flag {
 			Name:  "sort-dir",
 			Usage: "sort entries within a directory by name",
 		},
-		&cli.BoolFlag{
-			Name:  "readdir-stream",
-			Value: false,
-			Usage: "enable streaming readdir",
-		},
 	})
 }
 
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -128,7 +128,7 @@ type engine interface {
 	doGetFacl(ctx Context, ino Ino, aclType uint8, aclId uint32, rule *aclAPI.Rule) syscall.Errno
 	cacheACLs(ctx Context) error
 
-	newDirStream(inode Ino, plus bool, entries []*Entry) DirStream
+	newDirHandler(inode Ino, plus bool, entries []*Entry) DirHandler
 }
 
 type trashSliceScan func(ss []Slice, ts int64) (clean bool, err error)
@@ -2894,7 +2894,7 @@ func inGroup(ctx Context, gid uint32) bool {
 	return false
 }
 
-type DirStream interface {
+type DirHandler interface {
 	List(ctx Context, offset int) ([]*Entry, syscall.Errno)
 	Insert(inode Ino, name string, attr *Attr)
 	Delete(name string)
@@ -2902,7 +2902,7 @@ type DirStream interface {
 	Close()
 }
 
-func (m *baseMeta) NewDirStream(ctx Context, inode Ino, plus bool, initEntries []*Entry) (DirStream, syscall.Errno) {
+func (m *baseMeta) NewDirHandler(ctx Context, inode Ino, plus bool, initEntries []*Entry) (DirHandler, syscall.Errno) {
 	var attr Attr
 	var st syscall.Errno
 	defer func() {
@@ -2915,7 +2915,7 @@ func (m *baseMeta) NewDirStream(ctx Context, inode Ino, plus bool, initEntries [
 	if st = m.GetAttr(ctx, inode, &attr); st != 0 {
 		return nil, st
 	}
-	defer m.timeit("NewDirStream", time.Now())
+	defer m.timeit("NewDirHandler", time.Now())
 	var mmask uint8 = MODE_MASK_R
 	if plus {
 		mmask |= MODE_MASK_X
@@ -2938,7 +2938,7 @@ func (m *baseMeta) NewDirStream(ctx Context, inode Ino, plus bool, initEntries [
 		Attr:  &Attr{Typ: TypeDirectory},
 	})
 
-	return m.en.newDirStream(inode, plus, initEntries), 0
+	return m.en.newDirHandler(inode, plus, initEntries), 0
 }
 
 type dirBatch struct {
@@ -2963,7 +2963,7 @@ func (b *dirBatch) predecessor(offset int) bool {
 
 type dirFetcher func(ctx Context, inode Ino, cursor interface{}, offset, limit int, plus bool) (interface{}, []*Entry, error)
 
-type dirStream struct {
+type dirHandler struct {
 	sync.Mutex
 	inode       Ino
 	plus        bool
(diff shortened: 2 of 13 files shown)


cmd/flags.go
```diff
@@ -284,6 +284,11 @@ func metaFlags() []cli.Flag {
 			Name:  "sort-dir",
 			Usage: "sort entries within a directory by name",
 		},
+		&cli.BoolFlag{
+			Name:  "readdir-stream",
```
davies, line 288: For small dirs, this should be as fast as before, so this should be enabled by default, we don't need this flag

pkg/vfs/handle.go
```diff
@@ -35,10 +35,11 @@ type handle struct {
...
 
 	// for dir
-	children []*meta.Entry
-	readAt   time.Time
-	readOff  int
-	index    map[string]int
+	children  []*meta.Entry
+	readAt    time.Time
+	readOff   int
+	index     map[string]int
```
davies, line 41: we need to update this index

pkg/meta/tkv.go
```diff
@@ -3756,3 +3780,61 @@ func (m *kvMeta) loadDumpedACLs(ctx Context) error {
...
 }
+
+func (m *kvMeta) getDirFetcher(ctx Context) dirFetcher {
+	return func(inode Ino, cursor interface{}, offset, limit int, plus bool) (interface{}, []*Entry, error) {
+		var startKey []byte
+		total := limit
+		if cursor == nil {
+			startKey = m.entryKey(inode, "")
+
+			// should not happen
```
SonglinLife, line 3791: Also believe this should not happen. Even the syscall `readdir/getdents64` can pass a directory offset. For most app process, reading a directory is typically performed sequentially.

pkg/vfs/vfs.go
```diff
@@ -424,45 +424,27 @@ func (v *VFS) Readdir(ctx Context, ino Ino, size uint32, off int, fh uint64, plu
...
 	}
-	if off < len(h.children) {
-		entries = h.children[off:]
-		// we don't know how much of them will be sent, assume all of them
-		h.readOff = len(h.children) - 1
+	if entries, err = h.dirStream.List(ctx, off); err != 0 {
+		return
 	}
-	readAt = h.readAt
+	readAt = start
```
SandyXSD, line 446: `readAt` is used to indicate the read time from the metadata engine, in case some entries are invalidated before replied to FUSE.

pkg/meta/base.go
```diff
@@ -2889,3 +2897,182 @@ func inGroup(ctx Context, gid uint32) bool {
...
+	nextCursor, entries, err := h.fetcher(ctx, h.inode, cursor, offset, h.batchNum, h.plus)
+	if err != nil {
+		return nil, err
+	}
+	if entries == nil {
+		entries = []*Entry{}
+		nextCursor = cursor
+	}
+	isEnd := false
+	if len(entries) < h.batchNum {
```
SandyXSD, line 2998: `isEnd := len(entries) < h.batchNum`

pkg/meta/base.go
```diff
@@ -2889,3 +2897,182 @@ func inGroup(ctx Context, gid uint32) bool {
...
+		return
+	}
+
+	if idx, ok := h.batch.indexes[name]; ok && idx >= h.readOff {
+		delete(h.batch.indexes, name)
+		if n := len(h.batch.entries); n == 1 {
+			h.batch.entries = h.batch.entries[:0]
+		} else {
+			h.batch.entries[idx] = h.batch.entries[n-1]
+			h.batch.indexes[string(h.batch.entries[idx].Name)] = idx
```
SandyXSD, line 3050: When `idx == n-1`, there will be one invalid entry left in the `indexes` map.

pkg/meta/tkv.go
```diff
@@ -3756,3 +3780,89 @@ func (m *kvMeta) loadDumpedACLs(ctx Context) error {
...
+			initEntries: entries,
+			fetcher:     m.getDirFetcher(),
+			batchNum:    DirBatchNum["kv"],
+		},
+	}
+	s.batch, _ = s.fetch(Background, 0)
+	return s
+}
+
+func (m *kvMeta) getDirFetcher() dirFetcher {
```
SandyXSD, line 3808: define a fetcher function directly.

pkg/meta/tkv.go
```diff
@@ -3756,3 +3780,89 @@ func (m *kvMeta) loadDumpedACLs(ctx Context) error {
...
+	s := &kvDirHandler{
+		dirHandler: dirHandler{
+			inode:       inode,
+			plus:        plus,
+			initEntries: entries,
+			fetcher:     m.getDirFetcher(),
+			batchNum:    DirBatchNum["kv"],
+		},
+	}
+	s.batch, _ = s.fetch(Background, 0)
```
SandyXSD, line 3804: pass the `ctx` instead of `Background`? And the error should be returned?

jiefenghuang: error can be ignore, will retry in the following readdir

pkg/meta/tkv.go
jiefenghuang: we need *kvMeta in dirFetcher and used in base dirHandler.

SandyXSD (approved):

pkg/meta/sql.go
```diff
@@ -4446,3 +4446,94 @@ func (m *dbMeta) loadDumpedACLs(ctx Context) error {
...
+			if cursor != nil {
+				iCursor = cursor.(int)
+			}
+
+			var ids []int64
+			if err := s.Table(&edge{}).Cols("id").Where("parent = ?", inode).Limit(limit, iCursor).Find(&ids); err != nil {
+				return err
+			}
+
+			s = s.Table(&edge{}).In("jfs_edge.id", ids).OrderBy("jfs_edge.name")
```
SandyXSD, line 4504: These two queries for `edge` table can probably be merged as one.

SandyXSD: And ordered by `id` should be a simpler way.

jiefenghuang: Need to sort by name, otherwise in some cases such as reverse order, the results may be incorrect

jiefenghuang: The SQL query is splited which allows it to directly hit the index tree and reduce the amount of querying. 
Tried to merge statements in xorm, which would lose optimization effects.

A new pull request can be proposed for optimization.

Issue #5166, opened 2024-09-12
Fatal error and cannot view logs


Haivilo:
We keep receiving 
```
juicefs[19] <FATAL>: The mount point is not ready in 10 seconds, please check the log (/var/log/juicefs.log) or re-mount in foreground [mount_unix.go:57]
``` 
but since the log is in the container, we cannot check it when the pod is deleted. Is there a way to find the root cause?  
![Screenshot 2024-09-12 at 12 02 51 PM](https://github.com/user-attachments/assets/73f089f6-9db7-4b5b-b375-e4beb497ab76)

**Version**:
Helm Chart version: 0.19.4 (juicefs-csi-driver)
Juicefs version: v1.1.0

Haivilo:
cc @yetone @bojiang

zhijian-pro:
It is best to reproduce the error and then collect the log before deleting the pod

polyrabbit:
@Haivilo We also had similar issues occasionally, and the log does not provide any clues. We patched the code to log the stack trace of the hanging mount process, so that we can see where it happens. Here is a recent stack trace:

```
juicefs[1] <FATAL>: The mount point is not ready in 10 seconds (tried to kill mount process 57), exit it [mount_unix.go:237]
SIGABRT: abort
PC=0x40998e m=0 sigcode=0

goroutine 0 gp=0x5916760 m=0 mp=0x5917780 [idle]:
runtime/internal/syscall.Syscall6()
	/usr/local/go/src/runtime/internal/syscall/asm_linux_amd64.s:36 +0xe fp=0x7fff8f879f38 sp=0x7fff8f879f30 pc=0x40998e
runtime/internal/syscall.EpollWait(0x5?, {0x7fff8f879fd4?, 0x100000010?, 0x1?}, 0x9bea3b2c?, 0x9bea3b2c?)
	/usr/local/go/src/runtime/internal/syscall/syscall_linux.go:55 +0x45 fp=0x7fff8f879f88 sp=0x7fff8f879f38 pc=0x409905
runtime.netpoll(0xc00006e008?)
	/usr/local/go/src/runtime/netpoll_epoll.go:118 +0xd2 fp=0x7fff8f87a618 sp=0x7fff8f879f88 pc=0x43df72
runtime.findRunnable()
	/usr/local/go/src/runtime/proc.go:3470 +0x8c5 fp=0x7fff8f87a790 sp=0x7fff8f87a618 pc=0x44a725
runtime.schedule()
[...]

PR #5171, opened 2024-09-18
Kill mount process and dump its stack when mount point is not ready in time

polyrabbit:
Occationally, the mount process hangs and we only get the following error message - it provides no clue as to why.

```
<FATAL>: The mount point is not ready in 10 seconds ...
```

We added a patch to the `checkMountpoint` function, which dumps stack trace of the mount process and **KILLs** it when it's not ready in time. This patch helps us to spot several environmental issues that may be causing the mount to hang.

Issue #5166 shows a similar issue. So I think this patch should also benefit other users in the community.

### DIFF
--- a/cmd/mount_unix.go
+++ b/cmd/mount_unix.go
@@ -53,6 +53,8 @@ import (
 	"github.com/juicedata/juicefs/pkg/vfs"
 )
 
+var mountPid int
+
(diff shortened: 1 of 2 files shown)


pleirb:
In my case, it was the very new kernel version in CachyOS; switching to LTS (which is more stable) fixed the issue.

PR #5172, opened 2024-09-18
cmd/mount: rename the status path after the loading is successful

zhijian-pro:


Xuyuchao-juice:
Test Step：
1. Run hot update
2. Check bak file exist or not.

Test result:
image

PR #5173, opened 2024-09-18
cleanup duplicated code

davies:


jiefenghuang:
The original check was to allow other requests to quickly skip over (before cache.Lock) and read from object storage when an error occurs (e.g., the disk hangs). This modification will not achieve that purpose.

PR #3194, opened 2023-01-23
disable server name check for sentinel

davies:
close #3193

PR #5182, opened 2024-09-23
fix(redis): supports pass the tls server name for redis tls connection

yetone:
Because connecting to Azure Cache requires a [TLS server name](https://github.com/redis/go-redis/issues/1698), I'm not sure why this [PR](https://github.com/juicedata/juicefs/pull/3194/files) leaves the TLS server name empty. To avoid making disruptive updates, I added a separate query for users to specify the TLS server name.


SandyXSD:
In Redis sentinel mode, each sentinel node has its own name. Leaving the `ServerName` field empty allows the system to automatically detect the node name.

PR #5192, opened 2024-09-25
Fix(openfile): avoid returning stale slice caches

polyrabbit:
Some users cache opened files for a long time for performance reasons, which causes them read stale slices from `openfiles` cache - we only check expiration when opening a file, not reading slices.

One scenario is one client opens a file and reads occasionally - the read slices are cached forever. Another client compacts the file and old slices are deleted after `trash-days`. Now when the first client reads the file again, it gets stale slices. The penalty of false positive cache is one-second delay of a successful read. We could avoid this by applying `expire` to slices cache as well.

https://github.com/juicedata/juicefs/blob/ad6cf901579d65d89075619f1116069dabe2bdd3/pkg/chunk/cached_store.go#L711-L720

### DIFF
--- a/pkg/meta/openfile.go
+++ b/pkg/meta/openfile.go
@@ -199,6 +199,10 @@ func (o *openfiles) ReadChunk(ino Ino, indx uint32) ([]Slice, bool) {
 	if !ok {
 		return nil, false
 	}
+	if time.Since(time.Unix(of.lastCheck, 0)) >= o.expire {
+		of.invalidateChunk()
+		return nil, false
+	}
 	if indx == 0 {
 		return of.first, of.first != nil
 	} else {
@@ -222,6 +226,7 @@ func (o *openfiles) CacheChunk(ino Ino, indx uint32, cs []Slice) {
 		}
 		of.chunks[indx] = cs
 	}
+	of.lastCheck = time.Now().Unix()
 }
 
 func (o *openfiles) InvalidateChunk(ino Ino, indx uint32) {


pkg/meta/openfile.go
```diff
@@ -222,6 +226,7 @@ func (o *openfiles) CacheChunk(ino Ino, indx uint32, cs []Slice) {
 		}
 		of.chunks[indx] = cs
 	}
+	of.lastCheck = time.Now().Unix()
```
SandyXSD, line 229: The `lastCheck` is used to indicate the check time of attribute, so should not be updated here.

pkg/meta/openfile.go
```diff
@@ -199,6 +199,10 @@ func (o *openfiles) ReadChunk(ino Ino, indx uint32) ([]Slice, bool) {
 	if !ok {
 		return nil, false
 	}
+	if time.Since(time.Unix(of.lastCheck, 0)) >= o.expire {
```
SandyXSD, line 202: The chunk cache should also work when `open-cache=0`.

polyrabbit: Oh, I thought this line has a bug, not realizing it's a feature...
https://github.com/juicedata/juicefs/blob/ad6cf901579d65d89075619f1116069dabe2bdd3/pkg/meta/base.go#L1459

But if we cache chunks even when `open-cache` is disabled, I suppose we are breaking the close-to-open semantics? The old cache remains even when user reopens the file - he gets a fresh attr but stale data?

polyrabbit:
@SandyXSD Do you think chunk cache should also obey the `open-cache expire` flag? If not, I'm OK to close this PR and keep it privately.

pkg/meta/openfile.go
davies: No, the chunk cache is invalidated in `Update`

davies:
The current behavior is 1 seconds delay in the case of stale slice (rarely), so it's better to remove the retry or retry without delay.

pkg/meta/openfile.go
polyrabbit: Got it. It's updated according to mtime.

polyrabbit:
Agreed, I suppose it's better to remove the inner retry. The current read retry is twice as user expected.

polyrabbit:
Taken over by #5199 as discussed.

PR #5200, opened 2024-09-27
Fix warning log when creating a new cache dir

polyrabbit:
After updating to the latest master branch, a lot of warning logs show up (see attached screenshot). 

This happens because `dir` is appended with "/" in the CacheManager (I don't know why), and when calling `filepath.Dir` on a string ending with "/", the function returns the same dir (with "/" trimed). So same dir is passed to `os.Mkdir` twice, and the second call will raise a warning log. This may cause confusion to users.

image

### DIFF
--- a/pkg/chunk/disk_cache.go
+++ b/pkg/chunk/disk_cache.go
@@ -496,17 +496,12 @@ func (cache *cacheStore) createDir(dir string) {
 		mode := cache.mode | (readmode >> 2) | (readmode >> 1)
 		var st os.FileInfo
 		var err error
+		dir = filepath.Clean(dir)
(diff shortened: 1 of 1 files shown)


jiefenghuang (approved):

davies:
we should ignore the EEXISTS from Mkdir (no warning)

polyrabbit:
We have os.Stat check in the beginning, and only create if os.Stat says it doesn't exist. So I suppose the EEXISTS error is a unexpected behavior?

davies:
It's expected that more than one goroutine will call createDir(), so one fail with EEXISTS is expected.

polyrabbit:
OK, get it - it has concurrent access when writing caches.

PR #5202, opened 2024-09-29
Add metrics for current cache/object readers

polyrabbit:
Align with `object_request_uploading`, so we know the pressure we are exposing to the backend storage.

### DIFF
--- a/pkg/chunk/cached_store.go
+++ b/pkg/chunk/cached_store.go
@@ -27,6 +27,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/juicedata/juicefs/pkg/compress"
@@ -40,9 +41,7 @@ const chunkSize = 1 << 26 // 64M
 const pageSize = 1 << 16  // 64K
 const SlowRequest = time.Second * time.Duration(10)
 
-var (
-	logger = utils.GetLogger("juicefs")
-)
+var logger = utils.GetLogger("juicefs")
 
 type pendingItem struct {
 	key       string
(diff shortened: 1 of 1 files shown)


pkg/chunk/cached_store.go
```diff
@@ -921,6 +926,20 @@ func (store *cachedStore) regMetrics(reg prometheus.Registerer) {
...
+		ConstLabels: prometheus.Labels{"storage": "cache"},
+	}, func() float64 {
+		return float64(store.cacheReaders.Load())
+	}))
+	reg.MustRegister(prometheus.NewGaugeFunc(prometheus.GaugeOpts{
+		Name:        "store_readers",
+		Help:        "number of storage readers",
+		ConstLabels: prometheus.Labels{"storage": "object"},
+	}, func() float64 {
+		return float64(store.objectReaders.Load())
```
davies, line 941: We can add a limit for concurrent GET, and know the actual concurrency using that.

pkg/chunk/cached_store.go
```diff
@@ -921,6 +926,20 @@ func (store *cachedStore) regMetrics(reg prometheus.Registerer) {
 		func() float64 {
 			return float64(len(store.currentUpload))
 		}))
+	reg.MustRegister(prometheus.NewGaugeFunc(prometheus.GaugeOpts{
+		Name:        "store_readers",
+		Help:        "number of storage readers",
+		ConstLabels: prometheus.Labels{"storage": "cache"},
+	}, func() float64 {
+		return float64(store.cacheReaders.Load())
```
davies, line 934: We can know the concurrent IO from `dcState`

polyrabbit: `concurrency` counter is only updated in `unstableDC` implementation. To track concurrency in other states, I have to move it to `baseDC`?

pkg/chunk/cached_store.go
polyrabbit: Currently we don't have that limitation right?

davies:
We usually use the rate of object_request_durations_histogram_seconds as the metric of concurrency, no new metric needed.

Same to other operations or requests.

polyrabbit:
OK, closing

PR #5207, opened 2024-10-08
fix: limit dir batch num

jiefenghuang:


pkg/meta/sql.go
```diff
@@ -4488,6 +4488,9 @@ func (h *dbDirHandler) Delete(name string) {
 }
 
 func (m *dbMeta) newDirHandler(inode Ino, plus bool, entries []*Entry) DirHandler {
+	if m.Name() == "sqlite3" {
```
SandyXSD, line 4491: move it to `newSQLMeta`

PR #5216, opened 2024-10-10
disk-cache: keep a margin when evicting disk cache according to the free-space-ratio

ddrid:
Retain a margin when evicting disk cache according to the free-space-ratio to ensure we have enough free space after cleanup, just the same as the mechanism when evicting by cache-size.

### DIFF
--- a/pkg/chunk/disk_cache.go
+++ b/pkg/chunk/disk_cache.go
@@ -715,7 +715,7 @@ func (cache *cacheStore) cleanupFull() {
 		if toFree > cache.used {
 			goal = 0
 		} else if cache.used-toFree < goal {
-			goal = cache.used - toFree
+			goal = (cache.used - toFree) * 95 / 100
 		}
 	}
 	if fr < cache.freeRatio {
@@ -724,7 +724,7 @@ func (cache *cacheStore) cleanupFull() {
 		if toFree > len(cache.keys) {
 			num = 0
 		} else {
-			num = len(cache.keys) - toFree
+			num = (len(cache.keys) - toFree) * 99 / 100
 		}
 	}
 


jiefenghuang:
i think line 708 and 709 have already guaranteed the minimum release space.

ddrid:
> i think line 708 and 709 have already guaranteed the minimum release space.

Hi @jiefenghuang, 
We have user tend to set a very large --cache-size, much larger than the disk capacity, so it's necessary to use the --free-space-ratio condition as a fallback. Adding this margin here helps avoid some issues.
Take line 708 as an example, `goal := cache.capacity * 95 / 100` is calculated based on --cache-size parameter, while the modification in this commit, affects the logic within `if br < cache.freeRatio` (line 712), which is calculated based on --free-space-ratio. We do not have 95% margin here.

ddrid:
@jiefenghuang For example, suppose `cache.capacity` is set to 1000GB and `cache.freeRatio` is set to 0.2. The maximum capacity of the disk where the cache directory is located is 100GB, and currently, 85GB is used.

In line 708, the calculated goal is 1000GB * 0.95 = 950GB. After entering the logic of line 712, the calculated toFree is 100GB * (0.2 - 0.15) = 5GB. So, `cache.used - toFree` is 85GB - 5GB = 80GB, which is much less than the previously calculated goal of 950GB. Therefore, it can enter the if statement at line 717, where goal will be overwritten to 80GB.

jiefenghuang:
> @jiefenghuang For example, suppose `cache.capacity` is set to 1000GB and `cache.freeRatio` is set to 0.2. The maximum capacity of the disk where the cache directory is located is 100GB, and currently, 85GB is used.
> 
> In line 708, the calculated `goal` is 1000GB * 0.95 = 950GB. After entering the logic of line 712, the calculated `toFree` is 100GB * (0.2 - 0.15) = 5GB. So, `cache.used - toFree` is 85GB - 5GB = 80GB, which is much less than the previously calculated `goal` of 950GB. Therefore, it can enter the if statement at line 717, where `goal` will be overwritten to 80GB.

@ddrid Finally releasing 5GiB and retaining 80GiB, doesn't that conform to the expectation of freeRatio=0.2?

ddrid:
> > @jiefenghuang For example, suppose `cache.capacity` is set to 1000GB and `cache.freeRatio` is set to 0.2. The maximum capacity of the disk where the cache directory is located is 100GB, and currently, 85GB is used.
> > In line 708, the calculated `goal` is 1000GB * 0.95 = 950GB. After entering the logic of line 712, the calculated `toFree` is 100GB * (0.2 - 0.15) = 5GB. So, `cache.used - toFree` is 85GB - 5GB = 80GB, which is much less than the previously calculated `goal` of 950GB. Therefore, it can enter the if statement at line 717, where `goal` will be overwritten to 80GB.
> 
> @ddrid Finally releasing 5GiB and retaining 80GiB, doesn't that conform to the expectation of freeRatio=0.2?

@jiefenghuang If --cache-size is set to 100GB, then after each cleanup, 100 * 0.95 = 95GB will remain, leaving a 5GB margin, as referenced in line 708. The logic for --free-space-ratio should also follow this margin. Otherwise, when multiple clients use the same cache disk, it can easily cause conflicts and mutual interference.

jiefenghuang:
@ddrid  So you are trying to solve the issue of 'Otherwise, when multiple clients use the same cache disk, it can easily cause conflicts and mutual interference.'? 

Different clients can use different configurations without correctness and performance issues, right? Also, currently, the cache is mainly used for local caching, using a cloud disk as a distributed cache is not highly recommended.

ddrid:
> @ddrid So you are trying to solve the issue of 'Otherwise, when multiple clients use the same cache disk, it can easily cause conflicts and mutual interference.'?
> 
> Different clients can use different configurations without correctness and performance issues, right? Also, currently, the cache is mainly used for local caching, using a cloud disk as a distributed cache is not highly recommended.

@jiefenghuang 
1. Multiple users might mount local caches on the same disk simultaneously because a single machine might have multiple training resources. Each user might only use one card, so it's common for multiple users to share the same machine. This cache disk is not a cloud disk but a local physical machine disk. 
2. Why can't we limit the cache size for each client? Because we want each client to utilize the disk resources to the fullest. For example, if there are eight users on one disk, each can use a portion, but if there’s only one user, they can automatically use 100% of the disk cache.

jiefenghuang:
@ddrid This modification does not cover all cases, so I'm temporarily closing this PR. We can discuss your scenarios in the discussions, or through Slack or WeCom.

davies:
LGTM

PR #5219, opened 2024-10-10
Dedup prefetch/partial read with full block read to reduce read amplification

polyrabbit:
Prefetch/partial/full read are three ways to read from oss. In current implementation, they may overlap with each other. On sequential read when readahead is triggered, the overlap is trivial. But on other cases, the overlap may cost extra network bandwidth.

This PR groups them in a singleflight to avoid the read amplification.

image


jiefenghuang:
LGTM

Issue #5224, opened 2024-10-12
Make max-readahead configurable to enhance read performance


ddrid:
Currently, readAheadMax is a non-configurable parameter for fuse users, set by default to conf.Chunk.BlockSize * 8, typically 32MB. This isn't sufficient for low-concurrency read requests. I intend to make max-readahead a user-configurable parameter.

After testing, I set max-readahead to 512MB and observed that under single-threaded, with uncached local disk reads, the throughput increased from 230MB/s to 2.2GB/s, showing a 10x improvement.

solracsf:
Thanks for the PR. Maybe also document it (in docs) too?

davies:
we can provide `--max-readahead` to overwrite this default setting.

PR #5225, opened 2024-10-12
cmd/mount: add `max-readahead` option to set the max buffer of each readahead

ddrid:
#5224

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -174,7 +174,6 @@ func dataCacheFlags() []cli.Flag {
 		},
 		&cli.StringFlag{
 			Name:  "max-readahead",
-			Value: "32M",
 			Usage: "max buffering for read ahead in MiB",
 		},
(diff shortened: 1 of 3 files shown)


cmd/flags.go
```diff
@@ -172,6 +172,11 @@ func dataCacheFlags() []cli.Flag {
 			Value: "300M",
 			Usage: "total read/write buffering in MiB",
 		},
+		&cli.StringFlag{
+			Name:  "max-readahead",
+			Value: "32M",
```
jiefenghuang, line 177: A 32MiB default setting isn't necessary here.
If no configuration is explicitly provided, the original configuration logic should remain unchanged, like "conf.Chunk.BlockSize * 8"

ddrid: removed

jiefenghuang (approved):
LGTM

PR #5229, opened 2024-10-13
Allow user to disable readahead by setting `max-readahead` smaller than blocksize

polyrabbit:
Allow user to disable read-ahead when he is doing random read - to reduce read amplification and improve read latency.

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -174,7 +174,7 @@ func dataCacheFlags() []cli.Flag {
 		},
 		&cli.StringFlag{
 			Name:  "max-readahead",
-			Usage: "max buffering for read ahead in MiB",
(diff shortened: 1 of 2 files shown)


jiefenghuang:
Perhaps a more explicit flag would be better, such as -1 to disable readahead.

polyrabbit:
I suppose it's better to measure `max-readahead` in blocks, not MiB, in consistent with the code logic.

Current value of `max-readahead` results in inconsistent readahead blocks:
| max-readahead value            | readahead blocks                    |
|------------------|--------------------------|
| \[blockSize, +inf) | max-readahead/blockSize |
| (0, blockSize)   | 0                        |
| \[0]             | 8                        |

Issue #5234, opened 2024-10-17
juicefs sync not support UCloud multipart upload


chnliyong:
**What happened**:
Fail to sync a 300MiB+ file from local disk to UCloud storage:
```
2024/10/17 11:53:05.025863 juicefs[400555] <ERROR>: Failed to copy data of  in 8.856962ms: object size 371246902 is too large to copy [sync.go:612]
2024/10/17 11:53:05.026853 juicefs[400555] <ERROR>: Failed to copy object : object size 371246902 is too large to copy [sync.go:700]
2024/10/17 11:53:05.037176 juicefs[400555] <INFO>: Found: 1, skipped: 0 (0 B), copied: 0 (0 B), failed: 1 [sync.go:1359]
```

**What you expected to happen**:
Copied Successfully.

**How to reproduce it (as minimally and precisely as possible)**:
juicefs sync <300MiB large file> ufile://....

**Anything else we need to know?**
The problem is here: https://github.com/juicedata/juicefs/blob/fd191eadc403a30441a3aee291a8c8622b2aa32a/pkg/sync/sync.go#L549-L551
The UFile US3 support multipart upload but not define the Limits: https://github.com/juicedata/juicefs/blob/fd191eadc403a30441a3aee291a8c8622b2aa32a/pkg/object/interface.go#L66-L72

When I add the Limits override in pkg/object/ufile.go
```go
func (u *ufile) Limits() Limits {
        return Limits{
                IsSupportMultipartUpload: true,
                IsSupportUploadPartCopy:  true,
                MinPartSize:              4 << 20, 
                MaxPartSize:              100 << 20, 
                MaxPartCount:             1000000,
        }
}
```
then the sync success:
```
2024/10/17 12:58:38.459811 juicefs[404193] <DEBUG>: Copying data of  as 89 parts (size: 4194304): abca6333-8523-470c-b734-ae58b1d3d111 [sync.go:456]
2024/10/17 12:58:38.556916 juicefs[404193] <DEBUG>: Copied data of  part 46 [sync.go:512]
2024/10/17 12:58:38.569169 juicefs[404193] <DEBUG>: Copied data of  part 1 [sync.go:512]
2024/10/17 12:58:38.570971 juicefs[404193] <DEBUG>: Copied data of  part 44 [sync.go:512]
2024/10/17 12:58:38.577632 juicefs[404193] <DEBUG>: Copied data of  part 47 [sync.go:512]
[...]

PR #5236, opened 2024-10-18
object/ufile: implement limits interface  for us3 to support the multipart upload

zhijian-pro:
image

close https://github.com/juicedata/juicefs/issues/5234


jiefenghuang (approved):
LGTM

Issue #5206, opened 2024-10-05
Problem with usage of juicedata/mount docker image


halsbox:
**What happened**:  
Juicefs is not properly unmounted when stopping docker container created as documented in `juicefs_on_docker#method-2-use-the-officially-maintained-image`

**What you expected to happen**: 
Juicefs is unmounted on container stop

**How to reproduce it (as minimally and precisely as possible)**: 
Follow `juicefs_on_docker#method-2-use-the-officially-maintained-image` documentation to create and start a container, then stop container. Run `stat` or `ls` on directory that was bind-mounted in the container to get `Transport endpoint is not connected` error. Unmount the directory manually.

**Anything else we need to know?**
`juicefs mount` command starts two processes, of which only the second one is listening for signals (SIGTERM, etc) and only the second process does graceful stop with unmount when receiving a signal from docker engine (or anything else). When run in docker container as described in docs, the signal is sent to the first process with PID 1 and container is just killed without graceful unmount. The bug can be reproduced without using containers: just run `juicefs mount` in foreground, then go to another terminal and do `ps auxww | grep juicefs`, you'll get two processes with their PIDs. Run `kill -SIGTERM` with first PID as argument.

**Environment**:
- JuiceFS version (use `juicefs --version`) or Hadoop Java SDK version: 1.2.1+2024-08-30.cd871d19
- OS (e.g `cat /etc/os-release`): Debian GNU/Linux 12 (bookworm)
- Kernel (e.g. `uname -a`): 6.1.0-26-amd64 SMP PREEMPT_DYNAMIC Debian 6.1.112-1 (2024-09-30) x86_64 GNU/Linux

halsbox:
My ugly workaround is running `juicefs mount` command in shell with signal trap, e.g. in docker compose:
```
command: sh -c "trap 'test -e /fs/.config && umount /fs' TERM INT HUP; juicefs mount redis://redis:6379/1 /fs & p=$$!; wait $$p"
```
instead of:
```
command: ["juicefs", "mount", "redis://redis:6379/1", "/fs"]
```

halsbox:
I looked in source code and made a little patch against main branch to solve the problem:
```
diff --git a/cmd/mount_unix.go b/cmd/mount_unix.go
index 2e60e8d1..f43844c6 100644
--- a/cmd/mount_unix.go
+++ b/cmd/mount_unix.go
@@ -780,6 +780,16 @@ func launchMount(mp string, conf *vfs.Config) error {
[...]

PR #5238, opened 2024-10-21
cmd/mount: pass the signal to the child process

zhijian-pro:
Thanks @halsbox for the feedback

close https://github.com/juicedata/juicefs/issues/5206


jiefenghuang (approved):
LGTM

PR #5241, opened 2024-10-21
feat: Add ability to configure client TLS

achetronic:
Hello there! :)

There is no issue for this PR yet (I think I will create it)

This PR adds support for TLS client certificates at transport level. It's intended for those S3 endpoints that are behind mTLS protection. This has been already tested in our endpoints and it's working well

Screenshot from 2024-10-21 13-46-01

I have added those params as flags under `format` command as this is more expectable by the user. 
I saw you stored some of them as query params, for example `tls-insecure-skip-verify` but equally they are stored in metadata settings

WDYT?

CC: @davies @SandyXSD @zhijian-pro I mention you all as you are the top contributors, so I dunno who to invoke

### DIFF
--- a/cmd/format.go
+++ b/cmd/format.go
@@ -238,30 +238,15 @@ func createStorage(format meta.Format) (object.ObjectStorage, error) {
 		}
 
 		// Configure client TLS when params are provided
-		caCerts := values.Get("ca-certs")
-		if os.Getenv("JFS_CA_CERTS") != "" {
-			caCerts = os.Getenv("JFS_CA_CERTS")
-		}
+		if values.Get("ca-certs") != "" && values.Get("ssl-cert") != "" && values.Get("ssl-key") != "" {
 
-		sslCert := values.Get("ssl-cert")
-		if os.Getenv("JFS_SSL_CERT") != "" {
-			sslCert = os.Getenv("JFS_SSL_CERT")
-		}
-
-		sslKey := values.Get("ssl-key")
-		if os.Getenv("JFS_SSL_KEY") != "" {
-			sslKey = os.Getenv("JFS_SSL_KEY")
-		}
-
-		if caCerts != "" && sslCert != "" && sslKey != "" {
-
-			clientTLSCert, err := tls.LoadX509KeyPair(sslCert, sslKey)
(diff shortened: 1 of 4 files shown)


ivanprado:
This definitely would be useful for me. 👍

cmd/format.go
```diff
@@ -140,6 +141,18 @@ func formatStorageFlags() []cli.Flag {
 			Name:  "storage-class",
 			Usage: "the default storage class",
 		},
+		&cli.StringFlag{
+			Name:  "ca-certs",
```
davies, line 145: I think we could support these just like `insecure-skip-verify`, as a url query of bucket endpoint

achetronic: Well, I added them as flags to show them in the console in the same way most CLI out there do. But my mind is divided as they are not format-related params, they are transport-related

I checked the metadata store and i saw query params are there too, so if you prefer, I can parse them as query params, of course. If you confirm this, i will make proper changes, of course :)

davies: yes, we prefer query params

achetronic: What about current approach? :)

cmd/format.go
```diff
@@ -234,6 +236,43 @@ func createStorage(format meta.Format) (object.ObjectStorage, error) {
 			u.RawQuery = values.Encode()
 			format.Bucket = u.String()
 		}
+
+		// Configure client TLS when params are provided
+		caCerts := values.Get("ca-certs")
+		if os.Getenv("JFS_CA_CERTS") != "" {
```
davies, line 242: Since the ca could be used for meta-url or bucket, then `JFS_CA_CERTS` will be confusing, so it's better to don't support that.

achetronic: Would you like me to rename variables to JFS_BUCKET_ or you prefer to remove env vars at all? :)

I added them as an option, but i'm ok with query params only too

davies: It's better to remove them

achetronic: Done! :)

achetronic:
@davies Fixed

davies:
lgtm, please update the docs for these parameters, thanks!

PR #5251, opened 2024-10-23
Fix warmup only has one writer regardless of the number of `--threads` specified

polyrabbit:
`warmup` is very slow regardless of the number of concurrent workers. It is because `warmup` has many readers but only one writer - the cache flusher.

So instead of waiting for the single writer to drain the queue, we can write directly and concurrently to the cache. I suppose this is the smallest change to make `warmup` obey the `--threads` flag.

### DIFF
--- a/pkg/chunk/disk_cache.go
+++ b/pkg/chunk/disk_cache.go
@@ -407,14 +407,14 @@ func (cache *cacheStore) cache(key string, p *Page, force, dropCache bool) {
 	p.Acquire()
 	cache.pages[key] = p
 	atomic.AddInt64(&cache.totalPages, int64(cap(p.Data)))
-	select {
-	case cache.pending <- pendingFile{key, p, dropCache}:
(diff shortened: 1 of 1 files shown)


davies:
Usually the single flush thread should not be bottleneck, because it only write into page cache.

In the same time, there could be some read traffic on the same disk, so we don't want to use all the IPOS for writing cache.

polyrabbit:
Under normal circumstances, writing cache is still single-threaded - it has `force` flag set to `false`.

This change only affects behaviors of prefetch and warmup - so they will obey their own `--prefetch` and `--threads` flags.

davies:
The `threads` means the number of threads to down from object storage, so it's still meaningful.

davies:
The current behavior is friendly to production workload, because it will not easily throttle the disk bandwidth.

polyrabbit:
OK, got it

PR #5257, opened 2024-10-28
gateway: set xattr before rename to ensure consistency

zhijian-pro:


jiefenghuang (approved):

zhaohuxing:
Does CompleteMultipartUpload also need to set xattr before rename?

PR #5261, opened 2024-10-29
Support cache blocks after uploading

polyrabbit:
Benefit scenario: read the most recently written files. One good use case is save/load checkpoint.

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -219,8 +219,8 @@ func dataCacheFlags() []cli.Flag {
 			Usage: "cache only random/small read",
 		},
 		&cli.BoolFlag{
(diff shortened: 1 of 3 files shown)


cmd/flags.go
```diff
@@ -218,6 +218,10 @@ func dataCacheFlags() []cli.Flag {
 			Name:  "cache-partial-only",
 			Usage: "cache only random/small read",
 		},
+		&cli.BoolFlag{
+			Name:  "cache-on-write",
```
davies, line 222: cache-large-write

polyrabbit: OK, updated.

Issue #5268, opened 2024-11-01
Initialize staging_write_bytes metric as Counter for cumulative tracking


Looka149:
The `staging_write_bytes` metric is intended to be used as a Counter, but it is currently represented as a Gauge. 

```
# HELP juicefs_staging_write_bytes write bytes of blocks in the staging path.
# TYPE juicefs_staging_write_bytes gauge
```

<img width="849" alt="image" src="https://github.com/user-attachments/assets/541f9edf-8694-4fdb-b2f9-29021e1d8c39">

PR #5269, opened 2024-11-01
fix: initialize staging_write_bytes metric as Counter instead of Gauge

Looka149:
The variable for the `staging_write_bytes` metric is intended to be `prometheus.Counter`, but the code uses Gauge.

https://github.com/juicedata/juicefs/blob/9ddc6d422d1511b0d861849535afcb286bd1af05/pkg/chunk/metrics.go#L32

https://github.com/juicedata/juicefs/blob/9ddc6d422d1511b0d861849535afcb286bd1af05/pkg/chunk/metrics.go#L91-L94

close https://github.com/juicedata/juicefs/issues/5268

### DIFF
--- a/pkg/chunk/metrics.go
+++ b/pkg/chunk/metrics.go
@@ -88,7 +88,7 @@ func (c *cacheManagerMetrics) initMetrics() {
 		Name: "staging_block_bytes",
 		Help: "Total bytes of blocks in the staging path.",
 	})
-	c.stageWriteBytes = prometheus.NewGauge(prometheus.GaugeOpts{
+	c.stageWriteBytes = prometheus.NewCounter(prometheus.CounterOpts{
 		Name: "staging_write_bytes",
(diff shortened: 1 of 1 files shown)


zhijian-pro:
https://github.com/juicedata/juicefs/blob/9ddc6d422d1511b0d861849535afcb286bd1af05/pkg/chunk/metrics.go#L32
The expected type is gauge
```
 stageWriteBytes prometheus.Gauge
```

Looka149:
@zhijian-pro 
Thanks for the review! I understood this as a `Counter` since `stageWriteBytes.Add` is the only method used throughout the code. If it’s meant to be a `Gauge`, should `.Sub` be called somewhere? Let me know if there's something I missed.

zhijian-pro:
Sorry, I mistook it for  `staging_writing_blocks`

zhijian-pro (approved):

PR #5272, opened 2024-11-03
Fix gc tool never updates volume stats

polyrabbit:
Follow-up to PR #4979 , the `gc` tool never updates volume stats, even when files are deleted (e.g. run with `--delete`). So after several runs, volume will have a larger usage than expected.


davies:
What's kinds of stats are missing in gc? it will not delete any files.

polyrabbit:
The volume usage. GC tool cleans up trash files if `--delete` is specified. But this cleanup does not update volume usage to kv/sql engine, so user will get a larger usage than actual. This issue happens a lot in our environment, and I reported before.

PR #5281, opened 2024-11-06
Perform object storage test at startup

polyrabbit:
Test object storage at startup to fail fast instead of throwing `EIO` in the middle of user's workload.

### DIFF
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -644,7 +644,7 @@ func mount(c *cli.Context) error {
 		}
 		if blob != nil {
 			// test storage at startup to fail fast instead of throwing EIO in the middle of user's workload
-			if !metaConf.ReadOnly && !metaConf.NoBGJob {
+			if c.Bool("check-storage") && !metaConf.ReadOnly && !metaConf.NoBGJob {
 				start := time.Now()
 				if err = test(blob); err != nil {
 					logger.Errorf("Object storage test failed: %s", err)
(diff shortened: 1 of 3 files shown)


davies:
It's slow to test, which will slow down the mount.

The AK/SK is verified in format/config, which should be enough.

polyrabbit:
`format` only verifies the AK/SK but cannot check user's environment. A typical scenario is when a node cannot connect to OSS for various reasons, such as DNS issues or VPC network isolation problems. In such cases, users may encounter `EIO` errors in the middle of his task.
If we can perform some cheap self-checks at startup to expose errors early, we can avoid expensive user-side errors.

davies:
It could be useful in some cases, should be enabled by a flag.

polyrabbit:
Do you have a suggestion for this flag's name?

davies:
`--check-storage`?

polyrabbit:
Updated. Please review, thanks!

cmd/mount.go
```diff
@@ -643,6 +643,16 @@ func mount(c *cli.Context) error {
 			}
 		}
 		if blob != nil {
+			// test storage at startup to fail fast instead of throwing EIO in the middle of user's workload
+			if c.Bool("check-storage") && !metaConf.ReadOnly && !metaConf.NoBGJob {
```
davies, line 647: we can remove `metaConf.NoBGJob`

Also, we should test Get for read-only

polyrabbit: How about just removing those two checks `!metaConf.ReadOnly && !metaConf.NoBGJob`?

In `doTesting`, `store.Get` depends on `store.Put`, and `store.Delete` is always required (to do cleanup). It's not easy to just run one capability.

davies: Sounds good

polyrabbit: Updated

PR #5289, opened 2024-11-12
Better to close file explicitly

polyrabbit:
A minor fix

### DIFF
--- a/pkg/fuse/device_linux.go
+++ b/pkg/fuse/device_linux.go
@@ -103,6 +103,7 @@ func grantAccess() error {
 	if err != nil {
 		return errors.Wrapf(err, "open %s", deviceAllowPath)
 	}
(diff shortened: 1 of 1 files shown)


servusdei2018 (approved):
I would also note that there are two lines above which are unidiomatic:

```go
deviceListPath := path.Join("/sys/fs/cgroup/devices" + deviceCgroup + "/devices.list")
deviceAllowPath := "/sys/fs/cgroup/devices" + deviceCgroup + "/devices.allow"
```

should be

```go
deviceListPath := path.Join("/sys/fs/cgroup/devices", deviceCgroup, "/devices.list")
deviceAllowPath := path.Join("/sys/fs/cgroup/devices", deviceCgroup, "/devices.allow")
```

PR #5299, opened 2024-11-14
nit(cgroup): correct device path construction for cgroup management

servusdei2018:
This commit fixes incorrect device path construction in `cgroup` management. In the original code, [`path.Join`](https://pkg.go.dev/path#Join) was used incorrectly with a single concatenated string, rather than passing individual path components as separate arguments.


davies:
The behavior is different than before, have you check the correctness?

servusdei2018:
> The behavior is different than before, have you check the correctness?

Oh my bad, I misread it the first time. This should now preserve the original behavior

PR #5313, opened 2024-11-21
Memory cache also obeys `--cache-expire` flag

polyrabbit:
To release high memory usage caused by burst read/write.

### DIFF
--- a/pkg/chunk/mem_cache.go
+++ b/pkg/chunk/mem_cache.go
@@ -21,6 +21,8 @@ import (
 	"runtime"
 	"sync"
 	"time"
+
+	"github.com/dustin/go-humanize"
 )
 
 type memItem struct {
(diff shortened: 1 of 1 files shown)


pkg/chunk/mem_cache.go
```diff
@@ -146,6 +153,36 @@ func (c *memcache) cleanup() {
...
 
+func (c *memcache) cleanupExpire() {
+	var interval = time.Minute
+	if c.cacheExpire < time.Minute {
+		interval = c.cacheExpire
+	}
+	for {
+		var freed int64
+		var cnt, deleted int
+		cutoff := time.Now().Add(-c.cacheExpire)
```
jiefenghuang, line 164: put line 164 after Lock?

polyrabbit: OK, updated. Actually this patch is only a copy-paste of `disk_cache` version.

https://github.com/juicedata/juicefs/blob/2b006074f0db25843ded87cd5e2924031ff10af8/pkg/chunk/disk_cache.go#L333-L335

jiefenghuang (approved):

PR #5316, opened 2024-11-22
Add MariaDB to README.md

robertsilen:
Since you mention MariaDB in your documentation (https://juicefs.com/docs/community/databases_for_metadata/#mariadb), I suggested add MariaDB to your README too. Hope that is ok - no reason not to I guess? Cheers!

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
     <a href="https://go.juicefs.com/slack"><img alt="Join Slack" src="https://badgen.net/badge/Slack/Join%20JuiceFS/0abd59?icon=slack" /></a>
 </p>
 
-**JuiceFS** is a high-performance [POSIX](https://en.wikipedia.org/wiki/POSIX) file system released under Apache License 2.0, particularly designed for the cloud-native environment. The data, stored via JuiceFS, will be persisted in Object Storage _(e.g. Amazon S3)_, and the corresponding metadata can be persisted in various compatible database engines such as Redis, MySQL, and TiKV based on the scenarios and requirements.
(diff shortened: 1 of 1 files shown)


robertsilen:
> Since MariaDB is a fork of MySQL, it could be used just like MySQL, so we'd like to update the docs to reduce duplicated content.
> 
> There are other databases which are compatible to MySQL also, it's better to only say MySQL in the README to avoid fight between them, for example, #5318

As MariaDB is the default fork in most distros, many are using MariaDB instead of MySQL without even knowing it. Mentioning MariaDB compatibility removes doubts for the user and also makes projects aware of subtle differences in latest versions that are easily addressable. This is part of my efforts for the MariaDB Foundation, as explained in this blog: https://mariadb.org/improving-mariadb-support-in-open-source-projects/

davies:
Understood.

MySQL represents all the databases that are compatible to the protocol, including any forks. It is also the only one we use in the CI, so I think having MySQL in the README is enough, and other compatible ones  in the #database_for_metadata page should be enough.

ottok:
Note that most Linux distributions only ship MariaDB since almost a decade. There is no other "MySQL fork" alike, so I think it would make sense to be explicit about MariaDB support.

The picture in the front page README is pretty good indicator:
image

But still, I think Robert's suggestions would be a good addition.

JuiceFS should probably also consider having MariaDB in the CI. Some day in the future MySQL and MariaDB will diverge functionally and if JuiceFS has MariaDB in the CI, it will be easy to detect if there is something that no longer works identically on MySQL and MariaDB. Adding MariaDB now should be easy (just `apt install mariadb-server`, otherwise the same CI scripts).

PR #5317, opened 2024-11-22
Add timeout when reading cache to avoid IO hangs

polyrabbit:
Align with #5230.

Improve robustness of cache code - fallback to object storage when cache disk hangs.


jiefenghuang:
@polyrabbit  In low-performance instances with fewer cores,  introducing a new goroutine in the main read path significantly increases scheduling overhead, leading to performance degradation. This is also why we didn’t implement this timeout mechanism in disk fault tolerance.

PR #5323, opened 2024-11-26
Pre-create next trash entry

polyrabbit:
Occasionally, txn conflicts occurs when several clients create the next trash directory concurrently.

This can be avoided by prefetching the next trash directory, so we will get a stable and predictable `unlink` operation.

img_v3_02gt_db6cc09d-e12a-483f-acf3-6e604bde7e4g

### DIFF
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -2322,17 +2322,8 @@ func (m *baseMeta) checkTrash(parent Ino, trash *Ino) syscall.Errno {
 	}
 	m.Unlock()
 
-	st := m.en.doLookup(Background, TrashInode, name, trash, nil)
-	if st == syscall.ENOENT {
-		next, err := m.en.incrCounter("nextTrash", 1)
-		if err == nil {
-			*trash = TrashInode + Ino(next)
-			attr := Attr{Typ: TypeDirectory, Nlink: 2, Length: 4 << 10, Parent: TrashInode, Full: true}
-			st = m.en.doMknod(Background, TrashInode, name, TypeDirectory, 0555, 0, "", trash, &attr)
-		} else {
-			st = errno(err)
-		}
-	}
+	st := m.ensureTrash(name, trash)
+	go m.preCreateTrash()
 
 	m.Lock()
 	if st != 0 && st != syscall.EEXIST {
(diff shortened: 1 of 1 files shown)


pkg/meta/base.go
```diff
@@ -2322,17 +2322,8 @@ func (m *baseMeta) checkTrash(parent Ino, trash *Ino) syscall.Errno {
...
-		if err == nil {
-			*trash = TrashInode + Ino(next)
-			attr := Attr{Typ: TypeDirectory, Nlink: 2, Length: 4 << 10, Parent: TrashInode, Full: true}
-			st = m.en.doMknod(Background, TrashInode, name, TypeDirectory, 0555, 0, "", trash, &attr)
-		} else {
-			st = errno(err)
-		}
-	}
+	st := m.ensureTrash(name, trash)
+	go m.preCreateTrash()
```
davies, line 2326: This will double the call of of ensureTrash
I think we can call the preCreateTrash in the last minute of a hour.

We can also cache the trashInode for a hour

polyrabbit: Those two functions will only be called at most once per hour. Because current trash name is cached in memory, and most fetch requests will be intercepted by the cache.

https://github.com/juicedata/juicefs/blob/7b272d121794105fec74c1a6e2d816e801009785/pkg/meta/base.go#L2317-L2322

pkg/meta/base.go
```diff
@@ -2348,6 +2339,35 @@ func (m *baseMeta) checkTrash(parent Ino, trash *Ino) syscall.Errno {
...
+}
+
+func (m *baseMeta) ensureTrash(name string, trash *Ino) syscall.Errno {
+	st := m.en.doLookup(Background, TrashInode, name, trash, nil)
+	if st == syscall.ENOENT {
+		next, err := m.en.incrCounter("nextTrash", 1)
+		if err == nil {
+			*trash = TrashInode + Ino(next)
+			attr := Attr{Typ: TypeDirectory, Nlink: 2, Length: 4 << 10, Parent: TrashInode, Full: true}
+			st = m.en.doMknod(Background, TrashInode, name, TypeDirectory, 0555, 0, "", trash, &attr)
```
davies, line 2360: Maybe we can move nextTrash into doMknod() to avoid the conflict?

polyrabbit: Do you mean moving `next, err := m.en.incrCounter("nextTrash", 1)` into `m.en.doMknod`? then conflict will happen in `doMknod`, as it becomes a larger txn.

After applying the prefetch and `utils.SleepWithJitter` in our environment. I never see a conflict of `nextTrash` again.

davies: Once a client succeed, it will not try to increase nextTrash, so you will not see retried transaction.

PR #5325, opened 2024-11-26
Donot hang when showing stack traces

polyrabbit:
Mount process and supervisor process occasionally hang and require manual killing. It turns out the supervisor process hangs while accessing mount process's pprof interface. A timeout is needed to prevent it from waiting indefinitely.

img_v3_02gt_23b633fc-a613-4588-bc1c-0953eb1638bg

### DIFF
--- a/cmd/mount_unix.go
+++ b/cmd/mount_unix.go
@@ -59,12 +59,16 @@ func showThreadStack(agentAddr string) {
 	if agentAddr == "" {
 		return
 	}
-	resp, err := http.Get(fmt.Sprintf("http://%s/debug/pprof/goroutine?debug=2", agentAddr))
+	client := http.Client{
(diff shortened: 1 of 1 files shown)


anysql:
-- make sure http timeout is working

```
package main

import (
    "fmt"
    "net/http"
    "time"
)

func main() {
	client := http.Client{
	    Timeout: 10 * time.Millisecond,
	}
	resp, err := client.Get("http://www.google.com/")
	if err != nil {
		fmt.Println(err)
	} else {
		_ = resp.Body.Close()
	}
}

```
-- run the code
loufangxin@loufangxindeMacBook-Pro chan % go run .
Get "http://www.google.com/": context deadline exceeded (Client.Timeout exceeded while awaiting headers)

PR #5338, opened 2024-12-03
Multipart upload fully obeys keep-etag config

polyrabbit:
In current code - when `keep-etag` is disabled (by default), they are saved but never used.

https://github.com/juicedata/juicefs/blob/b63efb678efbb91fbfd112211298a677f323ada4/pkg/gateway/gateway.go#L1082-L1085

### DIFF
--- a/pkg/gateway/gateway.go
+++ b/pkg/gateway/gateway.go
@@ -987,7 +987,10 @@ func (n *jfsObjects) ListObjectParts(ctx context.Context, bucket, object, upload
 	for _, entry := range entries {
 		num, er := strconv.Atoi(string(entry.Name))
 		if er == nil && num > partNumberMarker {
-			etag, _ := n.fs.GetXattr(mctx, n.ppath(bucket, uploadID, string(entry.Name)), s3Etag)
+			var etag []byte
+			if n.gConf.KeepEtag {
+				etag, _ = n.fs.GetXattr(mctx, n.ppath(bucket, uploadID, string(entry.Name)), s3Etag)
+			}
 			result.Parts = append(result.Parts, minio.PartInfo{
 				PartNumber:   num,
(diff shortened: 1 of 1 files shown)


pkg/gateway/gateway.go
```diff
@@ -987,7 +987,10 @@ func (n *jfsObjects) ListObjectParts(ctx context.Context, bucket, object, upload
 	for _, entry := range entries {
 		num, er := strconv.Atoi(string(entry.Name))
 		if er == nil && num > partNumberMarker {
-			etag, _ := n.fs.GetXattr(mctx, n.ppath(bucket, uploadID, string(entry.Name)), s3Etag)
+			var etag []byte
+			if n.gConf.KeepEtag {
+				etag, _ = n.fs.GetXattr(mctx, n.ppath(bucket, uploadID, string(entry.Name)), s3Etag)
```
davies, line 992: I think we should keep this, and verify that in CompleteMultipartUpload

pkg/gateway/gateway.go
```diff
@@ -1027,7 +1030,7 @@ func (n *jfsObjects) PutObjectPart(ctx context.Context, bucket, object, uploadID
 	var etag string
 	if err = n.putObject(ctx, bucket, p, r, opts, func(tmpName string) {
 		etag = r.MD5CurrentHexString()
-		if n.fs.SetXattr(mctx, tmpName, s3Etag, []byte(etag), 0) != 0 {
+		if n.gConf.KeepEtag && n.fs.SetXattr(mctx, tmpName, s3Etag, []byte(etag), 0) != 0 {
```
davies, line 1033: I think we should keep this, and verify that in CompleteMultipartUpload

polyrabbit: OK, I will keep the code as is.

PR #5343, opened 2024-12-04
Reduce scan interval in cleanupDeletedFiles

polyrabbit:
In most cases, `cleanupDeletedFiles` runs without doing anything. We can reduce scan calls to meta engine by decreasing the check frequency to an hourly basis.


pkg/meta/base.go
```diff
@@ -604,8 +604,8 @@ func (m *baseMeta) Init(format *Format, force bool) error {
 
 func (m *baseMeta) cleanupDeletedFiles() {
 	for {
-		utils.SleepWithJitter(time.Minute)
-		if ok, err := m.en.setIfSmall("lastCleanupFiles", time.Now().Unix(), int64(time.Minute.Seconds())*9/10); err != nil {
+		utils.SleepWithJitter(time.Hour)
```
davies, line 607: we should increase the limit below

polyrabbit: OK, updated.

PR #5344, opened 2024-12-04
meta/sql: support new backup

jiefenghuang:


pkg/meta/backup.go
```diff
@@ -0,0 +1,413 @@
...
+	Typ uint32
+	Len uint64
+	Val proto.Message
+}
+
+func (s *BakSegment) String() string {
+	return string(proto.MessageName(s.Val).Name())
+}
+
+func (s *BakSegment) Marshal() ([]byte, error) {
```
davies, line 222: Write the marshaled message into io.Writer directly to avoid memory copy?

pkg/meta/pb/backup.proto
```diff
@@ -0,0 +1,153 @@
...
+  bytes target = 2;
+}
+
+message SymlinkList {
+  repeated Symlink list = 1;
+}
+
+message Footer {
+  message SegInfo {
+    repeated uint64 offset = 1;
```
davies, line 147: also type

davies: nvm, we have type in infos.

pkg/meta/redis_bak.go
```diff
@@ -0,0 +1,37 @@
+//go:build !noredis
+// +build !noredis
+
+/*
+ * JuiceFS, Copyright 2020 Juicedata, Inc.
```
davies, line 5: 2024

pkg/meta/backup.go
jiefenghuang: done

pkg/meta/redis_bak.go
jiefenghuang: done

PR #5359, opened 2024-12-10
Fix watchdog thread for umount may hang

polyrabbit:
We still observe unkillable juicefs processes in our environment, here is one scenario:

1. `doUmount` call has finished, but it failed.
2. watch dog thread for `doUmount` stucks in `FlushAll`.

The watch dog thread is used to monitor `doUmount` thread, but I think a reverse watch mechanism also needed.

image

### DIFF
--- a/cmd/mount_unix.go
+++ b/cmd/mount_unix.go
@@ -736,7 +736,13 @@ func installHandler(m meta.Meta, mp string, v *vfs.VFS, blob object.ObjectStorag
 				logger.Errorf("exit after receiving signal %s, but umount does not finish in 30 seconds, force exit", sig)
 				os.Exit(meta.UmountCode)
 			}()
-			go func() { _ = doUmount(mp, true) }()
+			go func() {
+				if err := doUmount(mp, true); err != nil {
+					logger.Warnf("Umount failed: %s", err)
+				}
+				time.Sleep(time.Second * 30)
+				os.Exit(meta.UmountCode) // Just in case the above watchdog thread hangs
+			}()
 		}
 	}()
 }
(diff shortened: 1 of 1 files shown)


davies:
The above lock is suspicious, have you find out the cause?

cmd/mount_unix.go
```diff
@@ -736,7 +736,13 @@ func installHandler(m meta.Meta, mp string, v *vfs.VFS, blob object.ObjectStorag
 				logger.Errorf("exit after receiving signal %s, but umount does not finish in 30 seconds, force exit", sig)
 				os.Exit(meta.UmountCode)
 			}()
-			go func() { _ = doUmount(mp, true) }()
+			go func() {
+				if err := doUmount(mp, true); err != nil {
+					logger.Warnf("Umount failed: %s", err)
+				}
+				time.Sleep(time.Second * 30)
+				os.Exit(meta.UmountCode) // Just in case the above watchdog thread hangs
```
davies, line 744: FlushAll may take long than 30 seconds, some data may lose if we exit before that.

polyrabbit:
> The above lock is suspicious, have you find out the cause?

I suppose it's because user sends too many write requests - there are more than 500 `commitThread` threads. User's process becomes unresponsive and unkillable (`[rclone] <defunct>`), so as the juicefs process. I suppose juicefs needs a way to protect itself from such a large write concurrency.

cmd/mount_unix.go
polyrabbit: How about only os.Exit when `doUmount` fails?

`FlushAll` only flushes memory buffers. Those are requests not acked to client, I suppose it's OK to drop them under abnormal conditions. Juicefs only promises a `close-to-open` consistency.

PR #5361, opened 2024-12-11
meta/backup: fix slice conflict

jiefenghuang:
1. A new slice is needed as concurrent operations might still use the same slice. 
3. Batch size for xattr is already limited at the outer layer.

### DIFF
--- a/pkg/meta/redis_bak.go
+++ b/pkg/meta/redis_bak.go
@@ -241,8 +241,8 @@ func (m *redisMeta) dumpDelFiles(ctx Context, opt *DumpOption, ch chan<- *dumped
 		return err
 	}
 
-	delFiles := make([]*pb.DelFile, 0, len(zs))
-	for _, z := range zs {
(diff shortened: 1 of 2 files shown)


pkg/meta/redis_bak.go
```diff
@@ -255,7 +255,7 @@ func (m *redisMeta) dumpDelFiles(ctx Context, opt *DumpOption, ch chan<- *dumped
 			if err := dumpResult(ctx, ch, &dumpedResult{msg: &pb.Batch{Delfiles: delFiles}}); err != nil {
 				return err
 			}
-			delFiles = delFiles[:0]
+			delFiles = make([]*pb.DelFile, 0, utils.Min(len(zs)-i-1, redisBatchSize))
```
davies, line 258: we should fix these in sql also

jiefenghuang: done

PR #5365, opened 2024-12-11
meta/kv: support new backup

jiefenghuang:


pkg/meta/redis.go
```diff
@@ -608,7 +608,7 @@ func (m *redisMeta) nextTrashKey() string {
 }
 
 func (m *redisMeta) counterKey(name string) string {
-	if name == "nextInode" || name == "nextChunk" || name == "nextSession" {
+	if name == "nextInode" || name == "nextChunk" || name == "nextSession" || name == "nextTrash" {
```
davies, line 611: nextTrash is not used as lower case

pkg/meta/tkv_bak.go
```diff
@@ -17,28 +17,641 @@
...
+	default:
+		logger.Warnf("skip segment type %d", typ)
+		return nil
+	}
 }
 
 func (m *kvMeta) prepareLoad(ctx Context, opt *LoadOption) error {
+	opt.check()
+	opt.Threads = 1
+	logger.Infof("concurrent load is currently not supported , may cause lots of txn conflicts.")
```
davies, line 103: ?

pkg/meta/tkv_bak.go
```diff
@@ -17,28 +17,641 @@
...
+		}
+		for _, list := range lists {
+			_ = dumpResult(ctx, ch, &dumpedResult{list, release})
+		}
+	}()
+
+	if opt.Threads > 0xFF {
+		opt.Threads = 0xFF
+	}
+	rs := splitInodeRange(byte(opt.Threads))
```
davies, line 320: We can always use 256 ranges

pkg/meta/tkv_tikv.go
```diff
@@ -184,8 +184,25 @@ func (c *tikvClient) shouldRetry(err error) bool {
 	return strings.Contains(err.Error(), "write conflict") || strings.Contains(err.Error(), "TxnLockNotFound")
 }
 
-func (c *tikvClient) txn(f func(*kvTxn) error, retry int) (err error) {
-	tx, err := c.client.Begin()
+func (c *tikvClient) config(key string) interface{} {
+	if key == "startTS" {
+		ts, err := c.client.CurrentTimestamp(oracle.GlobalTxnScope)
+		if err != nil {
+			logger.Warnf("TiKV get startTS: %s", err)
```
davies, line 191: This result will inconsistent, return error

jiefenghuang: fixed, return error when tikv get ts error.

davies: it's not fixed in last commit

PR #5366, opened 2024-12-11
Also clean up expired tmp dirs

polyrabbit:
`.sys` dir is created by gateway once user used it, but it's never deleted automatically. We should clean it up when expired to avoid confusion to users.


zhijian-pro:
The `f.Close(mctx)` is ok, but I don't think deleting an empty directory is necessary. Because even if you delete the empty directory, it will be created again when needed.

PR #5370, opened 2024-12-12
meta/redis: fix backup acl id

jiefenghuang:


pkg/meta/base.go
```diff
@@ -3160,6 +3160,9 @@ func (m *baseMeta) LoadMetaV2(ctx Context, r io.Reader, opt *LoadOption) error {
 			case task = <-taskCh:
 			}
 			if task == nil {
+				if err := m.en.load(ctx, segTypeAcl, opt, nil); err != nil { // insert max acl id
```
davies, line 3163: will be called in every thread

jiefenghuang: fix

PR #5376, opened 2024-12-13
windows: fixed running issues for sub commands

chenjie4255:
The current juicefs.exe fails to execute some sub-commands primarily due to the following issues:
* When opening the “.control” file, juicefs.exe attempts to acquire the inode information of the parent directory of the specified file. In the current implementation of ```GetFileInode()```, the code uses **os.Open** to obtain the file handle, which does not support directory opening. This PR changes the code to use **windows.CreateFile**.
   * windows.CreateFile is called with the windows.FILE_FLAG_BACKUP_SEMANTICS flag to allow directory opening. Since juicefs.exe does not have the SE_BACKUP or SE_RESTORE privilege, this should not have any additional effects on Windows ACL behavior.
* When receiving a open request for the “.control” file, WinFSP will perform a FUSE **getattr** callback, which ultimately calls the doResolve method.
   * In the current doResolve impl, for a “.control” file located in a subfolder (e.g., x:/jfs/aaa/.control), doResolve fails to perform its job.

## testing

After this PR, most of the subcommands work normally (tested with local storage and sqlite3). However, the following commands still cannot run due to non-logical reasons. The fix code wasn't included in this PR.

DEBUG
```
C:\Users\Administrator\Downloads>juicefsd.exe debug --out-dir=c://jfslog/ x://
2024/12/13 13:32:13.271364 juicefs[7484] <FATAL>: failed to create current out dir c:\jfslog\x:\-20241213133213: mkdir c:\jfslog\x:: The filename, directory name, or volume label syntax is incorrect. [main@main.go:31]
```

RESTORE
```
only ROOT Can restore files from trash. 
```


pkg/fs/fs.go
```diff
@@ -762,7 +763,9 @@ func (fs *FileSystem) doResolve(ctx meta.Context, p string, followLastSymlink bo
 		if len(name) == 0 {
 			continue
 		}
-		if parent == meta.RootInode && i == len(ss)-1 && vfs.IsSpecialName(name) {
+		isSpCtrlFile := i == len(ss)-1 && vfs.IsSpecialControlName(name)
```
davies, line 766: We don't want to expose this to Java SDK and S3 gateway, so it's better to have this only for winfs

chenjie4255: changes were made, now it only affects the behavior on windows.

pkg/fs/fs.go
```diff
@@ -762,7 +764,9 @@ func (fs *FileSystem) doResolve(ctx meta.Context, p string, followLastSymlink bo
 		if len(name) == 0 {
 			continue
 		}
-		if parent == meta.RootInode && i == len(ss)-1 && vfs.IsSpecialName(name) {
+		isWinSpCtrlFile := i == len(ss)-1 && runtime.GOOS == "windows" && vfs.IsSpecialControlName(name)
```
davies, line 767: we can run s3 gateway or Java SDK on Windows, so they are still affected.

chenjie4255: The related logic has been moved to winfs.getattr and winfs.open now.

PR #5383, opened 2024-12-16
dump/load all metadata in new binary format

jiefenghuang:


davies:
I think we can use `--binary` to existing dump/load commands

PR #5384, opened 2024-12-16
cmd/rmr: add `skip-trash` option to delete files directly

SandyXSD:


pkg/meta/base_test.go
```diff
@@ -1341,7 +1341,7 @@ func testRemove(t *testing.T, m Meta) {
 	} else if len(entries) != 4099 {
 		t.Fatalf("entries: %d", len(entries))
 	}
-	if st := m.Remove(ctx, 1, "d", nil); st != 0 {
+	if st := m.Remove(ctx, 1, "d", false, nil); st != 0 {
```
davies, line 1344: we should add a case for skip-trash

SandyXSD: Added in the `testTrash` since trash is disabled in other tests.

Issue #5407, opened 2024-12-19
we don't need to open and close a cache file if we just want to check if it exists


daimashusheng:
**What would you like to be added**:
When we are warming up files,  we can check if a cache file exists to replace opening and closing it.

**Why is this needed**:
Opening and closing a cache file is more costly than just checking if it exists.

PR #5408, opened 2024-12-19
check if a cache file exists to avoid opening and closing it

daimashusheng:
we don't need to open and close a cache file if we just want to check if it exists
https://github.com/juicedata/juicefs/issues/5407

### DIFF
--- a/pkg/chunk/disk_cache.go
+++ b/pkg/chunk/disk_cache.go
@@ -615,6 +615,39 @@ func (cache *cacheStore) load(key string) (ReadCloser, error) {
 	return f, err
 }
 
+func (cache *cacheStore) exist(key string) (bool, error) {
+	cache.Lock()
+	defer cache.Unlock()
+	if _, ok := cache.pages[key]; ok {
+		return true, nil
+	}
+	k := cache.getCacheKey(key)
+	if cache.scanned && cache.keys[k].atime == 0 {
+		return false, errNotCached
+	}
+	cache.Unlock()
+	var err error
+	err = cache.checkErr(func() error {
+		_, err = os.Stat(cache.cachePath(key))
+		return err
+	})
+
+	cache.Lock()
+	if err == nil {
+		if it, ok := cache.keys[k]; ok {
+			// update atime
+			cache.keys[k] = cacheItem{it.size, uint32(time.Now().Unix())}
+		}
+		return true, nil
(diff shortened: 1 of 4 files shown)


pkg/chunk/disk_cache.go
```diff
@@ -615,6 +615,48 @@ func (cache *cacheStore) load(key string) (ReadCloser, error) {
...
+		return err
+	})
+
+	cache.Lock()
+	if err == nil {
+		if it, ok := cache.keys[k]; ok {
+			// update atime
+			nowTime := time.Now()
+			err = cache.checkErr(func() error {
+				err = os.Chtimes(cache.cachePath(key), nowTime, fi.ModTime())
```
davies, line 642: After this, this one could be slower than before if the disk is mounted with noatime.

I think we can ignore this, the atime in memory could be enough.

davies:
Can you run a benchmark to show the performance difference?

daimashusheng:
> Can you run a benchmark to show the performance difference?

Run a simple test comparing open/close a file to stat a file, the code is 
```
package main

import (
        "fmt"
        "os"
        "time"
)

func main() {
        startTime := time.Now()
        for i := 0; i < 10000; i++ {
                file, err := os.Open("myfile")
                if err != nil {
                      fmt.Println("Error opening file:", err)
                      return
                }

                err = file.Close()
                if err != nil {
                        fmt.Println("Error closing file:", err)
                        return
                }
        }

        elapsedTime := time.Since(startTime)
        fmt.Printf("Total time for opening and closing the file 10000 times: %v\n", elapsedTime)
}
```
run it, and the result is
```
[root@localhost gotest]# go run openfile.go 
Total time for opening and closing the file 10000 times: 20.585789ms
[root@localhost gotest]# go run openfile.go 
Total time for opening and closing the file 10000 times: 26.070706ms
[root@localhost gotest]# go run openfile.go 
Total time for opening and closing the file 10000 times: 21.153124ms
```

after that, change open/close to stat, the result is
```
[root@localhost gotest]# go run openfile.go 
Total time for opening and closing the file 10000 times: 6.796816ms
[root@localhost gotest]# go run openfile.go 
Total time for opening and closing the file 10000 times: 6.055107ms
[root@localhost gotest]# go run openfile.go 
Total time for opening and closing the file 10000 times: 6.805013ms
```

we can see a significant improvement

davies (approved):

PR #5414, opened 2024-12-20
meta: fix the issue that trash files may be dangling

SandyXSD:
Assume the current time is just before 11:00, it's possible that
```
t1(10.999): A checkTrash, using trash directory xxx-10
t2(11.001): B cleanup, removing the trash directory one hour before, which is xxx-10
t3(11.002): A unlink, moving the file F under (already deleted) xxx-10, then F becomes dangling
```

### DIFF
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -2502,7 +2502,7 @@ func (m *baseMeta) scanTrashFiles(ctx Context, scan trashFileScan) error {
 }
 
 func (m *baseMeta) doCleanupTrash(days int, force bool) {
-	edge := time.Now().Add(-time.Duration(24*days+1) * time.Hour)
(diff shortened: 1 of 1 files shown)


polyrabbit:
It's better to check existence of trash inode in the `doUnlink` txn?

anysql:
How do deal with the force mode?

SandyXSD:
> It's better to check existence of trash inode in the `doUnlink` txn, or wrap `checkTrash` within that txn?

That's not worthy as it will harm the performance.

SandyXSD:
> How do deal with the force mode?

`Force` mode is only used in tests.

jiefenghuang (approved):

polyrabbit:
Do we have tool to find those "dangling" nodes?

SandyXSD:
> Do we have tool to find those "dangling" nodes?

Currently no.

PR #5460, opened 2024-12-27
perf - Improve MySQL/PostgreSQL meta concurrency

anysql:
1, Optimize the ForUpdate call of doMknod/Link/Unlink/Rmdir/Rename operation
2, Change nlink operation from "Nlink = ?" to "nlink = nlink + (?)" for atomic
3, Move the parent update operation to the end of transaction
4, Remove txLock for MySQL/PostgreSQL meta engine


pkg/meta/sql.go
```diff
@@ -2194,19 +2247,23 @@ func (m *dbMeta) doLink(ctx Context, inode, parent Ino, name string, attr *Attr)
...
-		}
 		if _, err := s.Cols("nlink", "ctime", "ctimensec", "parent").Update(&n, node{Inode: inode}); err != nil {
 			return err
 		}
 		if err == nil {
 			m.parseAttr(&n, attr)
 		}
+		if updateParent {
+			if n, err := s.Cols("mtime", "ctime", "mtimensec", "ctimensec").Update(&pn, &node{Inode: parent}); err != nil || n == 0 {
+				if err == nil {
```
davies, line 2258: we should double check if the parent exists or not for MySQL

anysql: In rare condition (high concurrency), no column will be real updated. Let me think and find out a way for MySQL.

anysql: Just read the code, Link operation does not update parent always since it does not change nlink, seems it's ok?
The c/mtimensec update is missing from the original code, fixed.

davies: Since we removed the lock, the parent could be delete in the same time as we try to link, so we should make sure that the parent exists when the update affected 0 rows.

anysql: It does not work as we use repeatable read, even if deleted by other session, you can still get one row in the current transaction.

anysql: Return syscall.EAGAIN to retry the transaction is a good way in this case.

anysql:
One question, when use EAGAIN to retry the transaction, some test cases failed.

After switch to EBUSY, all test cases passed, still need to evaluate is it ok to use EBUSY here.

PR #5462, opened 2024-12-27
Allow kernel to cache readdir entries to reduce overhead of meta engines

polyrabbit:
Some programs rely on frequent `readdir` calls to obtain updates from other processes. One example is [TensorBoard](https://www.tensorflow.org/tensorboard), which is widely used in machine learning area. This imposes a significant load on the backend metadata engine. In our environment, we observed over 200k QPS, but most directories are cold. Therefore, it is necessary to introduce a caching feature to reduce the metadata load.

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -349,6 +349,10 @@ func metaCacheFlags(defaultEntryCache float64) []cli.Flag {
 			Value: "1.0s",
 			Usage: "dir entry cache timeout",
 		},
+		&cli.BoolFlag{
+			Name:  "readdir-cache",
+			Usage: "enable kernel caching of readdir entries, with timeout controlled by attr-cache flag",
+		},
 		&cli.StringFlag{
 			Name:  "open-cache",
 			Value: "0s",
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -351,7 +351,7 @@ func metaCacheFlags(defaultEntryCache float64) []cli.Flag {
 		},
 		&cli.BoolFlag{
 			Name:  "readdir-cache",
-			Usage: "enable kernel caching of readdir entries, with timeout controlled by attr-cache flag",
+			Usage: "enable kernel caching of readdir entries, with timeout controlled by attr-cache flag (require linux kernel 4.20+)",
 		},
 		&cli.StringFlag{
 			Name:  "open-cache",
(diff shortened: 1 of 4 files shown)


solracsf:
https://juicefs.com/docs/community/guide/cache/#kernel-metadata-cache needs update?

cmd/flags.go
```diff
@@ -349,6 +349,10 @@ func metaCacheFlags(defaultEntryCache float64) []cli.Flag {
 			Value: "1.0s",
 			Usage: "dir entry cache timeout",
 		},
+		&cli.BoolFlag{
+			Name:  "readdir-cache",
+			Usage: "enable kernel caching of readdir entries, with timeout controlled by attr-cache flag",
```
davies, line 354: (require linux kernel 4.20+)

pkg/fuse/fuse.go
```diff
@@ -347,6 +347,9 @@ func (fs *fileSystem) OpenDir(cancel <-chan struct{}, in *fuse.OpenIn, out *fuse
 	defer releaseContext(ctx)
 	fh, err := fs.v.Opendir(ctx, Ino(in.NodeId), in.Flags)
 	out.Fh = fh
+	if fs.conf.ReaddirCache {
```
davies, line 350: We should check if the directory is changed or not, similar to files

cmd/flags.go
davies: attr-cache should be enabled, otherwise the readdir will be slow than before, because there will be lots of lookup/getattr operations.

pkg/fuse/fuse.go
davies: we should also check the version of kernel module,  require 4.20+

cmd/flags.go
polyrabbit: How to ensure `attr-cache` is enabled? set it to 1s or use `logger.Fatalf` to force user specify one in `Config.SelfCheck()`?

pkg/fuse/fuse.go
polyrabbit: How to track if the dir has changed or not, using mtime? kernel already did the [check](https://lore.kernel.org/lkml/20180928154234.19270-9-mszeredi@redhat.com/)

cmd/flags.go
davies: We can mention this in docs

polyrabbit: Updated code to print a warning log if the readdir-cache dependency is not met.

polyrabbit:
@solracsf I'm not professional at documentation, will leave it to @CaitinChen :)

PR #5486, opened 2024-12-31
meta/session: exit background tasks when close

### LINKED ISSUE #5455: Gracefully exit meta
Gracefully exit meta, wait and clean up background jobs, ex. goroutines in NewSession.

**Why is this needed**:
case: gc should wait for m.dslices to be done.

jiefenghuang:
close #5455 
fix [test](https://github.com/juicedata/juicefs/actions/runs/12541171203/job/34969355871)

### DIFF
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -83,7 +83,7 @@ type engine interface {
 	doFindDeletedFiles(ts int64, limit int) (map[Ino]uint64, error) // limit < 0 means all
 	doDeleteFileData(inode Ino, length uint64)
 	doCleanupSlices()
-	doCleanupDelayedSlices(edge int64) (int, error)
+	doCleanupDelayedSlices(ctx Context, edge int64) (int, error)
 	doDeleteSlice(id uint64, size uint32) error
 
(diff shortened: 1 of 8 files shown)


pkg/meta/base.go
```diff
@@ -477,27 +490,56 @@ func (m *baseMeta) NewSession(record bool) error {
...
+		go m.symlinks.clean(ctx, &m.sessWG)
+	}
+	return nil
+}
+
+func (m *baseMeta) startDeleteSliceTasks() {
+	if m.conf.MaxDeletes <= 0 || m.dslices != nil {
+		return
+	}
+	m.sessWG.Add(m.conf.MaxDeletes)
```
davies, line 514: we will wait for them in stopDeleteSliceTasks()

jiefenghuang: Normally, closing a session does not require waiting for the cleanup of m.dslices. Canceling the context causes tasks to exit early. The stopDeleteSliceTasks function is reserved for independent invocation in scenarios where m.dslices needs to be fully cleaned up, such as during garbage collection (GC).

jiefenghuang: Calling stopDeleteSliceTasks at line 658 is primarily for resetting dslices.

Issue #5478, opened 2024-12-30
Also add a limit to the number of cache item?


polyrabbit:
Currently, we have `cache-size` and `free-ratio` to limit usage of `cache-dir`. Both can restrict the space usage of a single volume.

Our approach is to set `free-ratio` to 0.1 to prevent a single volume from filling up the entire disk. Additionally, we set a `cache-size` for each volume to prevent any single volume from occupying the entire `cache-dir`, as we schedule multiple volumes on a single machine.

The issue we are encountering now is that some volumes still write a large number of small files, causing the number of inodes to hit the `free-ratio` limit. As a result, all volumes on the machine start evicting cache, affecting the performance of other volumes. 

So, I suppose we need a feature equivalent to `cache-size` to limit the number of cache items?

davies:
Yes, we could.

PR #5489, opened 2024-12-31
Support limiting number of cache items

polyrabbit:
As suggested in #5478 - to avoid inode usage confilicts when multiple volumes are scheduled to the same host.

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -216,8 +216,8 @@ func dataCacheFlags() []cli.Flag {
 		},
 		&cli.Int64Flag{
 			Name:  "cache-items",
-			Value: 1e9, // should be enough for most disks
-			Usage: "max number of cached items",
+			Value: 0,
+			Usage: "max number of cached items (0 for unlimited)",
 		},
 		&cli.Float64Flag{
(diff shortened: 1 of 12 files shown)


pkg/chunk/cached_store.go
```diff
@@ -650,6 +651,10 @@ func (c *Config) parseHours() (start, end int, err error) {
 	return
 }
 
+func (c *Config) CacheEnabled() bool {
+	return c.CacheSize > 0 && c.CacheItems > 0
```
davies, line 655: CacheItems could be 0 for unlimited, we can only check `CacheSize`

cmd/flags.go
```diff
@@ -214,6 +214,11 @@ func dataCacheFlags() []cli.Flag {
 			Value: "100G",
 			Usage: "size of cached object for read in MiB",
 		},
+		&cli.Int64Flag{
+			Name:  "cache-items",
+			Value: 1e9, // should be enough for most disks
```
davies, line 219: 0 for unlimited

polyrabbit: OK, all updated

PR #5496, opened 2025-01-02
meta/sql: fix nanosecond of link and readlink

davies:


pkg/meta/sql.go
```diff
@@ -1373,9 +1381,9 @@ func (m *dbMeta) doReadlink(ctx Context, inode Ino, noatime bool) (atime int64,
 		if !m.atimeNeedsUpdate(attr, now) {
 			return nil
 		}
-		nodeAttr.Atime = now.Unix()*1e6 + int64(now.Nanosecond())/1e3
-		nodeAttr.Atimensec = int16(now.Nanosecond() % 1e3)
+		nodeAttr.setAtime(now.UnixNano())
 		attr.Atime = now.Unix()
+		attr.Atimensec = uint32(now.Nanosecond())
```
jiefenghuang, line 1386: The return value only uses ‘Atime’，‘Atimensec’ is not used.

SandyXSD (approved):

PR #5497, opened 2025-01-02
cmd/sync: allocate memory from sync.Pool instead of Chunk.NewOffPage to avoid stall problem

### LINKED ISSUE #5492: Sync Process Stalls with High Thread Count Due to Buffer Limitations
**What Happened**:

When attempting to sync a 640 MiB object from MinIO to JuiceFS without using a mount point and utilizing 64 threads, the process stalls and never completes.

**What You Expected to Happen**:

The sync process should successfully complete.

**How to Reproduce It (as minimally and precisely as possible)**:

Try syncing a 640 MiB data file from MinIO to JuiceFS with Redis as the metadata engine using 64 threads. Execute the command below, and you may encounter a situation where the sync process fails to complete, albeit with a very low chance of success.

```bash
dst_jfs=redis://localhost/1 \
juicefs \
sync \
minio://minioadmin:minioadmin@localhost:9000/testbucket/sample_640MiB_file.data \
jfs://dst_jfs/sample_640MiB_file.data \
--threads=64 \
--force-update
```

**Additional Information**:

To address this issue, consider the following:

- JuiceFS is initialized with a default 300 MiB buffer, which users cannot modify. See `cmd/object.go: func getDefaultChunkConf`.

- In `pkg/sync/download.go: download`, a page is allocated by invoking `chunk.NewOffPage` from the buffer. These pages are only released when they are read.

- In `pkg/vfs/writer.go: func (f *fileWriter) Write`, writing is not permitted if the buffer usage is too high. Consequently, the lifecycle of the page does not end, and the page is not released.

Therefore, the sync cannot complete because writing requires pages to be freed to alleviate the buffer, but pages are only released when they are used in writing.

**Environment**:

- JuiceFS version (use `juicefs --version`) or Hadoop Java SDK version: JuiceFS version 1.2.1+2024-08-30.cd871d1

- OS (e.g., `cat /etc/os-release`):

- Kernel (e.g., `uname -a`):

- Object storage (cloud provider and region, or self-maintained):

- Metadata engine info (version, cloud provider managed or self-maintained):

- Others:

YinhaoHu:
close #5492

All unit tests related to this update are adjusted and passed.

The case mentioned in the issue can also run now.

### DIFF
--- a/pkg/sync/download.go
+++ b/pkg/sync/download.go
@@ -21,7 +21,6 @@ import (
 	"io"
 	"sync"
 
-	"github.com/juicedata/juicefs/pkg/chunk"
 	"github.com/juicedata/juicefs/pkg/object"
 )
 
@@ -33,7 +32,7 @@ type parallelDownloader struct {
 	fsize      int64
 	blockSize  int64
 	concurrent chan int
-	buffers    map[int64]*chunk.Page
+	buffers    map[int64]*[]byte
 	off        int64
 	err        error
 }
@@ -48,6 +47,15 @@ func (r *parallelDownloader) setErr(err error) {
 	r.Lock()
 	defer r.Unlock()
 	r.err = err
+}
+
+const downloadBufSize = 10 << 20
+
+var downloadBufPool = sync.Pool{
+	New: func() interface{} {
+		buf := make([]byte, 0, downloadBufSize)
+		return &buf
+	},
 }
 
 func (r *parallelDownloader) download() {
@@ -73,18 +81,19 @@ func (r *parallelDownloader) download() {
 					r.setErr(e)
 				} else { //nolint:typecheck
 					defer in.Close()
-					p := chunk.NewOffPage(int(size))
-					_, e = io.ReadFull(in, p.Data)
+					p := downloadBufPool.Get().(*[]byte)
+					*p = (*p)[:size]
+					_, e = io.ReadFull(in, *p)
 					if e != nil {
 						r.setErr(e)
-						p.Release()
+						downloadBufPool.Put(p)
 					} else {
 						r.Lock()
 						if r.buffers != nil {
 							r.buffers[off] = p
 							saved = true
 						} else {
-							p.Release()
+							downloadBufPool.Put(p)
 						}
 						r.Unlock()
 					}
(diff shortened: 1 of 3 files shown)


zhijian-pro:
@YinhaoHu   Thank you very much for your contribution, but we think the following approach would be more appropriate.
The issue lies in our sync download logic where we shouldn't be using chunk.NewOffPage to allocate memory. 
https://github.com/juicedata/juicefs/blob/b51019d2f66d2ea0a180ee99cdbd4aa5202bf06c/pkg/sync/download.go#L76-L76
In this line of code https://github.com/juicedata/juicefs/blob/b51019d2f66d2ea0a180ee99cdbd4aa5202bf06c/pkg/sync/sync.go#L362-L362
 we have hard-coded a block size of 10M, so we can directly use a fixed-size `sync.pool` to allocate memory.

YinhaoHu:
@zhijian-pro Good idea. I have updated this PR based on the more appropriate approach.

pkg/sync/download.go
```diff
@@ -73,18 +81,18 @@ func (r *parallelDownloader) download() {
 					r.setErr(e)
 				} else { //nolint:typecheck
 					defer in.Close()
-					p := chunk.NewOffPage(int(size))
-					_, e = io.ReadFull(in, p.Data)
+					p := downloadBufPool.Get().(*[]byte)
+					_, e = io.ReadFull(in, *p)
```
zhijian-pro, line 85: If it is the last block, the []byte obtained may be too long and needs to be trimmed according to size

zhijian-pro:
It is best to adjust the TestDownload unit test so that it passes

pkg/sync/download.go
YinhaoHu: Mentioned and updated.

YinhaoHu:
> It is best to adjust the `TestDownload` unit test so that it passes

All test cases in the `TestDownload` unit test are modified in an equivalent approach as before. And they all pass.

YinhaoHu:
> @YinhaoHu If the size of the last block is smaller than the default block size, it should not be put back into the pool, otherwise it will cause panic when you get it.

The capacity of each buf is always `downloadBufSize` even if the length is variable, maybe smaller than the default block size, and there will be no block requires capacity more than `downloadBufSize`. I don't think panic will occur. Or could you show me a case where the panic will occur?

zhijian-pro:
The last block returned when file A is downloaded may be removed from the pool when file B is downloaded.
But this block may be less than the length of downloadBufSize, and the B file wants it to be downloadBufSize.

zhijian-pro:
OK，I missed  `*p = (*p)[:size]`.

zhijian-pro (approved):

jiefenghuang (approved):

PR #5499, opened 2025-01-03
perf release mutex while waiting for retry

anysql:
Adjust the mutex scope for txn of redis/sql/tikv meta engine


anysql:
All other time.Sleep are all out of mutex scope.

davies:
The reason we have sleep here is to reduce the concurrency of transactions, it will be useless if it unlocks the mutex.

anysql:
We have 1024 mutex here, if all are active, still too high concurrency for the backend database, so I feel is not used to limit the overall transaction concurrency. It's used to limit the transaction concurrency of single file or directory, as it is usually processed serially in backend.

anysql:
#5512 limit the maximum connections is another good way to control the transaction concurrency.

PR #5512, opened 2025-01-06
meta/sql: support more connection options

anysql:
Add four new options for sql meta engine

1, max-open-conns,  limit the maximum connection, default 0 (unlimited)
2, max-idle-conns, limit the maximum idle connection, previous is cpu number * 2, the new default is 10
3, max-idle-time, previous is 5 mins, no changed, just expose the option.
4, max-life-time, if a connection has be created more than given time, it will be closed anyway, default is 3600,  as a backend connection may have some unclean resources.

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -297,6 +297,26 @@ func metaFlags() []cli.Flag {
 			Name:  "sort-dir",
 			Usage: "sort entries within a directory by name",
 		},
+		&cli.IntFlag{
+			Name:  "max-open-conns",
+			Value: 0,
+			Usage: "maximum number of connections allowed for meta engine (used for SQL engine only)",
+		},
+		&cli.IntFlag{
+			Name:  "max-idle-conns",
+			Value: 10,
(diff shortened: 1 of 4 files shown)


jiefenghuang:
Redis and KV have similar configurations, perhaps can change to a handling method similar to redis.ParseURL

anysql:
> Redis and KV have similar configurations, perhaps can change to a handling method similar to redis.ParseURL

Got.

anysql:
Redis connection pool is provided by the driver, you can control the settings by meta url, we just need to make it clear in docuemnt.

While sql's connection pool is provided by the database/sql module, not by driver.

zhijian-pro:
These parameters are not universal, so it is more appropriate to use query parameters in the meta-url, which may be supported natively by the driver library.

anysql:
A better default setting may be "max_open_conns=30&max_idle_conns=30&max_idle_time=300&max_life_time=3600".

Let the max_idle_conns >= max_open_conns can eliminate the short connections effectively

PR #5518, opened 2025-01-06
meta/redis: remove watch in doGetAttr and doGetFacl

jiefenghuang:


anysql:
The new logic removed the retry for some failures.

jiefenghuang:
> The new logic removed the retry for some failures.
> 
> So does for doRead and doGetParents, while the sql engine has the retry logic.

This is unrelated to the modification. Additionally, the Redis client will retry for common failures.

PR #5521, opened 2025-01-07
object: remove leading slash of object key for upload copy

winglq:
A leading '/' is added for file objects, and UploadPartCopy also add a '/' in CopySource, which leads to the following error.
This patch stripe the leading '/'. 

```
XMinioInvalidObjectName: Object name contains a leading slash.
 status code: 400, request id: 181758CE95205B1E, host id:
```
Tests:
1. sync from file to minio.
2. sync from minio to file.

Tests was done in a hack way to test the upload copy code path.


winglq:
The following test command is not correct, as SRC/DSR does not have a "/" suffix. If the suffix is added everything goes well.
`./juicefs sync --check-all /tmp/sync1 minio://xxx:xxx@127.0.0.1:9000/sync  -v`

PR #5522, opened 2025-01-07
cmd/rmr: add threads option

anysql:
The default rmr concurrency is hard coded as 50, make it a command line option

### DIFF
--- a/pkg/meta/utils.go
+++ b/pkg/meta/utils.go
@@ -352,10 +352,7 @@ func (m *baseMeta) Remove(ctx Context, parent Ino, name string, skipTrash bool, 
 	}
 	if numThreads <= 0 {
 		logger.Infof("Invalid threads number %d , auto adjust to 50 .", numThreads)
-		numThreads = 50
-	} else if numThreads < 2 {
-		logger.Infof("Invalid threads number %d , auto adjust to 2 .", numThreads)
-		numThreads = 2
+		numThreads = RmrDefaultThreads
 	} else if numThreads > 255 {
 		logger.Infof("Invalid threads number %d , auto adjust to 255 .", numThreads)
 		numThreads = 255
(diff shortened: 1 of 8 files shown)


pkg/meta/utils.go
```diff
@@ -350,10 +350,21 @@ func (m *baseMeta) Remove(ctx Context, parent Ino, name string, skipTrash bool,
 		}
 		return m.Unlink(ctx, parent, name)
 	}
-	concurrent := make(chan int, 50)
+	if numThreads <= 0 {
```
solracsf, line 353: Maybe also issue a warning in these cases?

pkg/meta/utils.go
```diff
@@ -350,7 +350,15 @@ func (m *baseMeta) Remove(ctx Context, parent Ino, name string, skipTrash bool,
...
+		logger.Infof("Invalid threads number %d , auto adjust to 50 .", numThreads)
+		numThreads = RmrDefaultThreads
+	} else if numThreads > 255 {
+		logger.Infof("Invalid threads number %d , auto adjust to 255 .", numThreads)
+		numThreads = 255
+	}
+	logger.Infof("Start emptyEntry with %d concurrent threads .", numThreads)
```
jiefenghuang, line 360: Debugf？

pkg/vfs/internal.go
```diff
@@ -298,13 +298,17 @@ func (v *VFS) handleInternalMsg(ctx meta.Context, cmd uint32, r *utils.Buffer, o
 		inode := Ino(r.Get64())
 		name := string(r.Get(int(r.Get8())))
 		var skipTrash bool
+		var numThreads int = meta.RmrDefaultThreads
 		if r.HasMore() {
 			skipTrash = r.Get8()&1 != 0
+			if r.HasMore() {
```
jiefenghuang, line 304: move outside?
```golang
if r.HasMore() {}
if r.HasMore() {}
```

anysql: If the first is false, no need to execute the second, so I put it inside.

pkg/meta/utils.go
anysql: Maybe log it as info will give us some help sometime?

PR #5524, opened 2025-01-07
sync: calculate chksum in flight while copying data

winglq:
This could reduce Get ops to src from 2 to 1 if a verification is needed for sync.

### DIFF
--- a/pkg/sync/sync.go
+++ b/pkg/sync/sync.go
@@ -483,7 +483,8 @@ func doCopySingle(src, dst object.ObjectStorage, key string, size int64, calChks
 			defer f.Close()
 			buf := bufPool.Get().(*[]byte)
 			defer bufPool.Put(buf)
-			if _, err = io.CopyBuffer(f, downer, *buf); err == nil {
+			// hide f.ReadFrom to avoid discarding buf
+			if _, err = io.CopyBuffer(struct{ io.Writer }{f}, downer, *buf); err == nil {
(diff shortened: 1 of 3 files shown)


pkg/sync/sync.go
```diff
@@ -368,19 +477,20 @@ func doCopySingle(src, dst object.ObjectStorage, key string, size int64) error {
...
 				logger.Warnf("create temp file: %s", err)
-				return doCopySingle0(src, dst, key, size)
+				return doCopySingle0(src, dst, key, size, calChksum)
 			}
 			_ = os.Remove(f.Name()) // will be deleted after Close()
 			defer f.Close()
 			buf := bufPool.Get().(*[]byte)
 			defer bufPool.Put(buf)
-			if _, err = io.CopyBuffer(struct{ io.Writer }{f}, downer, *buf); err == nil {
+			if _, err = io.CopyBuffer(f, downer, *buf); err == nil {
```
davies, line 486: We should keep the hack here

winglq: ok, but why?

davies: This avoid the zero copy for some reasons

davies: https://github.com/juicedata/juicefs/commit/e89563fb238d88f444e45148a3e076443964ecd1

winglq: Got it, thx.

PR #5530, opened 2025-01-08
cmd/sync: eliminate wasteful list for some stores in sync

### LINKED ISSUE #5434: Optimize sync performance for some stores by reducing redundant directory scans

Optimize the store.List behavior for some stores(filestore, nfsstore and gluster store) by increasing the limit argument when listing files during sync operations.

**Why is this needed**:

Currently, when syncing large directories with FileStore, there are significant performance issues:

1. Current behavior:

- listCommonPrefix uses maxResult=1000 for each list operation
- FileStore reads the entire directory(let's say over 10 million) to return just 1,000 files

This process repeats multiple times for large directories

2. Problems:

- Extremely inefficient for directories with millions of files
- Each list operation unnecessarily scans the entire directory
- Makes syncing large directories (e.g., 10 million files) practically impossible

I met the case recently. Read the entire directory needs around 90 seconds but only returns 1000 in filestore List, which is terrible to get the entire 10 millions files in listCommonPrefix.

3. Proposed solution:

Increase the limit to a higher value if the store is file store in listCommonPrefix.

YinhaoHu:
close #5434

### DIFF
--- a/pkg/sync/sync.go
+++ b/pkg/sync/sync.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"hash/crc32"
 	"io"
+	"math"
 	"os"
 	"path"
 	"runtime"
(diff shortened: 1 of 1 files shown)


pkg/sync/sync.go
```diff
@@ -1247,8 +1248,12 @@ func listCommonPrefix(store object.ObjectStorage, prefix string, cp chan object.
 	var nextToken string
 	var marker string
 	var hasMore bool
+	var thisListMaxResults int64 = maxResults
+	if strings.HasPrefix(store.String(), "file://") {
```
davies, line 1252: We should do this for all the filesystem

YinhaoHu: All object stores which list only a limited number of files by reading the entire directory are included.

zhijian-pro:
Still need to add `jfs://` `hdfs://` `webdav://`

YinhaoHu:
> @YinhaoHu Still need to add `jfs://` `hdfs://` `webdav://`

Added.

Issue #5367, opened 2024-12-11
JuiceFS gateway support for object metadata


stevapple:
**What would you like to be added**:
Add support for storing, fetching and modifying object metadata in JuiceFS gateway.

**Why is this needed**:
Object metadata is used for statistics and will be user-facing (as HTTP headers), compared to user tags. This is extremely useful when we want to expose some custom metadata to downstream users (eg. web clients).

davies:
Can you provide a more detailed example on how this feature will be used?

stevapple:
@davies Object metadata is tightly associated with the object, and is user-facing. If we want to allow end-users to provide and get metadata through a public URL (pre-signed or anonymous), this is the only choice. We’re especially using this feature to add MIME and other useful information for media/file management backed by S3.

prOOrc:
We have encountered an issue because of rclone tries to store mtime in the object metadata.

zhijian-pro:
@prOOrc  Can you provide the specific details for reproducing the error, we will not store the metadata, but it should not cause an error.

zhijian-pro:
@stevapple  We have no support plan for object metadata at this time. 
Supporting the object metadata  can lead to degraded performance in getObject operations.

prOOrc:
@zhijian-pro Yes, it doesn't cause an error. But rclone tries to preserve the mtime in metadata. We have to ignore the modification time during file synchronization from/to sources that support the mtime.
I added metadata support in my [fork](https://github.com/juicedata/juicefs/pull/5409). It can be enabled optionally.

davies:
@prOOrc Can you send a PR for it?

PR #5536, opened 2025-01-09
Gateway: add object metadata

prOOrc:
Enable object metadata for s3 gateway
#5367

### DIFF
--- a/pkg/gateway/gateway.go
+++ b/pkg/gateway/gateway.go
@@ -839,6 +839,10 @@ func (n *jfsObjects) PutObject(ctx context.Context, bucket string, object string
 					}
 				}
 			}
+			err = n.setObjMeta(tmpName, opts.UserDefined)
+			if err != nil {
+				logger.Errorf("set object metadata error, path: %s error %s", p, err)
+			}
 		}); err != nil {
 			return
 		}
@@ -846,11 +850,6 @@ func (n *jfsObjects) PutObject(ctx context.Context, bucket string, object string
 	fi, eno := n.fs.Stat(mctx, p)
 	if eno != 0 {
 		return objInfo, jfsToObjectErr(ctx, eno, bucket, object)
-	}
-
-	err = n.setObjMeta(p, opts.UserDefined)
-	if err != nil {
-		logger.Errorf("set object metadata error, path: %s error %s", p, err)
 	}
 
 	return minio.ObjectInfo{
@@ -1178,6 +1177,15 @@ func (n *jfsObjects) CompleteMultipartUpload(ctx context.Context, bucket, object
 		}
 	}
 
+	var objMeta map[string]string
+	if n.gConf.ObjMeta {
+		if objMeta, err = n.getObjMeta(n.upath(bucket, uploadID)); err != nil {
+			logger.Errorf("get object meta error, path: %s, error: %s", n.upath(bucket, uploadID), err)
+		} else if err = n.setObjMeta(tmp, objMeta); err != nil {
+			logger.Errorf("set object meta error, path: %s, error: %s", tmp, err)
+		}
+	}
+
 	name := n.path(bucket, object)
 	eno = n.fs.Rename(mctx, tmp, name, 0)
 	if eno == syscall.ENOENT {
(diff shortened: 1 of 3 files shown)


zhijian-pro:
Thanks，please sign the CLA first.

prOOrc:
@zhijian-pro Done.

pkg/gateway/gateway.go
```diff
@@ -871,6 +898,56 @@ const s3Etag = "s3-etag"
...
+	}
+	return objMeta, nil
+}
+
+func (n *jfsObjects) setObjMeta(p string, metadata map[string]string) error {
+	if n.gConf.ObjMeta && metadata != nil {
+		meta := make(map[string]string)
+		for k, v := range metadata {
+			k = strings.ToLower(k)
+			if strings.HasPrefix(k, amzMeta) {
```
zhijian-pro, line 934: s3 metadata does not have a fixed prefix rule.
> There are two kinds of object metadata in Amazon S3: system-defined metadata and user-defined metadata. System-defined metadata includes metadata such as the object's creation date, size, and storage class. User-defined metadata is metadata that you can choose to set at the time that you upload an object. This user-defined metadata is a set of name-value pairs. For more information, see [System-defined object metadata](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html#SysMetadata) and [User-defined object metadata](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html#UserMetadata).

zhijian-pro (changes requested):
support [User-defined object metadata](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html#UserMetadata).

pkg/gateway/gateway.go
prOOrc: Sorry, but I don't understand. The documentation you provide says:
> the optional user-defined metadata names must begin with **x-amz-meta-** to distinguish them from other HTTP headers

zhijian-pro: `Content-Type` Can it be saved now?

zhijian-pro:
@prOOrc  Why not use an `object-meta` flag switch to control System-define meta and User-define meta together ？

prOOrc:
> @prOOrc Why not use an `object-meta` flag switch to control System-define meta and User-define meta together ？

Done.

pkg/gateway/gateway.go
```diff
@@ -533,6 +535,12 @@ func (n *jfsObjects) CopyObject(ctx context.Context, srcBucket, srcObject, dstBu
 	}
 	dst := n.path(dstBucket, dstObject)
 	src := n.path(srcBucket, srcObject)
+
```
zhijian-pro, line 538: It should be handled like an object-tag, first set to the tmp object.

pkg/gateway/gateway.go
```diff
@@ -687,6 +695,18 @@ func (n *jfsObjects) GetObjectInfo(ctx context.Context, bucket, object string, o
...
+		return minio.ObjectInfo{}, err
+	}
+	if opts.UserDefined == nil {
+		opts.UserDefined = make(map[string]string)
+	}
+	for k, v := range objMeta {
+		opts.UserDefined[k] = v
+	}
+	contentType := utils.GuessMimeType(object)
+	contentType = n.getObjContentType(objMeta, contentType)
```
zhijian-pro, line 709: The code here is strange, the function `getObjContentType` is only used here, it doesn't need to be a separate function.

pkg/gateway/gateway.go
```diff
@@ -871,6 +900,73 @@ const s3Etag = "s3-etag"
 // less than 64k ref: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html#tag-restrictions
 const s3Tags = "s3-tags"
 
+// S3 object metadata
+const s3Meta = "s3-meta"
+const amzMeta = "x-amz-meta-"
+const metaContentType = "content-type"
```
zhijian-pro, line 906: There is no need to define a constant

zhijian-pro (changes requested):

pkg/gateway/gateway.go
prOOrc: Inlined.

pkg/gateway/gateway.go
prOOrc: Done.

zhijian-pro (approved):

jiefenghuang (approved):

PR #5542, opened 2025-01-10
meta/sql: add cleaner metaconns options

anysql:
Add a cleaner "metaconns" option for meta url


jiefenghuang:
A parsing method similar to Redis would be sufficient, no need to introduce additional processing method.

PR #5547, opened 2025-01-12
fix readahead bug then trying to turn off readahead totally

timurvankov:
When you try to turn off readahead, you still get traffic you don't expect.
in this fix we limit readaheadMax to zero in case any readahead is turned off.

The question remains what is the expected behavior in the case of the max-readahead==0 parameter

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -41,6 +41,14 @@ func Min(a, b int) int {
 	return b
 }
 
+// Max returns max of 2 int
+func Max(a, b int) int {
+	if a > b {
+		return a
+	}
+	return b
+}
+
 func Min64(a, b uint64) uint64 {
 	if a < b {
 		return a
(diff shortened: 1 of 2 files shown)


polyrabbit:
Could you elaborate the bug?

Setting `--max-readahead` smaller than `blocksize` will disable readahead, as implemented in #5229 .

timurvankov:
yes, it's true, but if you set 0, the behavior is unexpected, I expected it to be 0 and not block_size * 8. in fact, in general, this is a useless parameter in the end, since in the end, the entire block is going through the network anyway in asynchronous mode for caches

jiefenghuang:
hi @timurvankov，
The default behavior of 0 is 8 blocks to maintain compatibility with the behavior of the previous version. 
Here are some potential optimizations:
1.	Set readahead based on blocks, as mentioned by @polyrabbit here: https://github.com/juicedata/juicefs/pull/5229#issuecomment-2409895370
2.	Keep the default value of the parameter at 8 blocks.

jiefenghuang (changes requested):
TBD

PR #5548, opened 2025-01-12
No need to stat "/" in single bucket mode

polyrabbit:
single bucket mode, almost every request calls stat `/` - it's unnecessary.


zhijian-pro:
In single-bucket mode, a bucket may also be `.minio.sys`, this optimization should exclude this situation.

polyrabbit:
OK, updated.

PR #5549, opened 2025-01-12
gateway: add one more hierarchy to upload tmp/multiupload dir to reduce txn conflicts

polyrabbit:
In the current gateway implementation, the `tmp/upload` directory is a flat structure, which can easily lead to concurrency conflicts under high concurrency. These conflicts severely impact performance and prevent the gateway from scaling. 

We can reduce conflicts in those large directories by adding a layer of hierarchy.

### DIFF
--- a/pkg/gateway/gateway.go
+++ b/pkg/gateway/gateway.go
@@ -194,6 +194,10 @@ func (n *jfsObjects) ppath(bucket, uploadID, part string) string {
 	return n.tpath(bucket, "uploads", uploadID[:subDirPrefix], uploadID, part)
 }
 
+func (n *jfsObjects) ppathFlat(bucket, uploadID, part string) string { // compatible with tmp files uploaded by old versions(<1.2)
+	return n.tpath(bucket, "uploads", uploadID, part)
+}
+
 func (n *jfsObjects) DeleteBucket(ctx context.Context, bucket string, forceDelete bool) error {
 	if err := n.isValidBucketName(bucket); err != nil {
 		return err
@@ -1090,6 +1094,10 @@ func (n *jfsObjects) CompleteMultipartUpload(ctx context.Context, bucket, object
 	for _, part := range parts {
 		p := n.ppath(bucket, uploadID, strconv.Itoa(part.PartNumber))
 		copied, eno := n.fs.CopyFileRange(mctx, p, 0, tmp, total, 5<<30)
+		if eno == syscall.ENOENT { // try lookup from old path
+			p = n.ppathFlat(bucket, uploadID, strconv.Itoa(part.PartNumber))
+			copied, eno = n.fs.CopyFileRange(mctx, p, 0, tmp, total, 5<<30)
+		}
 		if eno != 0 {
 			err = jfsToObjectErr(ctx, eno, bucket, object, uploadID)
 			logger.Errorf("merge parts: %s", err)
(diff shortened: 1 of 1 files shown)


pkg/gateway/gateway.go
```diff
@@ -186,11 +186,11 @@ func (n *jfsObjects) tpath(p ...string) string {
 }
 
 func (n *jfsObjects) upath(bucket, uploadID string) string {
-	return n.tpath(bucket, "uploads", uploadID)
+	return n.tpath(bucket, "uploads", uploadID[:2], uploadID)
 }
 
 func (n *jfsObjects) ppath(bucket, uploadID, part string) string {
-	return n.tpath(bucket, "uploads", uploadID, part)
+	return n.tpath(bucket, "uploads", uploadID[:2], uploadID, part)
```
davies, line 193: this is not compatible to previous version, how can we provide a smooth upgrade?

polyrabbit: I suppose there are no compatibility issues. After upgrade, new objects will be written to the new path, and old objects will be cleaned up by `cleanupDir`.

davies: The running multipart-upload will fail after upgrade

polyrabbit: Do you mean uploading some parts using the old path and some parts using the new path? In this case, the `CompleteMultipartUpload` call will fail, and the user will get an error. Most tools will retry after receiving a failure.

polyrabbit: Commonly rclone will retry the entire upload if `CompleteMultipartUpload` fails multiple times.

For other tools, `CompleteMultipartUpload` will eventually fail, and the user will be notified. I suppose it is acceptable to encounter several failed requests during an upgrade.

davies: It's better to make the upgrade smooth if possible

polyrabbit: Do you have any suggestions?

davies: we can fallback to original path when 1) find uploaded part 2) list parts

davies:
This is a behavior change, I think we should introduce a parameter to enable it.

davies:
The `CompleteMultipartUpload` should also try to lookup parts in different dirs.

polyrabbit:
OK, updated `CompleteMultipartUpload` to also look up in old flat dir when part is not found in new dir. I suppose this can made `CompleteMultipartUpload` smooth during upgrade, and we can avoid one extra flag.

Please review:
https://github.com/juicedata/juicefs/compare/21200b52e86135325cd705138e7a20e6ab0ecd74..4bbe479feefc402a9c50fe309a5a0a680cd69e9a

pkg/gateway/gateway.go
```diff
@@ -1071,6 +1094,12 @@ func (n *jfsObjects) CompleteMultipartUpload(ctx context.Context, bucket, object
 	for _, part := range parts {
 		p := n.ppath(bucket, uploadID, strconv.Itoa(part.PartNumber))
 		copied, eno := n.fs.CopyFileRange(mctx, p, 0, tmp, total, 5<<30)
+		if eno != 0 {
```
davies, line 1097: remove this line

davies (approved):

pkg/gateway/gateway.go
polyrabbit: Done

Issue #5553, opened 2025-01-13
hadoop sdk: verify kerberos authenticated user


kyungwan-nam:
**What would you like to be added**:

I want all file's UID to be consistent in our kerberos Hadoop cluster using mapping username/UID for all kerberos hadoop users.
Currently, The JuiceFS Hadoop SDK authenticate as the OS user if there’s no kerberos credentials. so, can use the JuiceFS as unknown user not in the mapping username/UID. It makes it difficult to ensure consistent.

**Why is this needed**:

I'd like to add a new property ``juicefs.verify-kerberos-user`` (default: false) for JuiceFS Hadoop SDK.
It allows only for kerberos-authenticated user If ``juicefs.verify-kerberos-user`` is true

kyungwan-nam:
Requires kerberos credentials if ``juicefs.verify-kerberos-user`` is true. there's no side effect if false(default)

Could you take a look at PR: https://github.com/juicedata/juicefs/pull/5554

PR #5554, opened 2025-01-13
Verify kerberos authenticated user

kyungwan-nam:
fixes for https://github.com/juicedata/juicefs/issues/5553


kyungwan-nam:
close PR for now.
I think It needs some tests to make sure there's no problem.
I'll again shortly.

kyungwan-nam:
It doesn't work if authenticate using hadoop delegation token.

PR #5557, opened 2025-01-15
Limit buffer usage for read requests to avoid oom

polyrabbit:
Currently, there is no limitation when accepting read requests. Users may issue a large number of concurrent read requests to JuiceFS, causing it to OOM. JuiceFS should protect itself from such usage and provide a predictable memory usage.

image


CodingPoeta:
# reproduce:

```
./juicefs mount redis://localhost /tmp/jfs -d --cache-size=0
fio --name=big-file --rw=read --bs=1MB --size=50M --nrfiles=40 --numjobs=20 --group_reporting --directory=/tmp/jfs/fio-dir --filename_format=\$jobname.\$filenum.\$jobnum --runtime=300 --time_based
```


before:
Screenshot 2025-04-09 at 5 49 18 PM

this pr:
Screenshot 2025-04-09 at 5 49 34 PM

PR #5559, opened 2025-01-15
show warmup cache location info

anysql:
show warmup cache location info (for check mode only)


anysql:
sample output:

image

PR #5563, opened 2025-01-16
cache: fix buffer conflict in writeback mode

polyrabbit:
A bug introduced in #5230

Otherwise, data corruption may happen.
image


jiefenghuang:
Currently, WithTimeout does not interrupt the goroutine after a timeout. It is necessary to set an appropriate value for MaxStageWrite to prevent excessive concurrency under high load or abnormal conditions.

polyrabbit:
Updated

jiefenghuang (approved):

PR #5574, opened 2025-01-17
meta/rename: lock both src and dst parent for rename op to avoid txn conflicts and performance

polyrabbit:
Follow-up on #5549 - after uploading to a temporary file, gateway uses `rename` to move it to the target path. In current implementation, `rename` only locks `parentSrc` (in normal cases). #5549 adds an additional hierarchy to reduce conflicts on `parentSrc`, but conflicts still occur a lot on `parentDst` when user writes to the same directory concurrently.

So we should lock both `parentSrc` and `parentDst` in tkv/sql engines, in consistent with the redis engine.

### DIFF
--- a/pkg/meta/tkv.go
+++ b/pkg/meta/tkv.go
@@ -809,10 +809,7 @@ func (m *kvMeta) txn(ctx context.Context, f func(tx *kvTxn) error, inodes ...Ino
 	}
 	start := time.Now()
 	defer func() { m.txDist.Observe(time.Since(start).Seconds()) }()
-	if len(inodes) > 0 {
-		m.txLock(uint(inodes[0]))
-		defer m.txUnlock(uint(inodes[0]))
-	}
+	defer m.txBatchLock(inodes...)()
 	var lastErr error
 	for i := 0; i < 50; i++ {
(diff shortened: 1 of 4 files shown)


anysql:
In txBatchLock there is problem, need to lock according to txLock id, not by inode.

Get a map of (inode % nlocks), and lock them in order.

By the way, we use optimistic lock for parent dir update without holding the txLock to improve concurrency.

pkg/meta/tkv.go
```diff
@@ -1491,10 +1488,6 @@ func (m *kvMeta) doRename(ctx Context, parentSrc Ino, nameSrc string, parentDst
 	var dtyp uint8
 	var tattr Attr
 	var newSpace, newInode int64
-	lockParent := parentSrc
-	if isTrash(lockParent) {
```
davies, line 1495: we should not lock the source if it's trash, otherwise the `restore --putback --threads 10` will be slow

polyrabbit: OK, updated. I didn't recognize this optimization...

davies (approved):

PR #5598, opened 2025-01-22
fix stats output alignment

anysql:
Values are rounded in Printf

func main() {
	fmt.Printf("99.9 = %4.1f\n", 99.9) // 99.9
	fmt.Printf("99.949 = %4.1f\n", 99.949) // 99.9
	fmt.Printf("99.950 = %4.1f\n", 99.950) // 100.0
	fmt.Printf("99.99 = %4.1f\n", 99.99) // 100.0
}


zhijian-pro:
This change has little impact, but it will make the code look very strange, I do not recommend it.

anysql:
Rounding in printf is much complex than expected.

PR #5631, opened 2025-02-08
cond: stop the timer before calling Reset

SandyXSD:
> // For a Timer created with NewTimer, Reset should be invoked only on
// stopped or expired timers with drained channels.

### DIFF
--- a/pkg/utils/cond.go
+++ b/pkg/utils/cond.go
@@ -45,19 +45,25 @@ func (c *Cond) Broadcast() {
 
 var timerPool = sync.Pool{
 	New: func() interface{} {
-		return time.NewTimer(time.Second)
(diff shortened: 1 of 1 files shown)


pkg/utils/cond.go
```diff
@@ -45,19 +45,25 @@ func (c *Cond) Broadcast() {
 
 var timerPool = sync.Pool{
 	New: func() interface{} {
-		return time.NewTimer(time.Second)
+		t := time.NewTimer(time.Minute)
+		if !t.Stop() {
```
davies, line 49: Why stop a newly created timer?

penglaiyxy: We recently met an abnormality, flush timeout is deadlock on select. We do not know if the abnormality is relevent with this issue?

PR #5638, opened 2025-02-10
refactor: extract Attr's marshal and unmarshal from baseMeta to Attr

shenxingwuying:
I think the Attr's marshal and unmarshal in Attr is better than baseMeta.  Maybe some other basic structure need review later.


jiefenghuang:
Do you need to reference this function?

shenxingwuying:
> Hi @shenxingwuying Do you need to reference this function? Currently, there’s no need to expose this function.

I‘m planning to add a new meta service plugin to make some operations can push to it,
and it's a better way to keep the storage format in juicefs.

jiefenghuang (approved):

PR #5640, opened 2025-02-11
Remove one unnecessary `GetAttr` when deleting from trash

polyrabbit:
Currently, we do not have statistics on the trash directory, so it is unnecessary to update the statistics info for trash directory during unlink, and metadata operations can be eliminated to `Lookup` and `Unlink`.

Before:

img_v3_02j1_96769416-0f9d-44b2-862b-bf01d8a3210g
img_v3_02j1_d40e6752-97ba-4285-a399-b8c4f805032g


davies:
Maybe we should provide statistics for trash ?

polyrabbit:
Even better.
Also it's better to provide a metric for trash usage - like volume usage, so that user knows how much data can be freed.

I updated code to not update quota in trash, stats is not affected.

PR #5642, opened 2025-02-11
object/bos: reduce memory allocation in bos

zhijian-pro:


pkg/utils/dynamic_alloc.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func DynFree(b []byte) {
+	dynPools[powerOf2(cap(b))].Put(&b)
+}
+
+var dynPools []*sync.Pool
+
+func init() {
+	dynPools = make([]*sync.Pool, 33) // 1 - 8G
```
jiefenghuang, line 40: 4G

pkg/utils/alloc.go
```diff
@@ -41,7 +41,7 @@ func Alloc(size int) []byte {
 func Free(b []byte) {
 	// buf could be zero length
 	atomic.AddInt64(&used, -int64(cap(b)))
-	pools[PowerOf2(cap(b))].Put(&b)
+	pools[powerOf2(cap(b))].Put(&b)
```
jiefenghuang, line 44: rounding down here is safer？

pkg/utils/dynamic_alloc.go
```diff
@@ -0,0 +1,62 @@
...
+	return b[:size]
+}
+
+// DynFree b may be longer than the original, in the actual size into the pool
+func DynFree(b []byte) {
+	dynPools[FloorPowerOf2(cap(b))].Put(&b)
+}
+
+func FloorPowerOf2(s int) int {
+	var bits int
```
jiefenghuang, line 39: if s <= 0 {
  return 0
}

Issue #1546, opened 2022-03-10
Support: run JuiceFS in background in Windows


diluga:
JuiceFS does not run in the background of Windows as of now。

https://github.com/juicedata/juicefs/blob/main/cmd/mount_windows.go#L48
```
func makeDaemon(c *cli.Context, name, mp string, m meta.Meta) error {
	logger.Warnf("Cannot run in background in Windows.")
	return nil
}
```
It is recommended to use the following tools to achieve cross-platform compatibility.

https://github.com/kardianos/service

xiaogaozi:
As a workaround, you could use [NSSM](https://nssm.cc) run JuiceFS as a Windows service. There's a [user guide](https://mp.weixin.qq.com/s/nMqCuit4zRoNCK4m-b0hxA) (written in Chinese) show how to do this (thanks @diluga).

chenjie4255:
-d option added by https://github.com/juicedata/juicefs/pull/5643

PR #5643, opened 2025-02-11
Windows: add option to startup juicefs as system service. fix #347, link #1546

### LINKED ISSUE #347: Windows mount for all user
We use the juicefs client on Windows, but only the current session can use the mounted drive. The drive cannot be used by other remote desktop sessions.

chenjie4255:
When the --as-svc mount option is specified, the juicefs mount command updates the Windows registry to prepare the service definition needed by the Winfsp Launcher. After updating the registry, the juicefs mount process invokes net use to trigger the mount behavior.

https://github.com/winfsp/winfsp/wiki/WinFsp-Service-Architecture

Known limitations:
* Only one volume can be mounted at a time when using --as-svc. Mounting more than one volume may cause unexpected behavior depending on the implementation of winfsp.
* We essentially write the same command line to the winfsp service registry value, except for the --as-svc option, including the mount point. Therefore, when users use the “Map a network drive” feature in File Explorer, the mount process will ignore the mount point.

## Test

- [x] Run command with "--as-svr" option to mount volume 1 at x:.
- [x] Disconnect volume 1 mounting from File Explorer.
- [x] Run command with "--as-svr" option to mount volume 2 at y:

### DIFF
--- a/cmd/mount_windows.go
+++ b/cmd/mount_windows.go
@@ -70,7 +70,7 @@ func getDaemonStage() int {
 
 func mountMain(v *vfs.VFS, c *cli.Context) {
 	if c.Bool("as-svc") {
-		winfsp.RunAsSystemSerivce(v.Conf.Format.Name)
+		winfsp.RunAsSystemSerivce(v.Conf.Format.Name, c.Args().Get(1))
 		return
 	}
 
--- a/cmd/mount_windows.go
+++ b/cmd/mount_windows.go
@@ -47,6 +47,10 @@ func mountFlags() []cli.Flag {
 			Name:  "delay-close",
 			Usage: "delay file closing in seconds.",
 		},
+		&cli.BoolFlag{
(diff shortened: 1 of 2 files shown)


cmd/mount_windows.go
```diff
@@ -47,6 +47,10 @@ func mountFlags() []cli.Flag {
 			Name:  "delay-close",
 			Usage: "delay file closing in seconds.",
 		},
+		&cli.BoolFlag{
+			Name:  "as-svc",
```
davies, line 51: --daemon or --background  or --as-service?

chenjie4255: yeah, the as-service would be better.

chenjie4255:
changed to the "-d" and  the "--background" options, and re-ran the previous tests.

To simplify the modification, I did not make changes to the existing stage < 3 code. (Stage was forcibly set to 3 on Windows system at earlier code)

PR #5651, opened 2025-02-12
Delete trash exits timely

polyrabbit:
These warning logs appear frequently during massive deletions. CleanupTrash should exit timely like other deletion routines.

img_v3_02je_95bf4fbe-f1de-45d4-af9e-8d6b67ca33dg


davies:
Are you sure this PR will fix the above logging issue ?

polyrabbit:
This PR won't eliminate the above logs, I suppose it's OK to print occasionally. This PR tries to eliminate trash deletion that won't stop timely. If different deletion routines don't overlap, those logs won't be printed.

image

PR #5652, opened 2025-02-12
Fix panic in log format

polyrabbit:
This panic occurs almost weekly. I don't know why, so only to defend against it.

img_v3_02j7_10006097-0012-484e-ac9d-092a9f49b76g

### DIFF
--- a/pkg/utils/logger.go
+++ b/pkg/utils/logger.go
@@ -19,6 +19,7 @@ import (
 	"io"
 	"os"
 	"path"
+	"runtime"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


davies:
The Caller could be nil, we should check and ignore the function name

polyrabbit:
Yes, This PR adds a defensive check.

pkg/utils/logger.go
```diff
@@ -61,16 +62,20 @@ func (l *logHandle) Format(e *logrus.Entry) ([]byte, error) {
 		lvlStr = fmt.Sprintf("\033[1;%dm%s\033[0m", color, lvlStr)
 	}
 	const timeFormat = "2006/01/02 15:04:05.000000"
+	caller := &runtime.Frame{Function: "???", File: "???", Line: 0}
```
davies, line 65: make it global to avoid allocation

polyrabbit: Done

PR #5683, opened 2025-02-19
gc: improve the performance of Juicefs gc command

### LINKED ISSUE #5671: Improve the performance of Juicefs gc
When dealing with a large number of pending deleted files, we typically use `juicefs gc` to remove these files. However, the juicefs gc command invokes the scanPendingFiles function, where each soft - deleted file is processed sequentially.

https://github.com/juicedata/juicefs/blob/2dd38978aa183d4ce5a70f538afa11c0414b34da/pkg/meta/tkv.go#L2584-L2594

The issue is that if there are a large number of files with relatively small individual sizes, the file deletion speed will drop significantly. Although gc provides the --threads parameter, in reality, this parameter controls the parallel deletion speed within a single file.

I think this is how we can solve it in tkv.go: process each file pending deletion in parallel. The code is provided below. Please help me review it. :)

```go
	batchSize := 1000000

	threads := min(1, m.conf.MaxDeletes/3)
	deleteFileChan := make(chan pair, threads)
	var wg sync.WaitGroup

	for i := 0; i < threads; i++ {
		wg.Add(1)
		go func() {
			defer wg.Done()
			for pair := range deleteFileChan {
				key, value := pair.key, pair.value
				if len(key) != klen {
					logger.Errorf("invalid key %x", key)
					continue
				}
				ino := m.decodeInode([]byte(key)[1:9])
				size := binary.BigEndian.Uint64([]byte(key)[9:])
				ts := m.parseInt64(value)
				clean, err := scan(ino, size, ts)
				if err != nil {
					logger.Errorf("scan pending deleted files: %s", err)
					continue
				}
				if clean {
					m.doDeleteFileData(ino, size)
				}
			}
		}()
	}

	prefixKey := m.fmtKey("D")
	endKey := nextKey(prefixKey)
	for {
		keys, values, err := m.scan(prefixKey, endKey, batchSize, func(k, v []byte) bool {
			return len(k) == klen
		})
		if len(keys) == 0 {
			break
		}
		if err != nil {
			close(deleteFileChan)
			wg.Wait()
			return err
		}
		prefixKey = keys[len(keys)-1]

		for index, key := range keys {
			deleteFileChan <- pair{key, values[index]}
		}

		if len(keys) < batchSize {
			break
		}
	}

	close(deleteFileChan)
	wg.Wait()
	return nil
```

And, I tested the results and found that it can improve the gc performance by 10x in the scenario of deleting small files.

SonglinLife:
Improve the file deletion performance by processing multiple files in parallel

close #5671

### DIFF
--- a/pkg/meta/tkv.go
+++ b/pkg/meta/tkv.go
@@ -2581,20 +2581,19 @@ func (m *kvMeta) scanPendingFiles(ctx Context, scan pendingFileScan) error {
 	// deleted files: Diiiiiiiissssssss
 	klen := 1 + 8 + 8
 
+	var err error
 	m.client.scan(m.fmtKey("D"), func(key, val []byte) {
 		if len(key) != klen {
-			logger.Errorf("invalid key %x", key)
+			err = fmt.Errorf("invalid key %x", key)
 			return
 		}
 		ino := m.decodeInode(key[1:9])
 		size := binary.BigEndian.Uint64(key[9:])
 		ts := m.parseInt64(val)
-		if _, err := scan(ino, size, ts); err != nil {
-			logger.Warnf("scan pending deleted files: %s", err)
-		}
+		_, err = scan(ino, size, ts)
 	})
 
-	return nil
+	return err
 }
 
 func (m *kvMeta) doRepair(ctx Context, inode Ino, attr *Attr) syscall.Errno {
--- a/pkg/meta/tkv.go
+++ b/pkg/meta/tkv.go
@@ -2580,15 +2580,11 @@ func (m *kvMeta) scanPendingFiles(ctx Context, scan pendingFileScan) error {
 	}
 	// deleted files: Diiiiiiiissssssss
 	klen := 1 + 8 + 8
-	batchSize := 100000
 
-	threads := m.conf.MaxDeletes / 3
-	if threads < 1 {
-		threads = 1
-	}
+	threads := m.conf.MaxDeletes
 	deleteFileChan := make(chan pair, threads)
+		
 	var wg sync.WaitGroup
-
 	for i := 0; i < threads; i++ {
 		wg.Add(1)
 		go func() {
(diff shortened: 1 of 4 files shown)


SonglinLife:
I have submitted a pull request for issue #5671, which aims to improve the performance of the gc command, especially when handling a large number of small files. I would appreciate it if someone could take a moment to review it.

pkg/meta/tkv.go
```diff
@@ -2574,28 +2574,67 @@ func (m *kvMeta) scanPendingFiles(ctx Context, scan pendingFileScan) error {
...
+					m.doDeleteFileData(ino, size)
+				}
+			}
+		}()
+	}
+
+	startKey := m.fmtKey("D")
+	endKey := nextKey(startKey)
+	for {
+		keys, values, err := m.scan(startKey, endKey, batchSize, func(k, v []byte) bool {
```
jiefenghuang, line 2614: we can directly use client.scan to avoid copying here.

pkg/meta/tkv.go
```diff
@@ -2574,28 +2574,67 @@ func (m *kvMeta) scanPendingFiles(ctx Context, scan pendingFileScan) error {
...
 	// deleted files: Diiiiiiiissssssss
 	klen := 1 + 8 + 8
-	pairs, err := m.scanValues(m.fmtKey("D"), -1, func(k, v []byte) bool {
-		return len(k) == klen
-	})
-	if err != nil {
-		return err
+	batchSize := 100000
+
+	threads := m.conf.MaxDeletes / 3
```
jiefenghuang, line 2579: adding a new parameter makes it clearer

pkg/meta/tkv.go
jiefenghuang, line 2614: use client.scan directly

cmd/gc.go
```diff
@@ -70,6 +70,11 @@ $ juicefs gc redis://localhost --delete`,
 				Value:   10,
 				Usage:   "number threads to delete leaked objects",
 			},
+			&cli.IntFlag{
+				Name:  "cleanup-threads",
```
davies, line 74: I think we can reuse the `threads` for cleanup

pkg/meta/tkv.go
jiefenghuang: The scan implementation in TiKV and FDB already handles this situation by fetching in batches, but etcd doesn’t. Let’s keep it like this then.

SonglinLife: fully understand :)

jiefenghuang (approved):

jiefenghuang:
@SonglinLife  please fix the lint
```
 Error: pkg/meta/tkv.go:2585:15: Error return value of `m.client.scan` is not checked (errcheck)
  	m.client.scan(m.fmtKey("D"), func(key, val []byte) {
```

SonglinLife:
@jiefenghuang Hi Jiefeng,

Thank you for reviewing the PR. I noticed the golangci-lint check has failed with this error:
```
pkg/meta/tkv.go:2585:15: Error return value of m.client.scan is not checked (errcheck)
m.client.scan(m.fmtKey("D"), func(key, val []byte) {
```

Should I fix this lint issue before merging? I can update the PR immediately if needed.

Thanks!

jiefenghuang:
Yes, all checks must pass before merging the pull request. The ‘sdktest’ issue might be environment-related and can be ignored for now.

SonglinLife:
I have push a new commit, fix this lint error.

jiefenghuang:
@SonglinLife Our colleague has fixed sdktest. Could you rebase to main and push again with –force-with-lease?

SonglinLife:
I have rebased on main and pushed.

PR #5684, opened 2025-02-19
Add Threads and ListThreads progress bar into `sync`

umialpha:
The `sync` command has a `--threads` parameter, which is described as "number of concurrent threads." 

This description is too simplistic. The default value is 10, and before diving into the code, users might try values like 20 or 30, but it’s unlikely they would consider increasing it to 500, which could result in a significant performance boost.

After reviewing the code, I found that all concurrency in sync is affected by this parameter. When I set it to 500, the sync speed increased dramatically from 10MB/s to 10GB/s. Therefore, I suggest **adding a display that shows the current available number of threads**. If this number remains at 0, and the user has performance requirements, they would know to increase this parameter.

### DIFF
--- a/pkg/sync/sync.go
+++ b/pkg/sync/sync.go
@@ -60,6 +60,8 @@ var (
 	listedPrefix          *utils.Bar
 	concurrent            chan int
 	limiter               *ratelimit.Bucket
+	threadsBar            *utils.Bar
+	listThreadsBar        *utils.Bar
 )
 var crcTable = crc32.MakeTable(crc32.Castagnoli)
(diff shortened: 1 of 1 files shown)


zhijian-pro:
The Pending objects progress bar can also help with tuning.
If pending is 0, it means the list request is falling behind. If pending is very large, it means the transmission is falling behind. Usually, the transmission is falling behind.

davies:
The listing and copying should run with the number of threads as configured, unless the other one is the bottleneck, which could be told by the number of pending objects.

So I think these two progress bar is not that useful.

Issue #5687, opened 2025-02-20
.stats file does not show `juicefs_object_request_data_bytes_GET` correctly


SonglinLife:
`.stats` is an internal file used to collect metrics in a simple but useful way. 

Howerver, #4609 add a label `storage_class` for metrics `juicefs_object_request_data_bytes`, which made the .stats file collect metrics incorrectly.

https://github.com/juicedata/juicefs/blob/c92080423872418be921445d4fb243861b85c3e1/pkg/vfs/internal.go#L166-L181

When the juicefs_object_request_data_bytes metric has multiple storage_class types, such as:
```
In the .stats file:
juicefs_object_request_data_bytes_GET 99108970949790
juicefs_object_request_data_bytes_GET 93296720741168

In the Prometheus endpoint:
juicefs_object_request_data_bytes{instance="xxx",juicefs_version="xxxx",method="GET",mp="/juicefs",storage_class="IA",vol_name="juicefs"} 9.910897094979e+13
juicefs_object_request_data_bytes{instance="xx",juicefs_version="xxxx",method="GET",mp="/juicefs",storage_class="Standard",vol_name="juicefs"} 9.3296720741168e+13
```
The .stats file ends up with duplicate entries for the same metric juicefs_object_request_data_bytes_GET

PR #5688, opened 2025-02-20
fix .stats file collect metric incorrectly

SonglinLife:
construct metric name by excluding common static labels

ref: https://github.com/juicedata/juicefs/issues/5687

### DIFF
--- a/pkg/vfs/internal.go
+++ b/pkg/vfs/internal.go
@@ -149,6 +149,13 @@ func getInternalNodeByName(name string) *internalNode {
 	return nil
 }
 
+var commonLabels = map[string]struct{}{
+	"instance":        {},
+	"juicefs_version": {},
+	"mp":              {},
+	"vol_name":        {},
+}
+
 func collectMetrics(registry *prometheus.Registry) []byte {
 	if registry == nil {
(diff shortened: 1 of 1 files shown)


polyrabbit:
`stats` cmd should be robust enough to handle those duplicated metrics after PR #4941 .

Will this tool work correctly after adding `storage_class` to metric name or other labels added in the future?

SonglinLife:
Yes, I didn't realize that the `stats` tool also relies on the `.stats` file. I recently discovered this issue because I manually collected the .stats file to build a monitoring dashboard. 

It’s better to keep it this way for now, as I noticed that the metrics' names are hardcoded in the `stats` code.

jiefenghuang:
@SonglinLife Prometheus metrics can be collected through 127.0.0.1:9567

SonglinLife:
> @SonglinLife Prometheus metrics can be collected through 127.0.0.1:9567

Yes, but for some reason, it is difficult for me to collect Prometheus metrics from a host port.

jiefenghuang:
As polyrabbit mentioned, after making this modification, stats cmd may not work correctly.

PR #5692, opened 2025-02-21
cache: inherit atime and track logic size when re-scan

polyrabbit:
Scanning from disk overwrites `size` and `atime` of existing cache items in memory:

1. `size` is overwritten by the physical size of the file, which includes the logical size plus the CRC footer. However, when adding cache in other places, we use the logical size. I suppose it's better to be consistent on which size to use.

2. The granularity of file system's `atime` is relatively coarse (relatime), and some even have atime disabled (noatime). However, `atime` recorded in memory has finer granularity and is more accurate in most cases. Overwriting `atime` in memory blindly would cause the expire eviction strategy not working as expected.

### DIFF
--- a/pkg/chunk/disk_cache.go
+++ b/pkg/chunk/disk_cache.go
@@ -886,10 +886,7 @@ func (cache *cacheStore) uploadStaging() {
 func (cache *cacheStore) scanCached() {
 	cache.Lock()
 	cache.used = 0
-	lastAccessed := make(map[cacheKey]uint32, len(cache.keys))
-	for k, v := range cache.keys { // atime in memory is more accurate than on disk, inherit it for the next round
-		lastAccessed[k] = v.atime
-	}
+	lastAccessed := cache.keys // atime in memory is more accurate than on disk, inherit it for the next round
 	cache.keys = make(map[cacheKey]cacheItem)
 	cache.scanned = false
 	cache.Unlock()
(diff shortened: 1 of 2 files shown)


pkg/chunk/disk_cache.go
```diff
@@ -609,16 +609,20 @@ func (cache *cacheStore) load(key string) (ReadCloser, error) {
...
 	})
 
 	cache.Lock()
 	if err == nil {
+		// update atime
 		if it, ok := cache.keys[k]; ok {
-			// update atime
 			cache.keys[k] = cacheItem{it.size, uint32(time.Now().Unix())}
+		} else {
+			cache.keys[k] = cacheItem{int32(size), uint32(time.Now().Unix())}
```
jiefenghuang, line 624: cache file maybe removed in cleanupExpire after open

polyrabbit: I suppose it OK for this rare case. Periodical scan will align cache in disk and memory.

jiefenghuang: load will return error, and if scanInterval is set, we dont need to add key here.

polyrabbit: When scanning from local disk, if a read request comes in, this insertion will ensure `atime` is accurate, which is exactly what this PR is trying to solve.

Besides potential inconsistency with `cleanupExpire` for a short while, are there any other concerns?

polyrabbit:
Right, `atime` in memory should be passed to the next round, I missed that.

polyrabbit:
Updated code and added a UT to ensure `atime` is not lost between scans. (During the scan, `atime` caused by a read request may still be lost.)

pkg/chunk/disk_cache.go
```diff
@@ -914,10 +915,18 @@ func (cache *cacheStore) scanCached() {
 					key = strings.ReplaceAll(key, "\\", "/")
 				}
 				atime := uint32(getAtime(fi).Unix())
+				if memAtime := lastAccessed[cache.getCacheKey(key)].atime; memAtime > atime {
```
jiefenghuang, line 918: map may conflict

jiefenghuang (approved):

PR #5705, opened 2025-02-28
sync: delay deleting directories to avoid deleting empty directories

### LINKED ISSUE #5698: Some empty directories will be leaked after sync
sync tries to remove a directory before removing all the children, will fail with ENOTEMPTY and be leaked.

zhijian-pro:
close #5698

### DIFF
--- a/pkg/sync/sync.go
+++ b/pkg/sync/sync.go
@@ -924,6 +924,9 @@ func (o *withFSize) Size() int64 {
 	return o.nsize
 }
 
+var dstDelayDel []string
+var srcDelayDel []string
+
 func handleExtraObject(tasks chan<- object.Object, dstobj object.Object, config *Config) bool {
 	handled.IncrTotal(1)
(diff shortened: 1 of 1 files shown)


pkg/sync/sync.go
```diff
@@ -933,7 +936,11 @@ func handleExtraObject(tasks chan<- object.Object, dstobj object.Object, config
 		return false
 	}
 	config.Limit--
-	tasks <- &withSize{dstobj, markDeleteDst}
+	if dstobj.IsDir() {
+		dstDelayDel = append(dstDelayDel, dstobj.Key())
```
davies, line 940: not thread safe

pkg/sync/sync.go
```diff
@@ -1628,6 +1639,22 @@ func Sync(src, dst object.ObjectStorage, config *Config) error {
 		}()
 	}
 
+	delayDelFunc := func(storage object.ObjectStorage, keys []string) {
+		for i := len(keys) - 1; i >= 0; i-- {
```
davies, line 1643: add a logging if keys is not empty

PR #5706, opened 2025-02-28
meta/sql: fix flush quotas deadlock

### LINKED ISSUE #5695: Deadlock in a Loop During High-Concurrency File Updates
**What happened**:
We have encountered several deadlock issues with JFS in our environment, which consists of AWS RDS and AWS S3. Specifically, 500 nodes are simultaneously writing logs to JFS, and these logs are updated in real-time.
We also enable quota in JFS.
I don't know why deadlock happen, want to fix it.
Launch 500 aws ec2 with JFS mounted, start a python application on each of them, keep writing log into JFS.
We may destroy ec2 instance, and launch new ec2 instance from time to time.
here is some monitor data:
Image

Image

Image

Process 25448 waits for ExclusiveLock on tuple (2,32) of relation 24708 of database 5; blocked by process 31328.
Process 31328 waits for ShareLock on transaction 685022274; blocked by process 31374.
Process 31374 waits for ExclusiveLock on tuple (5,57) of relation 24708 of database 5; blocked by process 31902.
Process 31902 waits for ShareLock on transaction 685023299; blocked by process 25448.
 

**Environment**:
- JuiceFS version (use `juicefs --version`) or Hadoop Java SDK version: juicefs version 1.1.1+2023-11-28.437f4e6
- Cloud provider or hardware configuration running JuiceFS: aws RDS(postgresql) + aws s3
- OS (e.g `cat /etc/os-release`): VERSION="22.04.4 LTS (Jammy Jellyfish)"
- Kernel (e.g. `uname -a`): 6.5.0-1024-aws
- Object storage (cloud provider and region, or self maintained): aws s3, us-east-1
- Metadata engine info (version, cloud provider managed or self maintained): aws RDS(postgresql)
- Network connectivity (JuiceFS to metadata engine, JuiceFS to object storage): aws internal network
- Others:

jiefenghuang:
close #5695

### DIFF
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -96,7 +96,7 @@ type engine interface {
 	doSetQuota(ctx Context, inode Ino, quota *Quota) (created bool, err error)
 	doDelQuota(ctx Context, inode Ino) error
 	doLoadQuotas(ctx Context) (map[Ino]*Quota, error)
-	doFlushQuotas(ctx Context, quotas map[Ino]*Quota) error
+	doFlushQuotas(ctx Context, quotas []*iQuota) error
 
 	doGetAttr(ctx Context, inode Ino, attr *Attr) syscall.Errno
 	doSetAttr(ctx Context, inode Ino, set uint16, sugidclearmode uint8, attr *Attr) syscall.Errno
--- a/pkg/meta/sql.go
+++ b/pkg/meta/sql.go
@@ -3526,11 +3526,12 @@ func (m *dbMeta) doLoadQuotas(ctx Context) (map[Ino]*Quota, error) {
 	return quotas, nil
 }
 
-func (m *dbMeta) doFlushQuotas(ctx Context, quotas map[Ino]*Quota) error {
+func (m *dbMeta) doFlushQuotas(ctx Context, quotas []*iQuota) error {
+	sort.Slice(quotas, func(i, j int) bool { return quotas[i].inode < quotas[j].inode })
 	return m.txn(func(s *xorm.Session) error {
-		for ino, q := range quotas {
+		for _, q := range quotas {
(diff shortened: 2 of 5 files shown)


pkg/meta/base.go
```diff
@@ -96,7 +96,7 @@ type engine interface {
 	doSetQuota(ctx Context, inode Ino, quota *Quota) (created bool, err error)
 	doDelQuota(ctx Context, inode Ino) error
 	doLoadQuotas(ctx Context) (map[Ino]*Quota, error)
-	doFlushQuotas(ctx Context, quotas map[Ino]*Quota) error
+	doFlushQuotas(ctx Context, quotas []*iQuota) error
```
davies, line 99: is this necessary?

pkg/meta/sql.go
```diff
@@ -3526,11 +3526,12 @@ func (m *dbMeta) doLoadQuotas(ctx Context) (map[Ino]*Quota, error) {
 	return quotas, nil
 }
 
-func (m *dbMeta) doFlushQuotas(ctx Context, quotas map[Ino]*Quota) error {
+func (m *dbMeta) doFlushQuotas(ctx Context, quotas []*iQuota) error {
+	sort.Slice(quotas, func(i, j int) bool { return quotas[i].inode < quotas[j].inode })
```
davies, line 3530: Is sorting for sql enough?

jiefenghuang: Redis and kv do not have such issues; key sorting may improve performance for kv, but it increases the overhead on the client side.

pkg/meta/base.go
jiefenghuang: Use slices directly to avoid an additional map conversion

Issue #5693, opened 2025-02-21
Windows: Cannot set a file as read-only


chenjie4255:
**What happened**:

Cannot set a file as read-only.

**How to reproduce it (as minimally and precisely as possible)**:

Right-click a file in File Explorer, check the "Read-Only" option and save. Reopen the file property dialog the Read-only checkbox is still unselected.

PR #5722, opened 2025-03-05
Windows: add win attrs support;

chenjie4255:
close https://github.com/juicedata/juicefs/issues/5693

## Background

* When setting attributes for a file or folder, winfsp call juicefs.chflags() and pass the attribute flags to juicefs.
* According to this [article,](https://support.microsoft.com/en-us/topic/you-cannot-view-or-change-the-read-only-or-the-system-attributes-of-folders-in-windows-server-2003-in-windows-xp-in-windows-vista-or-in-windows-7-55bd5ec5-d19e-6173-0df1-8f5b49247165) on windows sytem, the readonly attribute for a folder basically doesn't have any effect. 

image

## Test

run ```winfsp-test --fuse-external create*```

* before this PR, it failed for the create_fileattr_test and create_readonly_test.
* after this PR, the failed tests passed. [tested on redis and sqlite, **didn't** test on tikv]

## Discussion

* Maybe we can ignore the FlagWindowsArchive? This is not used by any modern system accroding to this [topic](https://groups.google.com/g/winfsp/c/9QYSSxWo2r8). but if we want to align the behavior of samba client for windows, this should be kept.

### DIFF
--- a/pkg/winfsp/winfs.go
+++ b/pkg/winfsp/winfs.go
@@ -374,6 +374,18 @@ func attrToStat(inode Ino, attr *meta.Attr, stat *fuse.Stat_t) {
 	stat.Size = int64(size)
 	stat.Blocks = int64(blocks)
 	stat.Rdev = uint64(rdev)
+	if attr.Flags&meta.FlagImmutable != 0 {
+		stat.Flags |= fuse.UF_READONLY
+	}
+	if attr.Flags&meta.FlagWindowsHidden != 0 {
+		stat.Flags |= fuse.UF_HIDDEN
+	}
+	if attr.Flags&meta.FlagWindowsSystem != 0 {
+		stat.Flags |= fuse.UF_SYSTEM
+	}
+	if attr.Flags&meta.FlagWindowsArchive != 0 {
+		stat.Flags |= fuse.UF_ARCHIVE
+	}
 }
 
 func (j *juice) h2i(fh *uint64) meta.Ino {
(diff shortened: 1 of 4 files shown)


pkg/meta/redis.go
```diff
@@ -1296,8 +1296,10 @@ func (m *redisMeta) doMknod(ctx Context, parent Ino, name string, _type uint8, m
 		if st := m.Access(ctx, parent, MODE_MASK_W|MODE_MASK_X, &pattr); st != 0 {
 			return st
 		}
-		if (pattr.Flags & FlagImmutable) != 0 {
-			return syscall.EPERM
+		if runtime.GOOS != "windows" {
```
davies, line 1299: the file should be readonly in all system, same as append-only

pkg/winfsp/winfs.go
```diff
@@ -641,6 +653,46 @@ func (j *juice) Releasedir(path string, fh uint64) (e int) {
...
+
+	ctx := j.newContext()
+	fi, err := j.fs.Stat(ctx, path)
+	if err != 0 {
+		e = -fuse.ENOENT
+		return
+	}
+
+	ino := fi.Inode()
+	entry, errrno := j.vfs.GetAttr(ctx, ino, 0)
```
davies, line 667: this is not needed

chenjie4255:
In the latest commit, the behavior of the read-only attribute has been aligned with the unix system (even for directories).

PR #5737, opened 2025-03-06
meta/sql: adjust the database type from int to bigint to store a 32-bit guid

zhijian-pro:


davies:
No, we should fix the python sdk instead.

Issue #4198, opened 2023-11-26
Upgrade zstd package to 1.5.5


solracsf:
**What would you like to be added**:

https://github.com/DataDog/zstd/releases/latest

**Why is this needed**:

A lot of speed and memory improvements on v1.5.4 and v1.5.5

https://github.com/facebook/zstd/blob/v1.5.5/CHANGELOG
https://github.com/facebook/zstd/releases/tag/v1.5.4

davies:
I checked the changelog, no much inprovement for JuiceFS since we only use the level 1 block mode in x86/arm

PR #5741, opened 2025-03-07
Upgrade zstd package to v1.5.6

Xuyuchao-juice:


SandyXSD:
Why we need this?

Xuyuchao-juice:
> Why we need this?

because of https://github.com/juicedata/juicefs/issues/4198

davies:
Trino and other software are using the latest release, which will be conflict with older version.

PR #5758, opened 2025-03-11
docs: use same redis url in docs

mt5225:


mt5225:
the destroy command should match the `quick start`, which is
` juicefs format --storage s3     --bucket https://xxx     redis://default@127.0.0.1:6379`

```shell
juicefs status redis://default@127.0.0.1:6379/1 

## error
2025/03/11 00:40:42.589914 juicefs[743685] <FATAL>: load setting: database is not formatted, please run `juicefs format ...` first [status.go:96]
```

davies:
please fix the one in Chinese also.

PR #5761, opened 2025-03-11
Fix the issue that the username cannot contain the "@" character

davidzhangbj:
If the username contains the "@" character, JuiceFS cannot correctly identify the username. I want to add support for usernames that contain the "@" character.
As shown in the following figure, the username is "user@name", but JuiceFS recognizes the username as "user", which results in the inability to log in to the database.
image
After the modification, the user name without the @ character and the user name with the @ character can be recognized.
username contains '@' characters
image
username does not contains '@' characters
image

### DIFF
--- a/pkg/meta/interface.go
+++ b/pkg/meta/interface.go
@@ -541,7 +541,7 @@ func Register(name string, register Creator) {
 }
 
 func setPasswordFromEnv(uri string) (string, error) {
-	atIndex := strings.Index(uri, "@")
+	atIndex := strings.LastIndex(uri, "@")
 	if atIndex == -1 {
 		return "", fmt.Errorf("invalid uri: %s", uri)
 	}


pkg/meta/interface.go
```diff
@@ -541,7 +541,7 @@ func Register(name string, register Creator) {
 }
 
 func setPasswordFromEnv(uri string) (string, error) {
-	atIndex := strings.Index(uri, "@")
+	atIndex := strings.LastIndex(uri, "@")
```
davies, line 544: The password may also contain `@`

zhijian-pro:
This issue only affects passwords containing the @ symbol set through environment variables.
image

pkg/meta/interface.go
davidzhangbj: Because I modified the setPasswordFromEnv method，does not involve situations where the password contains the @ character.I tested the situation where the password contains the @ character and it is ok.
image

PR #5762, opened 2025-03-11
Show method label on transaction conflict metric

polyrabbit:
`txRestart` happens a lot in a busy cluster. Some are expected (incrCounter of usedSpace), some are not (doWrite) - they all mix together in the same metric, we cannot set up an alert for the latter.

This PR adds few labels to `txRestart` metric. It helps to understand the reason of transaction conflicts and gives a hint how to avoid/fix them, e.g. conflicts:

- `doCleanupDelayedSlices` - `cleanupDelayedSlices` doesnot respect time limits as expected, it can be fixed by modifying code, see #5463.
- `doMknod` - user is creating in the same directory simultaneously, this conflict hurts performance, and can be avoided sometimes, see #5549, #5330.
- `doWrite`/`doTruncate` - user is modifying the same file from different nodes accidentally. He must be notified to avoid data corruption and even `EIO`.

etc.

### DIFF
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -259,7 +259,7 @@ type baseMeta struct {
 	totalSpaceG  prometheus.Gauge
 	totalInodesG prometheus.Gauge
 	txDist       prometheus.Histogram
-	txRestart    prometheus.Counter
+	txRestart    *prometheus.CounterVec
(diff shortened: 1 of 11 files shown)


davies:
How many labels will be added by this PR?

PR #5780, opened 2025-03-15
CI: fix download & upload random test example database

zhoucheng361:


zhoucheng361:
Ref: https://github.com/actions/download-artifact/issues/148

PR #5786, opened 2025-03-17
Better to close session and db connection explicitly in gateway

polyrabbit:
To explicitly release resources.


pkg/gateway/gateway.go
```diff
@@ -102,7 +102,17 @@ func (n *jfsObjects) IsReady(_ context.Context) bool {
 }
 
 func (n *jfsObjects) Shutdown(ctx context.Context) error {
-	return n.fs.Close()
+	if err := n.fs.Close(); err != nil {
+		logger.Warnf("Close fs error: %s", err)
+	}
+	if err := n.fs.Meta().CloseSession(); err != nil {
```
davies, line 108: see  cmd/gateway.go L260

PR #5788, opened 2025-03-18
Allow kernel to cache negative dir entry lookups

polyrabbit:
When `lookup` a non-existent file, current code only returns the error. However, kernel allows an empty dentry with a timeout to be returned, which acts as a negative lookup cache.
  
One use case of this negative cache in our environment is to accelerate `ffmpeg` - it looks up many non-existent files in the working directory. With hundreds of instances, those negative lookups cause lots of overhead to meta engine, and slows down all `ffmpeg` instances.

Also, `libfuse` has a similar option `negative_timeout` - https://man7.org/linux/man-pages/man8/mount.fuse3.8.html

### DIFF
--- a/cmd/flags.go
+++ b/cmd/flags.go
@@ -354,6 +354,10 @@ func metaCacheFlags(defaultEntryCache float64) []cli.Flag {
 			Value: "1.0s",
 			Usage: "dir entry cache timeout",
 		},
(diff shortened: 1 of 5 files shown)


jiefenghuang (approved):
same as negativeEntry in go-fuse

PR #5809, opened 2025-03-23
Access log also displays caller's proc name

polyrabbit:
Pid in the access log does not provide much intuitive information, especially when reviewing the access log afterward. A more intuitive approach is to display the caller's process name.

This process name is particularly helpful when investigating accidental deletions afterward.

img_v3_02kk_2d5af1a3-6eee-40eb-9026-aa1a6816e8fg


davies:
Access log is different from error log, each process should generate separate access log rather than mixing together, then the process name is useless.

polyrabbit:
How do you separate access logs for different processes? In current implementation, access log can only be collected by `cat .accesslog`, which means all processes' access logs are mixed together.

PR #5853, opened 2025-03-25
sdk/python: quota get/list

CodingPoeta:


sdk/python/juicefs/juicefs/juicefs.py
```diff
@@ -345,6 +345,16 @@ def clone(self, src, dst, preserve=False):
         """Clone a file."""
         self.lib.jfs_clone(c_int64(_tid()), c_int64(self.h), _bin(src), _bin(dst), c_bool(preserve))
 
+    def quota(self, cmd, path="", capacity=0, inodes=0, strict=False, repair=False, create=False):
```
davies, line 348: it's better to have friendly API,  get_quota, list_quota

PR #5862, opened 2025-03-26
windows: fixed debug command

chenjie4255:
## keypoints

* use wmic to replace the ps command
* Fixed the permission issue for the .config file


zhijian-pro:
@chenjie4255  The newly added file is missing a License.

zhijian-pro:
Is there no item check for this in our CI?

Issue #5860, opened 2025-03-26
The sync progress bar displays a problem, and the ERROR log is refreshed


YunhuiChen:
**What happened**:
```
root@chaos-k8s-001:~/chenyunhui/juicefs# meta_url=$META_URL meta_url2=$META_URL2 ./juicefs sync --match-full-path jfs://meta_url jfs://meta_url2 --manager-addr 172.28.39.187:8081 --worker chaos-k8s-002,chaos-k8s-003 --dirs --links --check-all --list-threads 10 --list-depth 6
2025/03/26 11:02:12.368407 juicefs[608944] <INFO>: Meta address: mysql://root:****@(172.28.39.187)/test [NewClient@interface.go:578]
2025/03/26 11:02:12.371867 juicefs[608944] <INFO>: Meta address: redis://172.28.39.187:6379/1 [NewClient@interface.go:578]
2025/03/26 11:02:12.372474 juicefs[608944] <WARNING>: AOF is not enabled, you may lose data if Redis is not shutdown properly. [checkRedisInfo@info.go:84]
2025/03/26 11:02:12.372592 juicefs[608944] <INFO>: Ping redis latency: 52.711µs [checkServerConfig@redis.go:3653]
2025/03/26 11:02:12.373970 juicefs[608944] <INFO>: Prometheus metrics listening on 127.0.0.1:39927 [exposeMetrics@mount.go:135]
2025/03/26 11:02:12.374755 juicefs[608944] <INFO>: Listen at 172.28.39.187:8081 [startManager@cluster.go:219]
2025/03/26 11:02:12.374808 juicefs[608944] <INFO>: Syncing from jfs://juicefs/ to jfs://juicefs/ [Sync@sync.go:1657]
2025/03/26 11:02:12.680290 juicefs[608944] <INFO>: launch a worker on chaos-k8s-003 [launchWorker@cluster.go:309]
2025/03/26 11:02:12.680357 juicefs[608944] <INFO>: launch a worker on chaos-k8s-002 [launchWorker@cluster.go:309]
 Scanned objects: 13/453 [=>------------------------------------------------------------]  21.6/s ETA: 20s
Excluded objects: 0                              0.0/s                                   ⠼
  Excluded bytes: 0.0 b    (0 Bytes)             0.0 b/s                                 ⠼
 Skipped objects: 13                             21.6/s                                  ⠼
 Scanned objects: 200/3672 [==>-----------------------------------------------------------]  34.7/s ETA: 1m40s
[...]

PR #5874, opened 2025-03-26
cmd/sync: fix bug, parse and relay remote worker logs with correct le…

Xuyuchao-juice:
rel https://github.com/juicedata/juicefs/issues/5860


Xuyuchao-juice:
image

PR #5881, opened 2025-03-27
Fix fuse_fd_comm was never cleaned up

polyrabbit:
When main thread exits, deferred statements in child threads won't execute.

### DIFF
--- a/cmd/mount_unix.go
+++ b/cmd/mount_unix.go
@@ -775,6 +775,7 @@ func launchMount(mp string, conf *vfs.Config) error {
 	}
 	os.Setenv("_FUSE_FD_COMM", serverAddress)
 	serveFuseFD(serverAddress)
+	defer os.Remove(serverAddress)
 
 	path, err := os.Executable()
 	if err != nil {
--- a/cmd/passfd.go
+++ b/cmd/passfd.go
@@ -162,7 +162,6 @@ func serveFuseFD(path string) {
 		return
 	}
 	go func() {
-		defer os.Remove(path)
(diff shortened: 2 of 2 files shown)


TheClonerx:
My question is what happens if the process is killed by a deadly signal, such as `SIGSEGV` or even `SIGKILL`.
The socket will not be cleaned up then and once the PID number gets recycled this would still cause an `EADDRINUSE`.
What's more is that because this is FUSE, the socket can be created by a any user which then another user cannot remove.

I'm questioning if using a socket under `/tmp` using the PID number as the identifier was good choice to begin with, a much better option would've been using abstract UNIX socket instead, or else use a user specific directory such as `/run/user/$(id -u)/` with a truly random name (akin to `mkstemp`).

davies:
> My question is what happens if the process is killed by a deadly signal, such as `SIGSEGV` or even `SIGKILL`. The socket will not be cleaned up then and once the PID number gets recycled this would still cause an `EADDRINUSE`. What's more is that because this is FUSE, the socket can be created by a any user which then another user cannot remove.
> 
> I'm questioning if using a socket under `/tmp` using the PID number as the identifier was good choice to begin with, a much better option would've been using abstract UNIX socket instead, or else use a user specific directory such as `/run/user/$(id -u)/` with a truly random name (akin to `mkstemp`).

Sounds good, can you send a PR for it?

TheClonerx:
> can you send a PR for it?
I'm afraid I cannot. I'm not familiar enough with Go, and even then, due to the CLA, I'm not contractually allowed.

PR #5894, opened 2025-03-31
cmd/sync: use shortcut paths for files in "files-from"

zhijian-pro:


zhijian-pro:
@YunhuiChen  The test needs to be improved, do not have prefix overlap in the files-from file.

cmd/object.go
```diff
@@ -207,12 +208,22 @@ func (o *jObj) Mode() os.FileMode    { return o.fi.Mode() }
...
+			return eno
+		}
 	}
+	fi, eno := j.jfs.Lstat(ctx, j.path(key))
 	if eno != 0 {
-		return nil, eno
+		return nil, errConv(eno)
+	}
+	if fi.IsSymlink() {
+		fi, eno = j.jfs.Stat(ctx, j.path(key))
```
davies, line 223: we should return return for IsSymlink

davies (approved):

PR #5895, opened 2025-03-31
meta: fix missing attributes in ReadDirPlus

### LINKED ISSUE #5880: windows：Performance Regression: Slow Directory Loading with Large Folders in Main Branch
**What happened**:

When opening directories containing a large number of files/folders, the main branch shows significant performance degradation compared to the official release (v1.2.3).

Should match v1.2.3 performance where such directories open instantly

Checkout latest main branch 

Build with window11 and go1.23.3 windows/amd64

Open a directory containing ~6000 folders

Observe ~10 second loading delay

**Environment**:
- Kernel (e.g. `uname -a`):
- Others:

jiefenghuang:
close #5880

The attributes of the “init entries” are not full, causing winfs to re-execute getattr for each file.

### DIFF
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -3089,12 +3089,24 @@ func (m *baseMeta) NewDirHandler(ctx Context, inode Ino, plus bool, initEntries 
 	initEntries = append(initEntries, &Entry{
 		Inode: inode,
 		Name:  []byte("."),
-		Attr:  &Attr{Typ: TypeDirectory},
-	}, &Entry{
+		Attr:  &attr,
+	})
+
+	parent := &Entry{
 		Inode: attr.Parent,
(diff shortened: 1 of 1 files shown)


chenjie4255 (commented):
Would this cause multiple queries? One during NewDirHandler, and then another one when calling the List() method?

jiefenghuang:
> Would this cause multiple queries? One during NewDirHandler, and then another one when calling the List() method?

Currently yes. In SQL/TKV, parent’s attributes is challenging to integrate into the streaming retrieval process. However, if an attribute cache is available, this issue can be disregarded.

chenjie4255 (approved):
confirmed that the https://github.com/juicedata/juicefs/issues/5880 was fixed by this commit.

PR #5914, opened 2025-04-03
sdk/python: support no_agent & pyroscope

CodingPoeta:


sdk/java/libjfs/main.go
```diff
@@ -346,6 +348,8 @@ type javaConf struct {
 	DirEntryTimeout   string `json:"dirEntryTimeout"`
 	Debug             bool   `json:"debug"`
 	NoUsageReport     bool   `json:"noUsageReport"`
+	NoAgent           *bool  `json:"noAgent"`
```
davies, line 351: why pointer?

CodingPoeta: For compatible with java sdk, It will be nil if not set.

PR #5917, opened 2025-04-03
windows: fix rename bugs

### LINKED ISSUE #5871: windows: An io error occurred while changing the case of directory rename
**What happened**:
Change the directory name T to t:

Image

**Environment**:
- Kernel (e.g. `uname -a`):
- Others:

chenjie4255:
fix #5871 

This PR adjusts the following behaviors:

* Fixed a crash issue on Windows when renaming a.txt to A.TXT. The crash was caused by the original line 1736. see comment.
* After fixing the crash, it also fixed the issue where renaming had no effect (because the code returned early when ino == dino).
* Fixed a problem where the file name became all uppercase after renaming in File Explorer. (Solved by implementing the getpath interface.)
* Adjusted the original --access-log feature: instead of logging from the FS layer, it now logs from the WinFS layer.
* Added a case-sensitive parameter. (It is disabled by default, and can be enabled when needed.)

## Known issue:

* When case-sensitive is enabled, renaming a file to a similar name with different casing (e.g., a.txt → A.TXT) will cause the file to disappear in File Explorer. The notification sent from the WinFSP kernel contains both the old and new filenames correctly. The reason for the disappearance is currently unknown.

---

## updates 1:

* Setting windows registry hklm\system\currentcontrolset\sessionmanager\kernel\obcaseinsensitive=0 didn't help with the known issue

### DIFF
--- a/cmd/mount_windows.go
+++ b/cmd/mount_windows.go
@@ -41,8 +41,8 @@ func mountFlags() []cli.Flag {
 			Usage: "path of log file when running in background",
 		},
 		&cli.StringFlag{
-			Name:  "access-log",
-			Usage: "Access log file",
+			Name:  "fuse-trace-log",
+			Usage: "FUSE trace log file",
 		},
 		&cli.BoolFlag{
 			Name:  "as-root",
@@ -68,8 +68,9 @@ func mountFlags() []cli.Flag {
 			Value: min(runtime.NumCPU()*2, 16),
 		},
 		&cli.BoolFlag{
-			Name:  "case-sensitive",
-			Usage: "If set, the file system will be case sensitive",
+			Name:   "case-sensitive",
+			Usage:  "If set, the file system will be case sensitive",
+			Hidden: true,
 		},
 	}
 }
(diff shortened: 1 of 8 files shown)


cmd/mount_windows.go
```diff
@@ -64,6 +64,10 @@ func mountFlags() []cli.Flag {
 			Name:  "show-dot-files",
 			Usage: "If set, dot files will not be treated as hidden files",
 		},
+		&cli.BoolFlag{
+			Name:  "case-sensitive",
```
davies, line 68: It's known to be broken, let's hide it

pkg/winfsp/winfs.go
```diff
@@ -727,7 +727,29 @@ func (j *juice) Chflags(path string, flags uint32) (e int) {
...
+		}
+		ino = fi.Inode()
+	}
+
+	paths := j.vfs.Meta.GetPaths(ctx, ino)
+	if len(paths) == 0 {
+		ret = p
+		return
+	}
+	ret = path.Join(paths...)
```
davies, line 748: We should return one of them.

Multiple path is used for hardlink

cmd/mount_windows.go
```diff
@@ -91,8 +95,12 @@ func getDaemonStage() int {
 }
 
 func mountMain(v *vfs.VFS, c *cli.Context) {
-	v.Conf.AccessLog = c.String("access-log")
-	winfsp.Serve(v, c.String("o"), c.Float64("file-cache-to"), c.Bool("as-root"), c.Int("delay-close"), c.Bool("show-dot-files"))
+	acLog := c.String("access-log")
+	if acLog != "" {
+		winfsp.SetTraceOutput(acLog)
```
davies, line 100: the format of tracer is different than the access log in linux, should we use another name?

pkg/winfsp/winfs.go
chenjie4255: ok, I see. I thought they were the splited parts. added additional codes to choose the correct one.

cmd/mount_windows.go
chenjie4255: The gateway also uses the "access-log". but indeed it is better to rename it. I think the "fuse-trace-log" should be a 
 accurate one.

pkg/meta/redis.go
```diff
@@ -1760,7 +1762,11 @@ func (m *redisMeta) doRename(ctx Context, parentSrc Ino, nameSrc string, parentD
 		}
 		if dino > 0 {
 			if ino == dino {
-				return errno(nil)
+				if m.conf.CaseInsensi && nameSrc != nameDst {
```
davies, line 1765: please update the sql.go and tkv.go as well

chenjie4255: This has been updated, and the original logic has been simplified. 

The #5871 has been re-verfied on both SQL and redis engine.

PR #5920, opened 2025-04-03
windows: elevate some winfsp perf-related params

chenjie4255:
## tests

* Running juicefs bench with winfsp_threads = 16 vs 32, didn't see any significant improvement. This pr may be useful for some certain cases.

### DIFF
--- a/cmd/mount_windows.go
+++ b/cmd/mount_windows.go
@@ -19,6 +19,7 @@ package cmd
 import (
 	"os"
 	"path/filepath"
+	"runtime"
 
 	"github.com/juicedata/juicefs/pkg/meta"
 	"github.com/juicedata/juicefs/pkg/object"
@@ -64,6 +65,16 @@ func mountFlags() []cli.Flag {
 			Name:  "show-dot-files",
 			Usage: "If set, dot files will not be treated as hidden files",
 		},
+		&cli.IntFlag{
+			Name:  "winfsp-threads",
+			Usage: "WinFsp threads count option, Default is min(cpu core * 2, 16)",
+			Value: min(runtime.NumCPU()*2, 16),
+		},
+		&cli.Float64Flag{
+			Name:  "dirinfo-cache-to",
+			Usage: "Dir information timeout in seconds",
+			Value: 1,
+		},
 	}
 }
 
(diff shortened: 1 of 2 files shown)


cmd/mount_windows.go
```diff
@@ -64,6 +65,16 @@ func mountFlags() []cli.Flag {
...
 		},
+		&cli.IntFlag{
+			Name:  "winfsp-threads",
+			Usage: "WinFsp threads count option, Default is min(cpu core * 2, 16)",
+			Value: min(runtime.NumCPU()*2, 16),
+		},
+		&cli.Float64Flag{
+			Name:  "dirinfo-cache-to",
+			Usage: "Dir information timeout in seconds",
+			Value: 1,
```
davies, line 76: use String type to support '1s'

cmd/mount_windows.go
```diff
@@ -64,6 +65,16 @@ func mountFlags() []cli.Flag {
 			Name:  "show-dot-files",
 			Usage: "If set, dot files will not be treated as hidden files",
 		},
+		&cli.IntFlag{
+			Name:  "winfsp-threads",
```
davies, line 69: io-threads or threads?

cmd/mount_windows.go
chenjie4255: this was aligned to the existing option "file-cache-to". Should we change it as well?

cmd/mount_windows.go
chenjie4255: This is meant to expose the winfsp startup option, according to its document, 

```FspFileSystemStartDispatcher - Start the file system dispatcher.

FSP_API NTSTATUS FspFileSystemStartDispatcher(
    FSP_FILE_SYSTEM *FileSystem,
    ULONG ThreadCount);  
Parameters

FileSystem - The file system object.
ThreadCount - The number of threads for the file system dispatcher. A value of 0 will create a default number of threads and should be chosen in most cases.
Return Value

STATUS_SUCCESS or error code.

Discussion

The file system dispatcher is used to dispatch operations posted by the FSD to the user mode file system. Once this call starts executing the user mode file system will start receiving file system requests from the kernel.```

I think it is not limited to the I/O threads, but more like the number of worker, handling messages from the kernel driver.

File: cmd/mount_windows.go (hunk 1 above)
Comment (davies) [reply cmd/mount_windows.go:76]: yes, change them all

File: cmd/mount_windows.go [hunk 3]
```diff
@@ -64,6 +65,16 @@ func mountFlags() []cli.Flag {
...
+			Usage: "WinFsp threads count option, Default is min(cpu core * 2, 16)",
+			Value: min(runtime.NumCPU()*2, 16),
+		},
+		&cli.Float64Flag{
+			Name:  "dirinfo-cache-to",
```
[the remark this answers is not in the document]
chenjie4255: now it uses meta cache group:: dir-entry-cache directly.

cmd/mount_windows.go
chenjie4255: changed file-cache-toto "entry-cache" (from Meta Cache Gruop), and the default value changed to the default 1s as well.

PR #5922, opened 2025-04-04
chunk: correct path generation for expired and evicted disk cache file removal

### LINKED ISSUE #5921: cacheStore can not remove the expired and evicted cache file correctly
**What happened**:

When `pkg/chunk/disk_cache/cachedStore` cleanup expired cachefile, if a cacheFile is found expired, the `func (cache *cacheStore) cleanupExpire(): _ = cache.removeFile(cache.cachePath(cache.getPathFromKey(k)))` could not actually remove the cachefile successfully. 

Also, the cache file removal of eviceted cache could not be done correctly.

The cache file should be correctly removed from the file system.

Add the following test case to `pkg/chunk/disk_cache_test.go`:
``` go
func TestDiskCacheExpirationAndEviction(t *testing.T) {
	logger.SetLevel(logrus.DebugLevel)
	cacheDataDir := filepath.Join(t.TempDir(), "cache")
	conf := defaultConf
	conf.CacheExpire = 1 * time.Second
	conf.CacheDir = cacheDataDir
	conf.CacheSize = 11 << 20
	conf.AutoCreate = true
	cacheStore := newCacheManager(&conf, nil, nil)

	// Cache capcity could not satisfy 3 blocks, this would trigger eviction
	numBlocks := 3
	keys := make([]string, numBlocks)
	testData := make([]byte, 4<<20)
	for i := 0; i < numBlocks; i++ {
		keys[i] = fmt.Sprintf("%d_0_%d", i, len(testData))
		cacheStore.cache(keys[i], NewPage(testData), true, false)
	}

	// Wait for the cache to be flushed
	time.Sleep(2 * time.Second)

	numEvicted := 0
	for _, key := range keys {
		if _, err := os.Stat(filepath.Join(cacheDataDir, cacheDir, key)); os.IsNotExist(err) {
			numEvicted++
		}
	}
	assert.Equal(t, 1, numEvicted, "One cache entry should be evicted due to size limit")

	waitTime := conf.CacheExpire + 3*time.Second
	t.Logf("Waiting %v for cache entries to expire and removed...", waitTime)
	time.Sleep(waitTime)

	for _, key := range keys {
		_, err := os.Stat(filepath.Join(cacheDataDir, cacheDir, key))
		assert.True(t, os.IsNotExist(err), "Cache file should be removed after expiration or eviction")
	}
}


```

This bugs are caused by the incorrectness of the key cache file apth generation. 

``` go
func (cache *cacheStore) cleanupExpire() { 
		...
		for _, k := range todel {
			if !cache.available() {
				break
			}
			_ = cache.removeFile(cache.cachePath(cache.getPathFromKey(k)))
		...
}


func (cache *cacheStore) cleanupFull() {
        ...
		_ = cache.removeFile(cache.cachePath(getPathFromKey(k)))
	...
}
```

**Environment**:
- JuiceFS version (use `juicefs --version`) or Hadoop Java SDK version: juicefs version 1.3.0
- Kernel (e.g. `uname -a`): Darwin *** 23.4.0 Darwin Kernel Version 23.4.0

YinhaoHu:
close #5921

### DIFF
--- a/pkg/chunk/disk_cache.go
+++ b/pkg/chunk/disk_cache.go
@@ -367,7 +367,7 @@ func (cache *cacheStore) cleanupExpire() {
 			if !cache.available() {
 				break
 			}
-			_ = cache.removeFile(cache.cachePath(cache.getPathFromKey(k)))
+			_ = cache.removeFile(cache.cachePath(k.String()))
 		}
 		todel = todel[:0]
 		time.Sleep(interval * time.Duration((cnt+1-deleted)/(cnt+1)))
@@ -841,7 +841,7 @@ func (cache *cacheStore) cleanupFull() {
 		if !cache.available() {
 			break
 		}
-		_ = cache.removeFile(cache.cachePath(cache.getPathFromKey(k)))
+		_ = cache.removeFile(cache.cachePath(k.String()))
 	}
 	cache.Lock()
 }
--- a/pkg/chunk/disk_cache_test.go
+++ b/pkg/chunk/disk_cache_test.go
@@ -26,6 +26,7 @@ import (
 	"github.com/juicedata/juicefs/pkg/utils"
 	"github.com/prometheus/client_golang/prometheus"
 	dto "github.com/prometheus/client_model/go"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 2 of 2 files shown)


jiefenghuang:
hi @YinhaoHu , the cache directory also contains a chunk structure. you can check the cache and load calls. Therefore, a conversion is necessary.

YinhaoHu:
> hi @YinhaoHu , the cache directory also contains a chunk structure. you can check the cache and load calls. Therefore, a conversion is necessary.

hi @jiefenghuang , I added debug log entries in `cacheStore.flush` and `cacheStore.load`, there is no chunk directory in the cache file path. 


screenshot 2025-04-07 11 57 20

jiefenghuang:
Try reading the file directly and check the log output again, not from your unit test.

PR #5926, opened 2025-04-07
CI: fix dump/load test timeout.

zhoucheng361:


davies:
the test is too slow

zhoucheng361:
> the test is too slow

it is blocked when dump threads is bigger than max_open_conns set as parameter, I increase the max_open_conns and it works as expected. but we still need fix the problem if threads is bigger than max_open_conns

PR #5947, opened 2025-04-09
sync: custom cluster manager-addr port without ip

zxh326:
Currently, when we need to custom the manager-addr port, we must pass the manager's IP address; otherwise, an invalid address is passed to the worker

now, support custom manager port without ip like `--manager-addr=:1234`

### DIFF
--- a/pkg/sync/cluster.go
+++ b/pkg/sync/cluster.go
@@ -216,6 +216,13 @@ func startManager(config *Config, tasks <-chan object.Object) (string, error) {
 	if !strings.Contains(addr, ":") {
 		addr += ":"
 	}
+	if strings.HasPrefix(addr, ":") || strings.Contains(addr, "0.0.0.0") {
+		ip, err := utils.GetLocalIp(net.JoinHostPort(config.Workers[0], "22"))
+		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/sync/cluster.go
```diff
@@ -212,6 +212,13 @@ func startManager(config *Config, tasks <-chan object.Object) (string, error) {
 	if !strings.Contains(addr, ":") {
 		addr += ":"
 	}
+	if strings.HasPrefix(addr, ":") || strings.Contains("0.0.0.0", addr) {
```
zhijian-pro, line 215: `strings.Contains("0.0.0.0", addr)` ？

pkg/sync/cluster.go
```diff
@@ -212,6 +212,13 @@ func startManager(config *Config, tasks <-chan object.Object) (string, error) {
 	if !strings.Contains(addr, ":") {
 		addr += ":"
 	}
+	if strings.HasPrefix(addr, ":") || strings.Contains(addr, "0.0.0.0") {
+		ip, err := utils.GetLocalIp(net.JoinHostPort(config.Workers[0], "22"))
```
davies, line 216: config.Workers[0] could be  `user@host`

Issue #5900, opened 2025-04-01
should update-fstab use absolute path for mount path?


anysql:
When run "./juicefs mount ... --update-fstab", the mount path is not replaced with an absolute path in fstab.

just found that other mount paths are all absolute paths.

```
root@bench-03:/data/benchmark# cat /etc/fstab
# /etc/fstab: static file system information.
#
.......
LABEL=/data /data ext4 defaults,nodelalloc,noatime 0 0
sqlite3:///data/benchmark/localmeta.db  jfs  juicefs  _netdev,cache-dir=/data/benchmark/cachedir1:/data/benchmark/cachedir2:/data/benchmark/cachedir3,cache-expire=900  0 0
```

Should we replace it?

PR #5951, opened 2025-04-09
cmd/mount: use absolute path for mount path while --update-fstab

Xuyuchao-juice:
fix https://github.com/juicedata/juicefs/issues/5900

### DIFF
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -27,7 +27,6 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
-	"sort"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


cmd/mount.go
```diff
@@ -546,7 +546,10 @@ func getDefaultLogDir() string {
 
 func updateFstab(c *cli.Context) error {
 	addr := expandPathForEmbedded(c.Args().Get(0))
-	mp := c.Args().Get(1)
+	mp, err := filepath.Abs(c.Args().Get(1))
```
davies, line 549: use absPath

cmd/mount.go
```diff
@@ -544,9 +544,27 @@ func getDefaultLogDir() string {
...
+		}
+	}
+	d, err := filepath.Abs(d)
+	if err != nil {
+		logger.Fatalf("Expand %s: %s", d, err)
+	}
+	return d
+}
+
 func updateFstab(c *cli.Context) error {
```
davies, line 565: we should move updateFstab into mount_unix

PR #5957, opened 2025-04-09
cmd/sync: fix large buffer prevents sync throughput from scaling with the number of workers

polyrabbit:
Since version 1.2, we have observed that even when specifying multiple workers using the `--worker` option, the throughput of sync tool does not scale any more. Debugging revealed that because each worker's buffer is very large, a single worker might fetch (almost) all tasks into its own buffer, causing other workers to remain idle:

img_v3_02l6_4defac40-41f0-4aea-9d6a-1b9f13fb886g

After removing the buffer, workers can fetch as demand now:

img_v3_02l6_9994ff9f-a7a5-45af-bd96-1712ca24d8cg

### DIFF
--- a/pkg/sync/sync.go
+++ b/pkg/sync/sync.go
@@ -1577,7 +1577,8 @@ func Sync(src, dst object.ObjectStorage, config *Config) error {
 
 	var bufferSize = 10240
 	if config.Manager != "" {
-		bufferSize = 100
+		// No support for work-stealing, so workers shouldnot buffer tasks to prevent piling up in their own queues, which could cause imbalance among workers.
+		bufferSize = 0
 	}
 	tasks := make(chan object.Object, bufferSize)
 	wg := sync.WaitGroup{}


polyrabbit:
I hope this change can be backported to 1.2 because sync throughput in 1.1 scales linearly with the number of workers, but this functionality is lost in 1.2.

davies:
The buffer size is the same in 1.0-1.2, so it's not a regress, may not back port it into 1.2.

polyrabbit:
Then, maybe it's caused by other changes? We tested 1.1 and the EE version - the throughput of `sync` scales well.

davies:
Can you reproduce the difference using same dataset everytime?

polyrabbit:
Yes, our test is based on the same dataset - it's a checkpoint ~660GiB. We get the latest binary using `curl -sSL https://d.juicefs.com/install | sh -` and it only provides 800MiB/s throughput with 3 workers, requiring more than 10 minutes to sync one model. After switching to the EE version,  the speed stabilized to 2.4GiB/s. We also tested the latest 1.1 - it has the same speed.

Issue #5252, opened 2024-10-23
jfs mounted on Windows way slower than jfs mounted on Linux exposed via smb


happyc0ding:
**What happened**:
I have 1 client (Windows 10), 1 server (Linux).
(Background: Mounting an image via Arsenal Image Mounter)
Scenario 1:
Mounting jfs in Windows via network and opening a big file (~150GB): Takes about 60 seconds.
Scenario 2:
Mounting the same jfs on the Linux server directly and exposing the mount via smb, then mounting this share via the same network in Windows and opening the same file: Takes about 12 seconds.

Playing around with prefetch, cache-partial-only and setting cache-dir to memory did not change much (caching might have been an explanation, since the local disk is slower than the disk minio is running on).
I've tried several times in order to confirm the timing and ruling out any other caching mechanisms.

**What you expected to happen**:
I expected the performance in scenario 1 and 2 to be roughly the same or scenario 2 to be slower since it adds another protocol layer.

**How to reproduce it (as minimally and precisely as possible)**:
Install jfs on Windows, install jfs, samba, minio and mariadb on linux. Create a jfs and mount it.
Mounted on linux as described in the docs: `sudo juicefs mount -d --enable-xattr mysql://... /mnt/myjfs` and used a  default smb.conf provided by the package, only added a share as described in [the docs](https://juicefs.com/docs/community/deployment/samba/#step-3-create-a-samba-share).
Mounted jfs in windows with `juicefs.exe mount "mysql://..." x:`
Mounted smb in windows with `net use x: \\foo\bar`.

**Anything else we need to know?**
Not sure if this is a bug. However, I did not find anything in the docs explaining this. Is the windows driver that much slower than linux? If yes, it would be nice to know.

**Environment**:
- JuiceFS version (use `juicefs --version`) or Hadoop Java SDK version: juicefs version 1.1.4+2024-08-30.fc080fd on windows, juicefs version 1.2.1+2024-08-30.cd871d1 on linux
- OS (e.g `cat /etc/os-release`): Windows 10 / Debian 12
- Kernel (e.g. `uname -a`): Linux 6.1.0-26-amd64
- Object storage (cloud provider and region, or self maintained): self hosted minio version RELEASE.2024-10-02T17-50-41Z
- Metadata engine info (version, cloud provider managed or self maintained): mariadb from 11.5.2-MariaDB
- Network connectivity (JuiceFS to metadata engine, JuiceFS to object storage): 10G via Vmware
[...]

PR #5961, opened 2025-04-11
windows: added "enable-get-path" mount option

chenjie4255:
From the tests that I ran, I found that when disabling the get-get callback, the small file reading performance in the bench test could earn an improvement of about 30% on my machine. 

Since we are running a case-insensitive file system on Windows, I think it is better to disable this feature by default to get a better default performance.

### DIFF
--- a/cmd/mount_windows.go
+++ b/cmd/mount_windows.go
@@ -73,8 +73,8 @@ func mountFlags() []cli.Flag {
 			Hidden: true,
 		},
 		&cli.BoolFlag{
-			Name:  "enable-get-path",
-			Usage: "If set, it will improve the accuracy of letter case when returning file paths. (May incur a performance lost)",
+			Name:  "report-case",
+			Usage: "If set, juicefs will report the correct case of a file path for a case-insensitive filesystem. (May incur a performance lost)",
 		},
 	}
 }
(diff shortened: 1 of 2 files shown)


chenjie4255:
BTW, the tests I ran in https://github.com/juicedata/juicefs/issues/5252 were run without the "enable-get-path" option, If this option was enabled, the small file reading performance would decrease to around 100 files/s.

with_get_path

cmd/mount_windows.go
```diff
@@ -72,6 +72,10 @@ func mountFlags() []cli.Flag {
 			Usage:  "If set, the file system will be case sensitive",
 			Hidden: true,
 		},
+		&cli.BoolFlag{
+			Name:  "enable-get-path",
```
davies, line 76: Can we have a friendly name for user?  something related to case sensitivity

chenjie4255: The reason I chose a somewhat vague name is that the purpose of this interface is hard to describe with a short, clear word. Also, in some tests, it doesn’t always return the correct case-sensitive file names. So I thought it might be better to use a vague name and explain its function more clearly in the description. 

Maybe we could make the name more specific, for example, "--enable-winfsp-get-path" or "winfsp-enable-get-path"?

Winfsp changelog described this interface as follows:

 ```The `getpath` operation allows a case-insensitive file system to report the correct case of a file path. For example, `getpath` can be used to report that the actual path of a file opened as `/PATH/TO/FILE` is really `/Path/To/File`. This capability is important for some Windows file system scenarios and can sometimes result in a performance improvement.```

davies: `strict-case` or `report-case`?

PR #5970, opened 2025-04-12
feat(chunk): trigger eviction by freeratio in cache add path

myron369:
The size limit of diskcache is controlled by the direct parameter cacheSize as well as the disk's freeRatio. Therefore, in cacheStore.add function, eviction tirgger could also take freeRatio's size limit into consideration.


davies:
We check freeRatio in every second, it's OK that it's not trigger in `add`, because there are other ways to fill disk also.

This PR assume the capacity of disk will not change, which is not true some times.

myron369:
> We check freeRatio in every second, it's OK that it's not trigger in `add`, because there are other ways to fill disk also.
> 
> This PR assume the capacity of disk will not change, which is not true some times.

@davies Hi, Thanks for your reply. Here is my view: The add path could measure size and inode limit by freeratio in approximate calculations, so it could aggressively trigger eviction instead of relying on the cleanupFull goroutine's loop.
In this way cleanupFull goroutine becomes a backup mechanism to limit disk usage. But instead of updating cacheSize and maxItem, maybe it's better to use new parameters(like cacheSizeFreeRatio).

PR #5977, opened 2025-04-14
Fix higher concurrency leads to lower throughput

polyrabbit:
When buffer required by the user's write concurrency exceeds the `--buffer-size` option set by administrator, JuiceFS exhibits a behavior where higher concurrency results in lower throughput, which is counter-intuitive to the user. The reason is that when buffer is full, JuiceFS triggers a write stall. User requests cannot fill a block timely, and the buffer in memory can only be flushed by the condition `now.Sub(s.lastMod) > time.Second`, meaning each slice is flushed once per second. This not only slows down throughput but also results in many fragmented blocks to OSS. Here is an example: with a 100MiB buffer, we have less throughput and smaller blocks flushed to OSS when the number of concurrent writes increases from 1 to 60:

img_v3_02l5_c442d0a8-340a-4b4c-9fc3-d472f864b11g

So when buffer is full, we should actively flush buffer to OSS, instead of waiting passively. This way, JuiceFS will have a stable throughput even when user's write concurrency increases.


pkg/vfs/writer.go
```diff
@@ -453,29 +453,56 @@ func (w *dataWriter) flushAll() {
 	for {
 		w.Lock()
 		now := time.Now()
+		bufferExceeds := w.usedBufferSize() - w.bufferSize
```
jiefenghuang, line 456: Perhaps we can trigger it earlier, say at 95% usage?

pkg/vfs/writer.go
```diff
@@ -453,29 +453,56 @@ func (w *dataWriter) flushAll() {
...
+					case now.Sub(s.started) > flushDuration: // too old
+						flushDurTriggered += 1
+					case now.Sub(s.lastMod) > time.Second && now.Sub(s.started) > time.Second: // too stale
+						lastModTriggered += 1
+					case slicesExceeds > 0: // too many slices
+						slicesExceeds -= 1
+						sliceTriggered += 1
+					case bufferExceeds > 0: // too many buffer
+						myPending := s.writer.PendingBytes()
+						if float64(myPending)/float64(w.conf.Chunk.BlockSize) <= rand.Float64() {
```
jiefenghuang, line 481: Could you add a mock unit test to verify this effect? 

Also, how did the actual test results turn out after the modification?

pkg/vfs/writer.go
polyrabbit: Earlier freeze causes more fragmentation(small blocks), which I suppose is even worse.

polyrabbit:
Close this patch for now until there is a way to address this issue without causing fragmentation.

PR #5998, opened 2025-04-16
sdk/python: introduce a ffrecord dataloader example with pythonsdk

CodingPoeta:


sdk/python/examples/ffrecord/dataset.py
```diff
@@ -0,0 +1,50 @@
+import numpy as np
```
davies, line 1: license header

sdk/python/examples/ffrecord/dataloader.py
```diff
@@ -0,0 +1,94 @@
+# encoding: utf-8
+# JuiceFS, Copyright 2020 Juicedata, Inc.
```
davies, line 2: 2025

Issue #5989, opened 2025-04-16
After the mount path is changed, the fstab configuration is added instead of modified


YunhuiChen:
**What happened**:
```

./juicefs mount mysql://root:****@(172.28.39.187)/test /juicefs2 --cache-dir=/data/cache-mysql --log=/var/log/juicefs_mysql.log -d --update-fstab

./juicefs mount mysql://root:****@(172.28.39.187)/test /juicefs4 --cache-dir=/data/cache-mysql --log=/var/log/juicefs_mysql.log -d --update-fstab

mysql://root:root123@(172.28.39.187)/test  /juicefs2  juicefs  _netdev,cache-dir=/data/cache-mysql,log=/var/log/juicefs_mysql.log  0 0
mysql://root:root123@(172.28.39.187)/test  /juicefs4  juicefs  _netdev,cache-dir=/data/cache-mysql,log=/var/log/juicefs_mysql.log  0 0
```

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**

**Environment**:
- JuiceFS version (use `juicefs --version`) or Hadoop Java SDK version:
- root@chaos-k8s-001:~/chenyunhui/juicefs# ./juicefs version
juicefs version 1.3.0-dev+2025-04-15.66b50c74
- Cloud provider or hardware configuration running JuiceFS:
- OS (e.g `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Object storage (cloud provider and region, or self maintained):
- Metadata engine info (version, cloud provider managed or self maintained):
- Network connectivity (JuiceFS to metadata engine, JuiceFS to object storage):
- Others:

davies:
We don't know the actual abs path of a relative path in /etc/fstab, so there is no way to fix that.

PR #6002, opened 2025-04-17
cmd/mount: --update-fstab. The same file system needs to overwrite th…

Xuyuchao-juice:
fix https://github.com/juicedata/juicefs/issues/5989


davies:
We allow mount the same file system to different location, so it's OK to have multiple entry for same file system.

PR #6033, opened 2025-04-22
Exit compact timely and log a message when long-running background tasks start

polyrabbit:
Two minor fixes:
- When user cancels the `cmd/compact`, the compact task in the background just won't stop because of a deep queue(10000). It should exit timely as other cmd tasks.
- It's better to have a log message for background tasks when they start, so we can know what's happening when ceertain resources spike (cpu/memory/meta ops).

### DIFF
--- a/pkg/vfs/fill.go
+++ b/pkg/vfs/fill.go
@@ -26,7 +26,6 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/dustin/go-humanize"
 	"github.com/juicedata/juicefs/pkg/chunk"
 
 	"github.com/juicedata/juicefs/pkg/meta"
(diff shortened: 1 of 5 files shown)


pkg/vfs/fill.go
```diff
@@ -173,7 +172,7 @@ func (c *CacheFiller) Cache(ctx meta.Context, action CacheAction, paths []string
 	if ctx.Canceled() {
 		logger.Infof("%s cancelled", action)
 	}
-	logger.Infof("%s %d paths in %s", action, len(paths), time.Since(start))
+	logger.Infof("%s %s (%s) in %s", action, strings.Join(paths, ","), humanize.IBytes(resp.TotalBytes), time.Since(start))
```
SandyXSD, line 175: The `paths` could be a very large list.

polyrabbit: OK, reverted.

jiefenghuang (commented):
LGTM

PR #6037, opened 2025-04-22
cmd/info: add file offset column

polyrabbit:
Currently the `offset` field in cmd/info output is hard coded to `0` - it doesn't provide any useful information. It's better to display absolute offsets in the file, so when some chunk is corrupted, we can easily identify which part of the file is affected.

image


davies:
The offset is not 0 when a file is written using copy_file_range.

polyrabbit:
Got it, so I updated the commit message.

davies:
We can call it `pos`, and put it at the beginning or end.

polyrabbit:
OK, how about putting it to the end if not everyone will be using it?
image

Issue #6009, opened 2025-04-18
It is suggested that pysdk can obtain the.config file


YunhuiChen:
**What would you like to be added**:

It is suggested that pysdk can obtain the.config file

```
>>> with v.open('/.config','rb') as f:
...   f.read()
...
b''
```

**Why is this needed**:
It is not certain whether the configuration takes effect correctly

PR #6054, opened 2025-04-23
sdk/python: support Client.get_config()

CodingPoeta:


CodingPoeta:
A simple workaround for #6009

PR #6068, opened 2025-04-25
docs: fix missing fdb docs

anysql:
fix the index only, translating is on the way


anysql:
Else the chinese version fdb doc is not displayed.

PR #6073, opened 2025-04-25
chore: Meta module will check the inode name

fukua95:
The inode name cannot contain the characters '/' and '\0'.
Meta module will check the inode name when `Mknod()`, `Link()`, `Rename()` are called.

### DIFF
--- a/pkg/meta/random_test.go
+++ b/pkg/meta/random_test.go
@@ -203,15 +203,15 @@ func (m *fsMachine) create(_type uint8, parent Ino, name string, mode, umask uin
 	if _type < TypeFile || _type == TypeSymlink {
 		return syscall.EINVAL
 	}
+	if checkFSNodeName(name) != 0 {
+		return syscall.EINVAL
+	}
 	p := m.nodes[parent]
 	if p == nil {
 		return syscall.ENOENT
 	}
 	if p.children == nil {
 		return syscall.ENOTDIR
-	}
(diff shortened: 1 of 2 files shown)


pkg/meta/random_test.go
```diff
@@ -291,24 +297,23 @@ func (m *fsMachine) create(_type uint8, parent Ino, name string, mode, umask uin
 	return 0
 }
 
-func fsnodes_namecheck(name string) syscall.Errno {
-	return 0
-	nleng := len(name)
-	if nleng == 0 {
+func checkFSNodeName(name string) syscall.Errno {
```
davies, line 300: To make this useful, stringN should generate invalid names.

fukua95: I misunderstood. Thank you. @davies

pkg/meta/random_test.go
```diff
@@ -471,7 +468,7 @@ func (m *fsMachine) unlink(parent Ino, name string) syscall.Errno {
 	if _, ok := p.children[name]; !ok {
 		return syscall.ENOENT
 	}
-	if fsnodes_namecheck(name) != 0 {
+	if checkFSNodeName(name) != 0 {
```
davies, line 471: checkFSNodeName should return bool, otherwise we should use the returned errno

fukua95: Thank you. I have already fixed it.

PR #6093, opened 2025-05-08
cmd/sync: add check-change flag

zxh326:


pkg/sync/sync.go
```diff
@@ -913,6 +921,29 @@ func worker(tasks <-chan object.Object, src, dst object.ObjectStorage, config *C
...
+		checkedBytes.IncrInt64(obj.Size())
+
+		if currentObj.Size() == obj.Size() && currentObj.Mtime().Equal(obj.Mtime()) {
+			return true, nil
+		} else {
+			logger.Warnf("Source file %s changed during sync. Original: size=%d, mtime=%s; Current: size=%d, mtime=%s",
+				currentObj.Key(), obj.Size(), obj.Mtime(), currentObj.Size(), currentObj.Mtime())
+			return false, nil
+		}
+	} else {
```
davies, line 942: we can check the size of dst here

pkg/sync/config.go
```diff
@@ -220,6 +222,10 @@ func NewConfigFromCli(c *cli.Context) *Config {
 		logger.Warnf("threads should be larger than 0, reset it to 1")
 		cfg.Threads = 1
 	}
+	if (cfg.CheckAll || cfg.CheckNew) && cfg.CheckChange {
```
davies, line 225: we can do the both, check the change first, only check data later (don't update the metrics twice)

pkg/sync/sync.go
```diff
@@ -913,6 +921,38 @@ func worker(tasks <-chan object.Object, src, dst object.ObjectStorage, config *C
...
+			if currentObj.Size() != dstObj.Size() {
+				logger.Warnf("Destination object %s size mismatch: original=%d, current=%d", key, obj.Size(), dstObj.Size())
+				return false, nil
+			}
+		} else {
+			return false, headErr
+		}
+		return true, nil
+	} else {
+		return false, headErr
```
davies, line 952: handle  removed object as changed, not failure

pkg/sync/sync.go
```diff
@@ -913,6 +921,38 @@ func worker(tasks <-chan object.Object, src, dst object.ObjectStorage, config *C
...
+		}
+		equal := currentObj.Size() == obj.Size() && currentObj.Mtime().Equal(obj.Mtime())
+		if !equal {
+			logger.Warnf("Source file %s changed during sync. Original: size=%d, mtime=%s; Current: size=%d, mtime=%s",
+				currentObj.Key(), obj.Size(), obj.Mtime(), currentObj.Size(), currentObj.Mtime())
+			return false, nil
+		}
+		if dstObj, headErr := dst.Head(key); headErr == nil {
+			if currentObj.Size() != dstObj.Size() {
+				logger.Warnf("Destination object %s size mismatch: original=%d, current=%d", key, obj.Size(), dstObj.Size())
```
davies, line 944: return different error

Issue #6096, opened 2025-05-09
Support subdir option in hadoop sdk


kyungwan-nam:
**What would you like to be added**:

Support `subdir` option in hadoop-sdk. It is same as `subdir` option in [cmd/mount](https://juicefs.com/docs/community/command_reference/#mount-metadata-options)

**Why is this needed**:

In our use case, users share the juicefs volume between Hadoop and k8s.
It would be good if 'subdir' can be set also in Hadoop-sdk.
It provides isolation for each user or each task.

PR #6097, opened 2025-05-09
Support subdir option in hadoop sdk

kyungwan-nam:
PR for https://github.com/juicedata/juicefs/issues/6096

### DIFF
--- a/sdk/java/libjfs/main.go
+++ b/sdk/java/libjfs/main.go
@@ -356,7 +356,7 @@ type javaConf struct {
 	PushLabels        string `json:"pushLabels"`
 	PushGraphite      string `json:"pushGraphite"`
 	Caller            int    `json:"caller"`
-	AllowedSubdir     string `json:"allowedSubdir"`
+	Subdir            string `json:"subdir"`
 
 	SuperFS bool `json:"superFs,omitempty"`
 }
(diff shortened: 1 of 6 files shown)


sdk/java/libjfs/main.go
```diff
@@ -396,6 +397,25 @@ func getOrCreate(name, user, group, superuser, supergroup string, superFs bool,
...
+			if err != 0 {
+				logger.Errorf("Subdir %s must be valid if read-only. but failed to stat %v", conf.SubDir, err)
+				return 0
+			}
+			if !fi.IsDir() {
+				logger.Errorf("Subdir %s must be valid if read-only. but, it is not a directory", conf.SubDir)
+				return 0
+			}
+		}
+		if err := jfs.Meta().Chroot(w.ctx, conf.SubDir); err != 0 {
```
davies, line 413: We can limit the access of Java SDK under a subdir, but the SDK should always use the full path.

When the path is not within subdir, it should throw AccessDeniedException

kyungwan-nam: I rewritten it with reference to your comment.
please let me know that if you have a better way.
Thank you

sdk/java/src/main/java/io/juicefs/JuiceFileSystemImpl.java
```diff
@@ -924,6 +957,7 @@ private void setStorageId(BlockLocation bl) {
 
   @Override
   public BlockLocation[] getFileBlockLocations(FileStatus file, long start, long len) throws IOException {
+    checkSubPath(file.getPath());
```
davies, line 960: we move this into FileSystem.doResolve in pkg/fs.go

sdk/java/src/main/java/io/juicefs/JuiceFileSystemImpl.java
```diff
@@ -405,6 +405,16 @@ public void initialize(URI uri, Configuration conf) throws IOException {
     for (String key : bkeys) {
       obj.put(key, Boolean.valueOf(getConf(conf, key, "false")));
     }
+    String allowedSubdir = getConf(conf, "allowed-subdir", "");
```
davies, line 408: we can just call it `subdir`

davies: please also update docs

kyungwan-nam: thanks for your review. I'll update it shortly. :)

kyungwan-nam:
Thanks for review and merge

PR #6108, opened 2025-05-14
Exit StatFS routine when closing session

polyrabbit:
A complement to #5486

### DIFF
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -353,6 +353,9 @@ func (m *baseMeta) InitMetrics(reg prometheus.Registerer) {
 
 	go func() {
 		for {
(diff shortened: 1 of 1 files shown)


pkg/meta/base.go
```diff
@@ -353,6 +353,9 @@ func (m *baseMeta) InitMetrics(reg prometheus.Registerer) {
 
 	go func() {
 		for {
+			if m.sessCtx == nil || m.sessCtx.Canceled() {
```
jiefenghuang, line 356: m.sessCtx != nil && m.sessCtx.Canceled() ?

For java sdk, InitMetrics is called before NewSession.

polyrabbit: Got it, updated

Issue #5851, opened 2025-03-25
flaky test: sync_fsrand


zhijian-pro:
https://github.com/juicedata/juicefs/actions/runs/14036927706/job/39297084362

![Image](https://github.com/user-attachments/assets/ab7bb6d6-53e0-4a72-ba70-f7f456fc9cf2)

YunhuiChen:
Reproduced locally:

```
025/04/21 15:40:36.013068 juicefs[1125700] <INFO>: Ping redis latency: 55.222µs [checkServerConfig@redis.go:3656]
2025/04/21 15:40:36.017838 juicefs[1125700] <INFO>: Prometheus metrics listening on 127.0.0.1:45013 [exposeMetrics@mount.go:134]
2025/04/21 15:40:36.018576 juicefs[1125700] <INFO>: Syncing from file:///tmp/fsrand1/ to jfs://juicefs/jfs/fsrand1/ [Sync@sync.go:1699]
 Scanned objects: 365/365 [===]  3174.5/s used: 114.99058ms
Excluded objects: 0                          0.0/s
  Excluded bytes: 0.0 b     (0 Bytes)        0.0 b/s
 Skipped objects: 135                        1178.3/s
   Skipped bytes: 363.0 b   (363 Bytes)      3.1 KiB/s
   Extra objects: 0                          0.0/s
     Extra bytes: 0.0 b     (0 Bytes)        0.0 b/s
 Pending objects: 0                          0.0/s
  Copied objects: 229                        1999.6/s
    Copied bytes: 182.0 b   (182 Bytes)      1.6 KiB/s
 Checked objects: 358                        3127.3/s
   Checked bytes: 776.0 KiB (794589 Bytes)   6.6 MiB/s
 Deleted objects: 1                          8.7/s
  Failed objects: 0                          0.0/s
2025/04/21 15:40:36.133189 juicefs[1125700] <INFO>: Found: 365, excluded: 0 (0 B), skipped: 135 (363 B), copied: 229 (182 B), extra: 0 (0 B), checked: 358 (775.97 KiB), deleted: 1, failed: 0 [Sync@sync.go:1634]
2025/04/21 15:40:36.133540 juicefs[1125700] <INFO>: flush session 0: [FlushSession@base.go:706]
2025/04/21 15:40:36.134269 juicefs[1125700] <INFO>: close session 0: <nil> [CloseSession@base.go:695]
[...]

PR #6117, opened 2025-05-19
cmd/sync: Fix the multi-threaded out-of-order bug when --delete-dst is set

Xuyuchao-juice:
fix https://github.com/juicedata/juicefs/issues/5851

Description:
When sync carries -- deletre-dst and --list-threads>1 and overwrites the directory with the same name in jfs with one file (or overwrites the file with the same name in jfs with one directory), multithreading cannot ensure the atomicity of the delete operation and the create operation.

### DIFF
--- a/pkg/sync/sync.go
+++ b/pkg/sync/sync.go
@@ -1115,6 +1115,14 @@ func produce(tasks chan<- object.Object, srckeys, dstkeys <-chan object.Object, 
 				dstobj = nil
 				continue
 			}
+
+			if config.DeleteDst && (dstobj.IsDir() != obj.IsDir()) {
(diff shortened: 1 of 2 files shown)


pkg/sync/sync.go
```diff
@@ -1115,6 +1115,14 @@ func produce(tasks chan<- object.Object, srckeys, dstkeys <-chan object.Object,
 				dstobj = nil
 				continue
 			}
+
+			if config.DeleteDst && (dstobj.IsDir() != obj.IsDir()) {
+				tasks <- withSize(dstobj, markDeleteDst)
```
davies, line 1120: These two will be executed out of order.

PR #6132, opened 2025-05-23
fix: when encountering a non-empty error while deleting a directory, do not retry multiple times

zhijian-pro:


pkg/sync/sync.go
```diff
@@ -220,10 +221,27 @@ var bufPool = sync.Pool{
...
 
+func shouldRetry(err error) bool {
+	if err == nil || errors.Is(err, utils.ErrSkipped) {
+		return false
+	}
+
+	var eno syscall.Errno
+	if errors.As(err, &eno) {
+		switch eno {
+		case syscall.EAGAIN, syscall.EINTR, syscall.EBUSY, syscall.ETIMEDOUT, syscall.EIO:
```
davies, line 232: add external symlink here

Issue #6149, opened 2025-05-27
Display full address of s3 compatible object storage in log


yujunz:
**What would you like to be added**:
Display full address of s3 compatible object storage in log
**Why is this needed**:
Current display is confusing especially when source and destination bucket have the same name
```
Syncing between s3://juicefs/us-west/ and s3://juicefs/us-west/
```
Desired display
```
Syncing between s3://juicefs.source.minio.me/us-west/ and s3://juicefs.dest.minio.me/us-west/
```

jiefenghuang:
closed by #6151

PR #6151, opened 2025-05-27
Display full address of s3 compatible object storage in log

polyrabbit:
Looks like this is a common requirement as requested in #6149 - we have been using this output format internally for a long time.


pkg/object/s3.go
```diff
@@ -55,6 +55,9 @@ type s3client struct {
 }
 
 func (s *s3client) String() string {
+	if s.s3.Options().BaseEndpoint != nil {
+		return fmt.Sprintf("%s/%s/", *s.s3.Options().BaseEndpoint, s.bucket)
```
davies, line 59: Can we use `s3://endpoint/bucket`?

polyrabbit: `BaseEndpoint` already has the `s3://` prefix or the default `http://` prefix - there is no need to add another one.

polyrabbit: I also added a UT to demonstrate the output.

polyrabbit: `BaseEndpoint` always has a schema, even if user does not specify one, the constructor will prepend a default schema:
https://github.com/juicedata/juicefs/blob/f08f36a308ea53e9bd91781209172a58d50a2401/pkg/object/s3.go#L441-L447

Or do you prefer this string to always start with `s3://`? But in that case, it would override the bucket parameter specified by user.

davies: yes, we want always use s3:// to keep it consistent.

polyrabbit: Updated

PR #6164, opened 2025-05-30
docs: Add FAQ about delete file

Xuyuchao-juice:


jiefenghuang:
Including a flowchart of the state transitions would make it clearer and more straightforward.

PR #6165, opened 2025-05-30
fix: adjust sleep duration calculation in cache expiration loop

zhijian-pro:


pkg/chunk/disk_cache.go
```diff
@@ -388,7 +388,7 @@ func (cache *cacheStore) cleanupExpire() {
 			_ = cache.removeFile(cache.cachePath(cache.getPathFromKey(k)))
 		}
 		todel = todel[:0]
-		time.Sleep(interval * time.Duration((cnt+1-deleted)/(cnt+1)))
+		time.Sleep(interval / 1000 * time.Duration((cnt+1-deleted)*1000/(cnt+1)))
```
solracsf, line 391: Could be

```
time.Sleep(interval * time.Duration(cnt+1-deleted) / time.Duration(cnt+1))
```

This avoids integer division entirely by casting to `time.Duration` before dividing.

zhijian-pro: There are similar processing methods in our project code, so I continued to use this style.

jiefenghuang (approved):

PR #6168, opened 2025-06-03
meta: add a option `--fast-statfs` to speed up statfs

lidaohang:
#### Problem
In large-scale deployments with thousands of clients, the statRootFs method triggers high-concurrency access to two metadata counters:

```go
m.en.getCounter(usedSpace)
m.en.getCounter(totalInodes)
```

These counters are implemented as hot keys in TiKV. When each client frequently queries them, it results in:
- TiKV node QPS spiking to 50k–100k;
- TiKV nodes restarting due to overload;
- PD nodes being overwhelmed and becoming unstable.

This severely impacts the availability and performance of the metadata service.

#### Solution
This PR removes direct calls to getCounter(usedSpace) and getCounter(totalInodes). Instead, the values are now read only from in-memory atomic variables:

```go
used = atomic.LoadInt64(&m.usedSpace)
inodes = atomic.LoadInt64(&m.usedInodes)
```
These variables are already maintained in memory and provide sufficient accuracy for filesystem statistics.

####  Benefits
- Eliminates hot key access in TiKV.
- Greatly reduces metadata server and TiKV load under heavy client concurrency.
- Prevents PD and TiKV from being overwhelmed.
- Improves stability and scalability of statfs under load.

### DIFF
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -813,21 +813,9 @@ func (m *baseMeta) StatFS(ctx Context, ino Ino, totalspace, availspace, iused, i
 
 func (m *baseMeta) statRootFs(ctx Context, totalspace, availspace, iused, iavail *uint64) syscall.Errno {
 	var used, inodes int64
-	var err error
-	err = utils.WithTimeout(func() error {
-		used, err = m.en.getCounter(usedSpace)
-		return err
-	}, time.Millisecond*150)
-	if err != nil {
-		used = atomic.LoadInt64(&m.usedSpace)
-	}
-	err = utils.WithTimeout(func() error {
-		inodes, err = m.en.getCounter(totalInodes)
-		return err
-	}, time.Millisecond*150)
-	if err != nil {
-		inodes = atomic.LoadInt64(&m.usedInodes)
-	}
+	used = atomic.LoadInt64(&m.usedSpace)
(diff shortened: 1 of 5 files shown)


pkg/meta/base.go
```diff
@@ -813,21 +813,9 @@ func (m *baseMeta) StatFS(ctx Context, ino Ino, totalspace, availspace, iused, i
...
-		used = atomic.LoadInt64(&m.usedSpace)
-	}
-	err = utils.WithTimeout(func() error {
-		inodes, err = m.en.getCounter(totalInodes)
-		return err
-	}, time.Millisecond*150)
-	if err != nil {
-		inodes = atomic.LoadInt64(&m.usedInodes)
-	}
+	used = atomic.LoadInt64(&m.usedSpace)
```
davies: This reduces the accuracy of used space and inodes, it's better to make it optional by adding an option for this.

lidaohang: OK

lidaohang: done

davies (approved):

polyrabbit:
When `--fast-statfs` is enabled, it is better to proactively fetch `m.usedSpace` from meta engine at startup; otherwise, before the first heartbeat cycle, user may get an empty usage value, which could cause confusion.

PR #6178, opened 2025-06-10
Avoid calling `GetAttr("/.config")` every 5 seconds in `watchdog`

polyrabbit:
Follow up of #6168, reduce workload on metadata engine caused by too many clients.

The result of `Stat("/.config")` does not affect liveness check of watchdog.

### DIFF
--- a/cmd/mount_unix.go
+++ b/cmd/mount_unix.go
@@ -140,16 +140,18 @@ func watchdog(ctx context.Context, mp string) {
 	var dev uint64
 	go func() {
 		time.Sleep(time.Millisecond * 100) // wait for child process
+		var confStat syscall.Stat_t
+		var confName = ".config"
+		if !vfs.IsSpecialName(confName) {
+			confName = ".jfs" + confName
+		}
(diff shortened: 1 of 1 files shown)


cmd/mount_unix.go
```diff
@@ -140,16 +140,18 @@ func watchdog(ctx context.Context, mp string) {
...
+			confName = ".jfs" + confName
+		}
+		expectedIno, _ := vfs.GetInternalNodeByName(confName)
 		atomic.StoreInt64(&lastActive, time.Now().Unix())
 		for ctx.Err() == nil {
-			var confName = ".config"
-			if !vfs.IsSpecialName(confName) {
-				confName = ".jfs" + confName
+			if confStat.Ino == 0 {
+				_ = syscall.Stat(filepath.Join(mp, confName), &confStat)
```
davies, line 152: Without this, how do we know the the mount point is accessible?

polyrabbit: Right, I thought the liveness is also checked via `loadConfig(mp)`, didn’t realize it's also only executed once.

Issue #6181, opened 2025-06-11
JuiceFS Gateway: CompleteMultipartUpload inserts null xattr when ObjTag is enabled on Kylin


aucusaga:
**What happened**:

When using the `CompleteMultipartUpload` method via MinIO Gateway with JuiceFS, an internal error occurs due to a NULL value being passed to the `SetXattr` method. The error results in a PostgreSQL constraint violation:
`time="2025-06-11T06:30:46Z" level=error msg="2025/06/11 06:30:46.295435 juicefs[6] <ERROR>: error: ERROR: null value in column \"value\" violates not-null constraint (SQLSTATE 23502)"    source=minio`

This appears to be caused by an attempt to set an xattr with a NULL `value`, which is not allowed by the `NOT NULL` constraint in the JuiceFS metadata database schema. Stack trace excerpt:
<img width="849" alt="Image" src="https://github.com/user-attachments/assets/243046b7-35da-4af3-b415-a65207c9b81b" />

**What you expected to happen**:

JuiceFS should either reject NULL xattr values early or gracefully handle them without passing to the metadata layer. The `SetXattr` call should validate and skip setting attributes with nil values.

**How to reproduce it (as minimally and precisely as possible)**:
1. Start JuiceFS Gateway with `--enable-obj-tag` or with `gConf.ObjTag` set to `true`.
2. Initiate a multipart upload via the S3 SDK, targeting the JuiceFS Gateway endpoint:
    - Generate a pre-signed URL for each part.
    - Return the `UploadID` and part URLs to the frontend.
3. Upload each part using the pre-signed URLs.
4. Call `CompleteMultipartUpload` via the S3 API.

**Anything else we need to know?**

A null check should be added before calling SetXattr in CompleteMultipartUpload to prevent inserting null values into the metadata store.

**Environment**:
- JuiceFS version (use `juicefs --version`) or Hadoop Java SDK version: v1.2.3
- Cloud provider or hardware configuration running JuiceFS: postgresql+ minio
- OS (e.g `cat /etc/os-release`): Kylin Linux Advanced Server V10 (Halberd)
- Kernel (e.g. `uname -a`): 4.19.90-89.23.v2401.ky10.aarch64
- Object storage (cloud provider and region, or self maintained): minio
- Metadata engine info (version, cloud provider managed or self maintained): postgresql

PR #5258, opened 2024-10-28
gateway: fix etag and set xattr before complete multipart rename to ensure consistency

zhijian-pro:

PR #6182, opened 2025-06-11
fix(gateway): add a null check for tagStr in CompleteMultipartUpload

aucusaga:
### What this PR does

This PR fixes [issue #6181](https://github.com/juicedata/juicefs/issues/6181) by adding a `nil` check for the `tagStr` field before calling `SetXattr` in the `CompleteMultipartUpload` method. This prevents unnecessary errors caused by inserting a `nil` value into the metadata store.

### Why we need it

When `ObjTag` is enabled and no tag value is provided, the JuiceFS Gateway may attempt to set an extended attribute (`xattr`) with a null value. This leads to a runtime error when the underlying metadata engine (e.g., PostgreSQL) enforces a NOT NULL constraint.

### How I tested

This change has been tested locally with the following steps:

1. Start JuiceFS in Gateway mode.
2. Use the S3 SDK to initiate a multipart upload via the Gateway endpoint.
3. Upload all parts using pre-signed URLs.
4. Call `CompleteMultipartUpload` without specifying tags.

After applying the fix, the multipart upload completed successfully without triggering the null value error.

### Related issue
[#6181](https://github.com/juicedata/juicefs/issues/6181)

### DIFF
--- a/pkg/gateway/gateway.go
+++ b/pkg/gateway/gateway.go
@@ -1232,8 +1232,10 @@ func (n *jfsObjects) CompleteMultipartUpload(ctx context.Context, bucket, object
 			if eno != meta.ENOATTR {
 				logger.Errorf("get object tags error, path: %s, error: %s", n.upath(bucket, uploadID), eno)
 			}
-		} else if eno = n.fs.SetXattr(mctx, tmp, s3Tags, tagStr, 0); eno != 0 {
-			logger.Errorf("set object tags error, path: %s, tags: %s, error: %s", tmp, string(tagStr), eno)
+		} else if len(tagStr) > 0 {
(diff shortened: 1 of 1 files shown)


zhijian-pro:
This issue exists in v1.2.3, but it does not exist in the main line.
This issue has been fixed by https://github.com/juicedata/juicefs/pull/5258
We need to backport this PR to v1.2

davies:
> This issue exists in v1.2.3, but it does not exist in the main line. This issue has been fixed by #5258 We need to backport this PR to v1.2

Yes, this PR is useless, since the stored tags should not be “” if the store does not support it.  but we can still keep this patch if the old version has set it.

PR #6199, opened 2025-06-20
vfs/metrics: histogram bucket range of FUSE latency should exceed range of object storage

polyrabbit:
Currently, the latency range for object storage is [0.01, 168]s, but range for FUSE latency is [0.0001, 12.7]s. The gap is too large that FUSE latency p99 being very insensitive when object storage latency increases, and sometimes the p99 is even much smaller than average:

image

It is better to make FUSE latency range cover the range of object storage latency.

### DIFF
--- a/pkg/fs/fs.go
+++ b/pkg/fs/fs.go
@@ -196,7 +196,7 @@ func NewFileSystem(conf *vfs.Config, m meta.Meta, d chunk.ChunkStore, registry *
 		opsDurationsHistogram: prometheus.NewHistogram(prometheus.HistogramOpts{
 			Name:    "sdk_ops_durations_histogram_seconds",
 			Help:    "Operations latency distributions.",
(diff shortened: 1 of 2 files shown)


jiefenghuang:
factor 1.8的话，ms-s级别的区间精度降低了，这部分应该也是重点观察部分？

With a factor of 1.8, the precision across the millisecond‑to‑second interval actually decreases, these metrics should also be a key focus of observation？

polyrabbit:
`ExponentialBuckets(0.00001, 1.8, 30)` is a compromise between cacheReadHist(`ExponentialBuckets(0.00001, 2, 24)`) and objectReqsHistogram(`ExponentialBuckets(0.01, 1.5, 25)`). These parameters do not affect the avg metric (`bucket_sum/bucket_total`), but only impact p99. 

To cover the range of `objectReqsHistogram`, we either need to increase `factor` or `count`. By using a factor of 1.8, we can even reduce the `count` by one, which reduces one metric.

PR #6200, opened 2025-06-20
meta: fix leaking objects when quota is full

polyrabbit:
When quota/volume is full, no one will be tracking those pending slices, we should delete them as much as possible.

### DIFF
--- a/pkg/vfs/writer.go
+++ b/pkg/vfs/writer.go
@@ -203,16 +203,15 @@ func (c *chunkWriter) commitThread() {
 
 		f.Lock()
 		if err != 0 {
-			if err == syscall.ENOENT {
+			if err == syscall.ENOENT || err == syscall.ENOSPC || err == syscall.EDQUOT {
 				go func(id uint64, length int) {
 					_ = f.w.store.Remove(id, length)
 				}(s.id, int(s.length))
-			} else if err != syscall.ENOSPC && err != syscall.EDQUOT {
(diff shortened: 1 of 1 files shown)


jiefenghuang (approved):

pkg/vfs/writer.go
```diff
@@ -203,16 +203,15 @@ func (c *chunkWriter) commitThread() {
 
 		f.Lock()
 		if err != 0 {
-			if err == syscall.ENOENT {
+			if err == syscall.ENOENT || err == syscall.ENOSPC || err == syscall.EDQUOT {
 				go func(id uint64, length int) {
 					_ = f.w.store.Remove(id, length)
 				}(s.id, int(s.length))
-			} else if err != syscall.ENOSPC && err != syscall.EDQUOT {
-				logger.Warnf("write inode:%d error: %s", f.inode, err)
```
jiefenghuang, line 211: Keep the original error output here.

polyrabbit: Updated

polyrabbit:
Also, why do we convert `ENOENT` to `EBADF` at the end of a `Write` request? In one scenario, if the file is deleted by another node during a write process, returning `ENOENT` should be more intuitive, as the user can realize the file no longer exists. But why do we now return `EBADF`? The file handle is still valid.

davies:
write() shoud not return ENOENT, the fd is invalid after the inode is deleted.

polyrabbit:
OK, I will keep it as it is.

PR #6201, opened 2025-06-23
objbench: correct the time-consuming calculation method

zhijian-pro:


cmd/objbench.go
```diff
@@ -188,6 +187,13 @@ func objbench(ctx *cli.Context) error {
 		listCount = 1000
 	}
 	threads := int(ctx.Uint("threads"))
+	if threads > bCount || threads > sCount {
```
jiefenghuang, line 190: Adjust threads separately for different scenarios？

PR #6217, opened 2025-06-28
meta/kv: fix potential data corruption issues caused by clock diff and tikv gc.

mzygQAQ:
In the current TiKV meta implementation, garbage collection (GC) uses local timestamps, while the timestamps (TSO) allocated for actual transaction writes are uniformly assigned by the Placement Driver (PD).

​​Potential Risks:​​

​​If the JuiceFS client machine's clock lags behind PD's time:​​
GC may fail to effectively clean up data.

​​If the client machine's clock significantly leads PD's time:​​
The safepoint could be advanced to a timestamp that disrupts transactional behavior. This may break ongoing transactions/queries (e.g., triggering tikverr.ErrGCTooEarly).

### DIFF
--- a/pkg/meta/tkv_tikv.go
+++ b/pkg/meta/tkv_tikv.go
@@ -37,8 +37,8 @@ import (
 	"github.com/tikv/client-go/v2/tikv"
 	"github.com/tikv/client-go/v2/txnkv"
 	"github.com/tikv/client-go/v2/txnkv/txnutil"
-	"go.uber.org/zap"
 	pd "github.com/tikv/pd/client"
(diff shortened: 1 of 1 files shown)


davies (approved):

daxiguafan:
为什么不在1.3.3合入？

jiefenghuang:
> 为什么不在1.3.3合入？

默认gc是3小时, 节点之间的时钟偏移差距过大导致事务失败的概率很小(基本都是小事务). 如果有需求, 可以考虑升级到1.4.1版本.

PR #6218, opened 2025-06-28
chunk: disable `writeback` flag in `cache-partial-only` mode

polyrabbit:
Two minor fixes:
1. `cache-partial-only` and `writeback` are two conflicting flags and cannot be enabled together. Stage files must always be linked to the cache directory; otherwise, even the local client won't be able to access those blocks before background upload is completed.
2. Stage metrics are inaccurate in some corner cases. For example, when two JuiceFS instances of the same volume are scheduled to the same node and both processes upload the stage directory simultaneously, the metrics will never return to zero.

### DIFF
--- a/pkg/chunk/cached_store.go
+++ b/pkg/chunk/cached_store.go
@@ -235,6 +235,7 @@ func (s *rSlice) Remove() error {
 	for i := 0; i <= lastIndx; i++ {
 		if e := s.delete(i); e != nil {
 			err = e
+			logger.Warnf("Failed to delete block %s: %s", s.key(i), e)
 		}
 	}
 	return err
@@ -456,7 +457,7 @@ func (s *wSlice) upload(indx int) {
 			block.Acquire()
 			err := utils.WithTimeout(func() (err error) { // In case it hangs for more than 5 minutes(see fileWriter.flush), fallback to uploading directly to avoid `EIO`
 				defer block.Release()
-				stagingPath, err = s.store.bcache.stage(key, block.Data, s.store.shouldCache(blen))
+				stagingPath, err = s.store.bcache.stage(key, block.Data)
 				if err == nil && stageFailed { // upload thread already marked me as failed because of timeout
 					_ = s.store.bcache.removeStage(key)
 				}
@@ -594,6 +595,10 @@ func (c *Config) SelfCheck(uuid string) {
 			c.Prefetch = 0
 		}
 		c.CacheDir = "memory"
+	}
+	if !c.CacheFullBlock && c.Writeback {
+		logger.Warnf("writeback is disabled in non-full-block cache mode")
+		c.Writeback = false
 	}
 	if !c.Writeback {
 		if c.UploadDelay > 0 || c.UploadHours != "" {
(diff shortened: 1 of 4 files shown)


pkg/chunk/cached_store.go
```diff
@@ -595,6 +596,10 @@ func (c *Config) SelfCheck(uuid string) {
 		}
 		c.CacheDir = "memory"
 	}
+	if !c.CacheFullBlock && c.Writeback {
```
jiefenghuang, line 599: We can just add a warning.

polyrabbit: I suppose enabling both flags will lead to `EIO`, which is a severe error. `SelfCheck` should prevent such usage and a warning is not enough. It's the same rule as we also forbid using `--writeback` and `--cache-size=0` together in `SelfCheck`.

jiefenghuang: 纯写场景可以不需要读缓存link
we don't need the cached link if in write-only mode

pkg/chunk/cached_store.go
```diff
@@ -998,7 +1003,13 @@ func (store *cachedStore) uploadStagingFile(key string, stagingPath string) {
 	f, err := openCacheFile(stagingPath, blen, store.conf.CacheChecksum)
 	if err != nil {
 		if store.isPendingValid(key) {
-			logger.Errorf("Open staging file %s: %s", stagingPath, err)
+			if os.IsNotExist(err) {
```
jiefenghuang, line 1006: 缓存的多客户端协同处理可能需要别的机制, 读缓存也是

coordinated handling of cache among multiple clients may require a different mechanism.

pkg/chunk/disk_cache.go
```diff
@@ -750,30 +746,30 @@ func (cache *cacheStore) add(key string, size int32, atime uint32) {
...
 	err := cache.flushPage(stagingPath, data, false)
-	if err == nil {
+	if err != nil {
+		return stagingPath, err
+	}
+	path := cache.cachePath(key)
+	cache.createDir(filepath.Dir(path))
+	if err = os.Link(stagingPath, path); err == nil {
+		cache.add(key, -int32(len(data)), uint32(time.Now().Unix()))
 		cache.m.stageBlocks.Add(1)
```
jiefenghuang, line 768: 这里link的失败不应该影响stage的metrics统计

Failures of the link here should not affect the stage’s metrics.

pkg/chunk/cached_store.go
```diff
@@ -1043,16 +1054,19 @@ func (store *cachedStore) addDelayedStaging(key, stagingPath string, added time.
 	store.pendingMutex.Unlock()
 	if item.uploading {
 		logger.Debugf("Key %s is ignored since it's already being uploaded", key)
-		return true
+		return false // don't update staging metrics
```
jiefenghuang, line 1057: 返回值更改为error？原先bool代表是否成功，下面warn日志跟其他地方uploader重复了

Change the return value to error? Previously it was a bool indicating success or failure.
Also, the warning log below is duplicated with other places in the uploader.

polyrabbit:
OK, I'll keep this patch internally.

PR #6232, opened 2025-07-01
object/tos: verify using the crc32 algorithm

zhijian-pro:
It is mainly to reduce CPU usage, the CPU consumption of using crc32c algorithm is lower than crc64. 
However, the drawback is that if data corruption occurs during transmission, the error will be discovered only when reading.

### DIFF
--- a/pkg/object/tos.go
+++ b/pkg/object/tos.go
@@ -293,6 +293,10 @@ func newTOS(endpoint, accessKey, secretKey, token string) (ObjectStorage, error)
 	if err != nil {
 		return nil, fmt.Errorf("invalid endpoint: %v, error: %v", endpoint, err)
 	}
+	disableChecksum := strings.EqualFold(uri.Query().Get("disable-checksum"), "true")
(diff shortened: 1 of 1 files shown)


pkg/object/tos.go
```diff
@@ -289,7 +300,8 @@ func newTOS(endpoint, accessKey, secretKey, token string) (ObjectStorage, error)
 		hostParts[1]+"."+hostParts[2],
 		tos.WithRegion(strings.TrimPrefix(hostParts[1], "tos-")),
 		tos.WithCredentials(credentials),
-		tos.WithEnableVerifySSL(httpClient.Transport.(*http.Transport).TLSClientConfig.InsecureSkipVerify))
+		tos.WithEnableVerifySSL(httpClient.Transport.(*http.Transport).TLSClientConfig.InsecureSkipVerify),
+		tos.WithEnableCRC(false))
```
davies: similar to S3, we can disable the checksum using `?disable-checksum=true`

PR #6233, opened 2025-07-01
cmd/sync: clear the redundant delimiters in the files-from

zhijian-pro:
In order to be able to handle the format `a//b`

PR #6237, opened 2025-07-02
Revert "cmd/sync: clear the redundant delimiters in the files-from"

davies:
Reverts juicedata/juicefs#6233


zhijian-pro:
Unable to handle the case where the object contains `//`, as this `a//b` itself may be an object name, which should be handled by the user.

PR #6239, opened 2025-07-03
object: add support for enabling/disabling default CRC checksum via query parameters

zhijian-pro:


pkg/object/oss.go
```diff
@@ -400,8 +400,12 @@ func newOSS(endpoint, accessKey, secretKey, token string) (ObjectStorage, error)
 	config.RetryMaxAttempts = oss.Ptr(1)
 	config.ConnectTimeout = oss.Ptr(time.Second * 2)
 	config.ReadWriteTimeout = oss.Ptr(time.Second * 5)
-	config.DisableUploadCRC64Check = oss.Ptr(true)
-	config.DisableDownloadCRC64Check = oss.Ptr(true)
+	enableChecksum := strings.EqualFold(uri.Query().Get("enable-default-checksum"), "true")
```
davies: use disable-checksum, but the default value is true

PR #6271, opened 2025-07-18
cache: support LRU eviction policy

polyrabbit:
`2-random` cache eviction policy works well for random read scenarios, but there are cases where a strict LRU is more appropriate — for example, during checkpoint loading in LLM training, where newly cached data is always more valuable than old ones. The current `2-random` eviction policy make performance unpredictable when cache is full.

This PR adds support for a third value - `lru`, to the `--cache-eviction` parameter. Additional cost of maintaining a LRU includes:
* **Memory**: `cache-items * (sizeof(int64) + sizeof(int64))`
* **Latency**: `log₂(cache-items)`

**Test case**:

When `cache-size` is much smaller than dataset size (which is the normal case), the LRU policy even has a better hit rate than `2-random`:

image


pkg/chunk/disk_cache.go
```diff
@@ -782,6 +796,50 @@ func (cache *cacheStore) uploaded(key string, size int) {
 	cache.add(key, int32(size), 0)
 }
 
+// locked
+func (cache *cacheStore) evictionIter(yield func(k cacheKey, v *cacheItem) bool) {
+	switch cache.eviction {
+	case EvictionLRU:
+		for cache.lruHeap.Len() > 0 {
+			item := cache.lruPop()
+			if item.size < 0 { // It's OK to pop staging blocks from heap
```
jiefenghuang: if stage cached -> evict before uploaded -> uploaded, stage cache will never be cleaned?

pkg/chunk/disk_cache.go
```diff
@@ -609,10 +613,12 @@ func (cache *cacheStore) remove(key string, staging bool) {
 		if it.size > 0 {
 			cache.used -= int64(it.size + 4096)
 			delete(cache.keys, k)
+			cache.lruRemove(it.index)
```
jiefenghuang: 这块代码是否可以抽象出一个不同cache（目前三种，None，LRU， 2RANDOM）实现？ex. 都有cache.keys引用，同时提供add,remove,iter接口，后面拓展其他cache实现也清晰一些。

Can this part of code be refactored to abstract the different cache implementations (currently three types: None, LRU, 2RANDOM)? For example, by defining a common interface that provides cache.keys access, plus add, remove, and iter methods. That way, it would be cleaner to extend with additional cache types in the future.

polyrabbit: Good idea, so we won't need to handle so many if conditions in the code. I will find time to refactor the cache eviction code later to avoid a huge PR. For now, let's keep it simple and atomic.

pkg/chunk/disk_cache.go
polyrabbit: Right, by design staging blocks should not be added to LRU eviction heap. I missed that in `cacheStore.add()`. Thanks for the reminder — I’ve fixed it and added a test case to cover this scenario.

polyrabbit:
Refactored so that each eviction policy is encapsulated in its own struct, and each struct maintains its own `keys`. This way, the default `2-random` policy retains its original memory footprint without carrying unnecessary fields for other policies.

pkg/chunk/disk_cache.go
```diff
@@ -647,20 +640,17 @@ func (cache *cacheStore) load(key string) (ReadCloser, error) {
...
 	cache.Lock()
-	if err == nil {
-		if it, ok := cache.keys[k]; ok {
-			// update atime
-			cache.keys[k] = cacheItem{it.size, uint32(time.Now().Unix())}
-		}
-	} else if it, ok := cache.keys[k]; ok {
-		if it.size > 0 {
+	if err != nil {
+		if it := cache.keys.remove(k, false); it != nil {
```
jiefenghuang, line 659: staging = true

pkg/chunk/disk_cache.go
```diff
@@ -605,17 +605,10 @@ func (cache *cacheStore) remove(key string, staging bool) {
...
-			delete(cache.keys, k)
-		} else if !staging {
-			path = "" // for staging block
-		} else {
-			delete(cache.keys, k)
-		}
-	} else if cache.scanned {
-		path = "" // not existed
+	if it := cache.keys.remove(k, staging); it != nil {
+		cache.used -= int64(it.size + 4096)
```
jiefenghuang: if staging = true, "if it.size > 0" is still needed.

pkg/chunk/disk_cache.go
```diff
@@ -684,16 +674,9 @@ func (cache *cacheStore) exist(key string) (bool, error) {
...
-			cache.keys[k] = cacheItem{it.size, uint32(time.Now().Unix())}
-		}
 		return true, nil
-	} else if it, ok := cache.keys[k]; ok {
-		if it.size > 0 {
-			cache.used -= int64(it.size + 4096)
-		}
-		delete(cache.keys, k)
+	} else if it := cache.keys.remove(k, false); it != nil {
+		cache.used -= int64(it.size + 4096)
```
jiefenghuang, line 690: remove staging and "if it.size > 0" is still needed.

pkg/chunk/cache_eviction.go
```diff
@@ -0,0 +1,375 @@
...
+	if iter, ok := p.keys[key]; !ok {
+		iter = &lruItem{cacheItem: item, pos: notInLru}
+		p.keys[key] = iter
+		if iter.size > 0 { // don't add staging blocks to lru as they should not be evicted in `cleanupFull`
+			heap.Push(&p.lruHeap, heapItem{iter, &key})
+		}
+	} else {
+		iter.cacheItem = item
+		if iter.pos == notInLru {
+			heap.Push(&p.lruHeap, heapItem{iter, &key})
```
jiefenghuang: if iter.size > 0?
check in disk_cache.go:add func

pkg/chunk/disk_cache.go
polyrabbit: Right, I missed that, updated.

pkg/chunk/cache_eviction.go
polyrabbit: Updated, although in normal cases, a `size < 0` and indexed item should not reach here.

pkg/chunk/disk_cache.go
polyrabbit: I suppose it's dangerous to silently remove a staging block in any case. For a transient IO error, a later retry may succeed; for an unrecoverable IO error, the failed block should be kept for subsequent analysis. This is different from a cache block.

pkg/chunk/disk_cache.go
polyrabbit: Same above - dangerous to remove a staging block

Issue #97, opened 2021-01-21
Enable Redis client cache


davies:
**What would you like to be added**:

We can enable the redis client cache by command options, which could be useful for readonly workload.

**Why is this needed**:

Faster is always good.

davies:
After looking into the details of client cache, it's complicated, so we move this out of 1.0 milestone.

rueian:
Hi @davies,

Would you consider using https://github.com/redis/rueidis? It has the client cache implemented.

PR #6272, opened 2025-07-18
Redis csc 1.3.0

valandreev:
added client side meta cache for Redis based meta engines

### DIFF
--- a/docs/en/reference/redis-csc.md
+++ b/docs/en/reference/redis-csc.md
@@ -1,6 +1,6 @@
 # Redis Client-Side Caching Support in JuiceFS
 
-Starting with version 7.4, Redis provides [Client-Side Caching](https://redis.io/docs/latest/develop/reference/client-side-caching/) which allows clients to maintain local caches of data in a faster and more efficient way.
+Starting with version 7.4, Redis provides [Client-Side Caching](https://redis.io/docs/latest/develop/reference/client-side-caching/) which allows clients to maintain local caches of data in a faster and more efficient way. JuiceFS includes full support for this feature, offering significant performance improvements for metadata operations.
 
 ## How it works
 


pkg/meta/redis.go
```diff
@@ -254,20 +278,104 @@ func newRedisMeta(driver, addr string, conf *Config) (Meta, error) {
...
+
+			// Preload the first 10000 inodes into the cache in a safe background goroutine
+			// Wrap with a recovery function to prevent any panics from affecting the mount process
+			go func() {
+				defer func() {
+					if r := recover(); r != nil {
+						logger.Errorf("Recovered from panic in cache preloading goroutine: %v", r)
+					}
+				}()
+				m.preloadInodeCache(10000)
```
davies, line 371: there should be a option for this, otherwise it could overload the Redis when there are many clients.

pkg/meta/redis_attr_csc.go
```diff
@@ -0,0 +1,61 @@
...
+	if err != nil {
+		return errno(err)
+	}
+
+	m.parseAttr(a, attr)
+
+	// Add to cache
+	cachedAttr := *attr
+	m.cacheMu.Lock()
+	m.inodeCache.Add(inode, &cachedAttr)
```
davies, line 49: there could be a race when a notification is received AFTER GET returned inmediately.

pkg/meta/redis_csc.go
```diff
@@ -0,0 +1,288 @@
...
+		}()
+		m.handleCacheInvalidation()
+	}()
+
+	return nil
+}
+
+// handleCacheInvalidation processes invalidation messages from Redis
+func (m *redisMeta) handleCacheInvalidation() {
+	ch := m.cacheSubscription.Channel()
```
davies, line 73: the tracking and notifications are bind to each connection. Redis client use a pool of connections for concurrent requests. What will happen when  a connection is closed?

pkg/meta/redis_csc.go
```diff
@@ -0,0 +1,288 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	// Enable tracking - first disable tracking if already enabled
+	_ = m.rdb.Do(ctx, "CLIENT", "TRACKING", "OFF").Err() // Ignore errors if not previously enabled
+
+	// Always use BCAST mode for simplicity
+	err := m.rdb.Do(ctx, "CLIENT", "TRACKING", "ON", "BCAST").Err()
```
davies, line 50: BCAST mode may not work when there are many connections (>100). Especially when a single mount point may use tens of connections.

davies:
This is cool overall, thanks! We should be very careful on cache consistency.

jiefenghuang:
related issue: #97

目前已经有kernel cache + openfile cache（chunk cache），更通用一些，元数据引擎的客户端缓存是否必须，有哪些场景需要做这样一个替换，是否真的必要？

JuiceFS already have a kernel cache plus an open‑file cache (chunk cache), which are more general-purpose. Is a client-side cache in the metadata necessary? In what scenarios would it be useful to add or switch to such a cache？

valandreev:
In edge scenarios, when the latency to the meta engine exceeds 20-25ms, JuiceFS becomes slow; client-side caching makes JuiceFS usable.

polyrabbit:
The more attractive aspect of CSC is that server can proactively invalidate client’s cache. This allows us to safely set a longer cache TTL and significantly reduce redundant requests to Redis. 

Of course, it would be even better if this invalidation logic could be integrated into existing caches, so we don’t have to maintain another layer of cache beneath a general-purpose one.

docs/en/reference/redis-csc.md
```diff
@@ -0,0 +1,90 @@
...
+7. File read operations benefit from chunk-level caching to improve read performance for frequently accessed files
+
+## Troubleshooting
+
+If you experience crashes or instability with CSC enabled:
+
+1. Update to the latest JuiceFS version which contains important fixes for CSC
+2. Try reducing the cache size with `client-cache-size`
+3. Check Redis server logs for any memory or client tracking issues
+4. Make sure your Redis server version is 7.4 or higher
```
jiefenghuang: 6.0

jiefenghuang:
closed by #6495

PR #6302, opened 2025-08-11
cmd/mount: add a warning message when `negative-entry-cache` is enabled on kernel versions below 5.11

polyrabbit:
When the kernel has a negative dentry cached for a directory, concurrent `mkdir -p` calls from multiple nodes can still result in a `File exists` error.

Reproduce:

1. On node 1, `stat` or `ls` a non-existent directory to populate the negative dentry cache for a path.
2. On node 2, create the directory at that same path.
3. On node 1, running `mkdir -p` for the same directory will now fail with a `File exists` error.

image

This is a simple reproduction flow. For a user calling Python's `os.makedirs(exist_ok=True)`, this bug can make it seem as if the `exist_ok=True` parameter is not working on JuiceFS.

This bug is fixed on kernel version 6.6 ([fuse: invalidate dentry on EEXIST creates or ENOENT deletes · torvalds/linux@7d875e6](https://github.com/torvalds/linux/commit/7d875e66859a4359acd29ce0d188e1aff048e7ed)). Before that, users should be warned about the potential stale cache issue.

### DIFF
--- a/cmd/mount_unix.go
+++ b/cmd/mount_unix.go
@@ -1003,17 +1003,20 @@ func mountMain(v *vfs.VFS, c *cli.Context) {
 	conf.DirEntryTimeout = utils.Duration(c.String("dir-entry-cache"))
 	conf.NegEntryTimeout = utils.Duration(c.String("negative-entry-cache"))
 	conf.ReaddirCache = c.Bool("readdir-cache")
+	major, minor := utils.GetKernelVersion()
 	if conf.ReaddirCache {
 		if conf.AttrTimeout == 0 {
 			logger.Warnf("readdir-cache is enabled without attr-cache, it's performance may be affected")
 		}
-		major, minor := utils.GetKernelVersion()
 		if major < 4 || (major == 4 && minor < 20) {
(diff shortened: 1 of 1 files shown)


jiefenghuang:
1. 在5.4上可以复现，单个挂载点上stat后mkdir，从accesslog上看，没有额外的lookup操作
2. 在5.15和6.6上没法复现，单个挂载点上stat后mkdir，从accesslog上看，都会有额外的lookup操作（强制刷新negtive dentry cache）

相关的patch可能是这个 https://github.com/torvalds/linux/commit/df8629af293493757beccac2d3168fe5a315636e；5.11左右的有几个增强revalidate的patch。

5.11的patch是前置刷新cache，6.6是后置刷新cache，加上[mkdir工具](https://github.com/coreutils/gnulib/blob/master/lib/mkdir-p.c#L155)/python库在mkdir失败后都会stat查询double check，保证了不冲突

jiefenghuang:
mark: python os.makedirs在mkdir失败后，有容错处理，不会出现cache并发问题。
image

PR #6320, opened 2025-08-23
fix: correct trash check logic in compactChunk

CodeRusher:
- Fix incorrect toTrash(0) call to use actual parent inode
- Resolve delayed slice deletion judgment in compaction process

### DIFF
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -2316,8 +2316,14 @@ func (m *baseMeta) compactChunk(inode Ino, indx uint32, once, force bool) {
 		return
 	}
 
+	var a Attr
+	st = m.en.doGetAttr(Background(), inode, &a)
+	if st != 0 {
+		return
+	}
+
 	var dsbuf []byte
-	trash := m.toTrash(0)
+	trash := m.toTrash(a.Parent)
 	if trash {
 		dsbuf = make([]byte, 0, len(compacted)*12)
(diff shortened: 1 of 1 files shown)


pkg/meta/base.go
```diff
@@ -2316,8 +2316,14 @@ func (m *baseMeta) compactChunk(inode Ino, indx uint32, once, force bool) {
...
 
+	var a Attr
+	st = m.en.doGetAttr(Background(), inode, &a)
+	if st != 0 {
+		return
+	}
+
 	var dsbuf []byte
-	trash := m.toTrash(0)
+	trash := m.toTrash(a.Parent)
```
SandyXSD, line 2326: Compacted slices should always be kept for `TrashDays`, no matter it belongs to a trash file or not.

CodeRusher: Does this mean the code logic here can be removed? trash is always false, and dsbuf is always nil.

```
var dsbuf []byte
trash := m.toTrash(0)
if trash {
dsbuf = make([]byte, 0, len(compacted)*12)
for _, s := range compacted {
	if s.id > 0 {
		dsbuf = append(dsbuf, m.encodeDelayedSlice(s.id, s.size)...)
	}
}
}
```

CodeRusher: yes, get it!

Issue #6323, opened 2025-08-25
Stale attribute cache breaks Close-to-Open consistency


yangliping:
**What happened**:

Client A appends data to a shared file and then closes it to ensure all data is flushed. While the kernel’s attribute cache on Client B is still valid, Client B opens and reads the file but only retrieves data up to the previously cached size—violating Close-to-Open consistency.

**What you expected to happen**:

Close-to-Open consistency is achieved under the default `attr_timeout=1` setting.

**How to reproduce it (as minimally and precisely as possible)**:

As stated above.

**Anything else we need to know?**

N/A

**Environment**:
- JuiceFS version (use `juicefs --version`) or Hadoop Java SDK version:

```
# juicefs --version
juicefs version 1.3.0+2025-07-03.30190ca
```

- Cloud provider or hardware configuration running JuiceFS:

KVM

- OS (e.g `cat /etc/os-release`):

```
# cat /etc/os-release
PRETTY_NAME="Ubuntu 22.04.5 LTS"                                                                                                                                NAME="Ubuntu"
VERSION_ID="22.04"                                                                                                                                              VERSION="22.04.5 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy                                                                                                                                          ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

- Kernel (e.g. `uname -a`):

```
# uname -a
Linux testuser-KVM 5.15.0-130-generic #140-Ubuntu SMP Wed Dec 18 17:59:53 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
```

- Object storage (cloud provider and region, or self maintained):

self maintained MinIO

- Metadata engine info (version, cloud provider managed or self maintained):

self maintained Redis

- Network connectivity (JuiceFS to metadata engine, JuiceFS to object storage):

Ethernet

- Others:

solracsf:
> While the kernel’s attribute cache on Client B is still valid

If the cache is still valid, there are only 2 ways to invalidate it; wait for [cache](https://juicefs.com/docs/community/command_reference#mount-metadata-cache-options) to expire, or simply disabling it.

[...]

PR #6326, opened 2025-08-25
fuse: add inode cache invalidation on file open

yangliping:
Add InodeNotify call in Open method to force kernel cache invalidation when files are opened. This ensures that clients get fresh inode attribute information instead of relying on potentially stale cached data.

The change calls fsserv.InodeNotify() with the opened file's inode ID to notify the kernel to invalidate its cached attribute information. This improves consistency across multiple clients accessing the same files.

Resolves: #6323

### DIFF
--- a/pkg/fuse/fuse.go
+++ b/pkg/fuse/fuse.go
@@ -250,8 +250,9 @@ func (fs *fileSystem) Open(cancel <-chan struct{}, in *fuse.OpenIn, out *fuse.Op
 		out.OpenFlags |= fuse.FOPEN_DIRECT_IO
 	} else if entry.Attr.KeepCache {
 		out.OpenFlags |= fuse.FOPEN_KEEP_CACHE
+	} else {
+		fsserv.InodeNotify(uint64(in.NodeId), -1, 0)
 	}
-	fsserv.InodeNotify(uint64(in.NodeId), -1, 0)
 	return 0
 }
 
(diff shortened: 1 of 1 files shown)


polyrabbit:
Be careful to not introduce a deadlock when calling `InodeNotify`.

yangliping:
### 1. Frequency  
InodeNotify is invoked at most once per file—right when it is opened.

### 2. Safety  
It never endangers the JuiceFS daemon:  
- Non-blocking, one-way invalidation signal.  
- Executed after the file is already opened.  
- Failure does not roll back the open.  

### 3. Deadlock Risk  
- Pure user-space → kernel notification; no two-way locking.  
- go-fuse’s InodeNotify is thread-safe and deadlock-free.  
- Called with no internal JuiceFS locks held.  
- The kernel handles invalidation under its own concurrency rules.

pkg/fuse/fuse.go
```diff
@@ -251,6 +251,7 @@ func (fs *fileSystem) Open(cancel <-chan struct{}, in *fuse.OpenIn, out *fuse.Op
 	} else if entry.Attr.KeepCache {
 		out.OpenFlags |= fuse.FOPEN_KEEP_CACHE
 	}
+	fsserv.InodeNotify(uint64(in.NodeId), -1, 0)
```
davies: The  OpenFlags is used to tell kernel to invalidate data cache.

The returned attribute of Open should overwrite stale attr in kernel, so InodeNotify is not needed, I think.

yangliping: As far as I know, certain OpenFlags can affect how the kernel handles file caching, but there is no specific flag dedicated to 'invalidating the cache'.

jiefenghuang: maybe change to this, only invalidate cache when mtime changed. @yangliping 
```golang
else if entry.Attr.KeepCache {
	out.OpenFlags |= fuse.FOPEN_KEEP_CACHE
} else  {
	fsserv.InodeNotify(uint64(in.NodeId), -1, 0)
}
```

jiefenghuang (approved):

Issue #6328, opened 2025-08-25
Python SDK clone method docstring is misleading


Looka149:
**What happened**:

Fix Python SDK clone method docstring

The docstring says "Clone a file" but it actually supports directories too.
The `clone` method in the Python SDK uses the `clone` function in the Meta interface.
I tested it and directory cloning works.

https://github.com/juicedata/juicefs/blob/00b5ebcffd22265c58e248f576a8ffa0cb3705b9/sdk/python/juicefs/juicefs/juicefs.py#L352-L354

Should be:

```diff
- """Clone a file."""
+ """Clone a file or directory."""
```

This matches the Meta interface comment.

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**

**Environment**:
- JuiceFS version (use `juicefs --version`) or Hadoop Java SDK version:
- Cloud provider or hardware configuration running JuiceFS:
- OS (e.g `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Object storage (cloud provider and region, or self maintained):
- Metadata engine info (version, cloud provider managed or self maintained):
- Network connectivity (JuiceFS to metadata engine, JuiceFS to object storage):
- Others:

PR #6329, opened 2025-08-25
sdk: fix Python clone method docstring to include directory support

Looka149:
fix: https://github.com/juicedata/juicefs/issues/6328


Looka149:
@davies Thank you for the review and merge :)

PR #6332, opened 2025-08-27
sdk/python: raise exception when jfs_init returns 0

### LINKED ISSUE #6331: Python SDK: Client initialization does not raise exception on connection failure
**What happened**:

JuiceFS Python SDK doesn't raise exception when connection to metadata service fails.

Should raise exception like Java SDK does (`handle <= 0`).

```python
import juicefs
client = juicefs.Client(name='test', meta='redis://invalid-host:6379')
# No exception raised, but should fail
```

`check_error()` only checks `r < 0`, but `jfs_init` returns `0` on failure.

**Environment**:
- JuiceFS version (use `juicefs --version`) or Hadoop Java SDK version: main branch
- Kernel (e.g. `uname -a`):
- Others:

Looka149:
When `jfs_init` fails, it returns `0`, but Python SDK does not raise an exception because `check_error()` only checks `r < 0`. I modified `check_error()` to handle `jfs_init` failure case when it returns `0`.

This aligns with Java SDK behavior:
https://github.com/juicedata/juicefs/blob/f26ce2b1d5edbc7ca498fc79dd123fae860fe371/sdk/java/src/main/java/io/juicefs/JuiceFileSystemImpl.java#L465-L468

Changes
- Added special case for `jfs_init` when `r == 0` in `check_error()`
- Error message follows Java SDK format for consistency

Tests
```python
import juicefs
try:
    client = juicefs.Client(name='test', meta='redis://invalid-host:6379')
except OSError as e:
    print(f"Exception raised: {e}")  # Now works correctly
```

Fixes #6331

### DIFF
--- a/sdk/python/juicefs/juicefs/juicefs.py
+++ b/sdk/python/juicefs/juicefs/juicefs.py
@@ -35,7 +35,12 @@ XATTR_CREATE = 1
 XATTR_REPLACE = 2
 
 def check_error(r, fn, args):
-    if r < 0:
+    if fn.__name__ == "jfs_init" and r == 0:
+        name = args[0].decode()
+        e = OSError(f'JuiceFS initialized failed for {name}')
+        e.errno = 1
+        raise e
+    elif r < 0:
         formatted_args = []
         for arg in args[2:]:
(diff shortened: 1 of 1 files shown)


Looka149:
@davies Thank you for the review and merge :)

PR #6339, opened 2025-09-01
 cmd/sync: directly ignore the not exist paths that in the filefrom

zhijian-pro:


cmd/sync.go
```diff
@@ -175,6 +175,10 @@ func selectionFlags() []cli.Flag {
 			Name:  "files-from",
 			Usage: "read list of files or dirs to sync from FILE",
 		},
+		&cli.BoolFlag{
```
jiefenghuang: There’s no need to introduce any new configuration here？ 
The files-from list supports simple * wildcard matching—if a suffix ends with *, it should be treated as a prefix search. If the key doesn’t exist, just ignore it.

But this is slightly different from the original behavior.

davies: Since the help message only say "files or dirs", so we can drop the support of prefix (side effects).

davies: remove this flag, and change the behavior.

pkg/sync/sync.go
```diff
@@ -1508,9 +1508,14 @@ func produceFromList(tasks chan<- object.Object, src, dst object.ObjectStorage,
 }
 
 var ignoreDir = errors.New("ignore dir")
+var ignoreFiles int64
 
 func produceSingleObject(tasks chan<- object.Object, src, dst object.ObjectStorage, key string, config *Config) error {
 	obj, err := src.Head(ctx, key)
+	if !strings.HasSuffix(key, "/") && os.IsNotExist(err) {
```
davies: 这里的 key 已经没有 / 结尾了，可以去掉

davies (approved):

PR #6359, opened 2025-09-14
vfs: Add support to skip trash if 's' attribute is set

### LINKED ISSUE #5385: Support special directory that can always skip trash
Configure a directory (must be empty?) so that all files created under it will be deleted directly, ignoring the trash.

**Why is this needed**:
Some application may create a lot of temporary files that should be purged completely no matter the trash is enabled or not.

mcassaniti:
This change adds support for the FlagSkipTrash in the VFS and meta portions of the code and maps to `FS_SECRM_FL` / secure removal / 's' in chattr. Setting the attribute on a file will prevent the file from being sent to trash. New files and sub-directories will inherit the flag if the directory has the flag set.

Closes #5385

### DIFF
--- a/pkg/meta/base_test.go
+++ b/pkg/meta/base_test.go
@@ -2240,6 +2240,9 @@ func testTrash(t *testing.T, m Meta) {
 		t.Fatalf("entries: %d", len(entries))
 	}
 	entries = entries[:0]
+	if st := m.Unlink(ctx, parent, "f3"); st != 0 {
+		t.Fatalf("unlink secrmd/f3: %s", st)
+	}
 	if st := m.Rmdir(ctx, 1, "secrmd"); st != 0 {
 		t.Fatalf("rmdir secrmd: %s", st)
 	}
--- a/pkg/meta/base_test.go
+++ b/pkg/meta/base_test.go
@@ -2188,6 +2188,62 @@ func testTrash(t *testing.T, m Meta) {
 		t.Fatalf("entries: %d", len(entries))
 	}
 
+	// Selectively skip trash based on FS_SECRM_FL
+	if st := m.Mkdir(ctx, 1, "secrmd", 0755, 022, 0, &parent, attr); st != 0 {
+		t.Fatalf("mkdir secrmd: %s", st)
+	}
+	if st := m.SetAttr(ctx, parent, SetAttrFlag, 0, &Attr{Flags: FlagSkipTrash}); st != 0 {
+		t.Fatalf("setattr secrmd secrm: %s", st)
+	}
+	if st := m.Create(ctx, parent, "f1", 0644, 022, 0, &inode, attr); st != 0 {
+		t.Fatalf("create secrmd/f1: %s", st)
+	}
+	if st := m.GetAttr(ctx, inode, attr); st != 0 || (attr.Flags&FlagSkipTrash) == 0 {
+		t.Fatalf("getattr secrmd/f1(%d): %s, attr %+v", inode, st, attr)
+	}
+	if st := m.Unlink(ctx, parent, "f1"); st != 0 {
+		t.Fatalf("unlink secrmd/f1: %s", st)
+	}
+	if st := m.Readdir(ctx, TrashInode+1, 0, &entries); st != 0 {
+		t.Fatalf("readdir: %s", st)
+	}
+	if len(entries) != 12 {
+		t.Fatalf("entries: %d", len(entries))
+	}
+	entries = entries[:0]
+	if st := m.Mkdir(ctx, parent, "d1", 0755, 022, 0, &inode, attr); st != 0 {
+		t.Fatalf("mkdir secrmd/d1: %s", st)
+	}
+	if st := m.Rmdir(ctx, parent, "d1"); st != 0 {
+		t.Fatalf("rmdir secrmd/d1: %s", st)
+	}
+	if st := m.Readdir(ctx, TrashInode+1, 0, &entries); st != 0 {
+		t.Fatalf("readdir: %s", st)
+	}
+	if len(entries) != 12 {
+		t.Fatalf("entries: %d", len(entries))
+	}
+	entries = entries[:0]
+	if st := m.Create(ctx, parent, "f2", 0644, 022, 0, &inode, attr); st != 0 {
+		t.Fatalf("create secrmd/f2: %s", st)
+	}
+	if st := m.Create(ctx, parent, "f3", 0644, 022, 0, &inode, attr); st != 0 {
+		t.Fatalf("create secrmd/f3: %s", st)
+	}
+	if st := m.Rename(ctx, parent, "f2", parent, "f3", 0, &inode, attr); st != 0 {
+		t.Fatalf("rename secrmd/f2 -> f3: %s", st)
+	}
+	if st := m.Readdir(ctx, TrashInode+1, 0, &entries); st != 0 {
+		t.Fatalf("readdir: %s", st)
+	}
+	if len(entries) != 12 {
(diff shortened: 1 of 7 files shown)


mcassaniti:
I wanted to get feedback early before I went down the wrong path. I've only modified the SQL meta-data back-end but I'm hoping this will provide enough code for a reasonable review.

pkg/meta/sql.go
```diff
@@ -1830,6 +1833,9 @@ func (m *dbMeta) doUnlink(ctx Context, parent Ino, name string, attr *Attr, skip
 			if (n.Flags&FlagAppend) != 0 || (n.Flags&FlagImmutable) != 0 {
 				return syscall.EPERM
 			}
+			if ((pn.Flags&FlagSkipTrash) != 0 || (n.Flags&FlagSkipTrash) != 0) {
```
jiefenghuang: From the design and definition perspective, when deleting check only the flag of the file itself, independently of its parent directory.

pkg/meta/sql.go
```diff
@@ -1961,6 +1967,9 @@ func (m *dbMeta) doRmdir(ctx Context, parent Ino, name string, pinode *Ino, attr
 		if pn.Flags&FlagImmutable != 0 || pn.Flags&FlagAppend != 0 {
 			return syscall.EPERM
 		}
+		if pn.Flags&FlagSkipTrash != 0 {
```
jiefenghuang: same as unlink, we may remove this checking.

pkg/meta/sql.go
```diff
@@ -2206,6 +2218,12 @@ func (m *dbMeta) doRename(ctx Context, parentSrc Ino, nameSrc string, parentDst
 			if (dn.Flags&FlagAppend) != 0 || (dn.Flags&FlagImmutable) != 0 {
 				return syscall.EPERM
 			}
+			if spn.Flags&FlagSkipTrash != 0 {
```
jiefenghuang: For `rename`, we can skip inheritance, and keep the logic and usage simple. We only need to handle the `replace` scenario: if the destination file or directory is marked with `FlagSkipTrash`, it should be deleted directly.

mcassaniti:
Thanks for the feedback. Do you want an updated commit or should I force a new one?

jiefenghuang:
> Thanks for the feedback. Do you want an updated commit or should I force a new one?

An updated commit is fine.

pkg/meta/sql.go
```diff
@@ -2206,6 +2215,9 @@ func (m *dbMeta) doRename(ctx Context, parentSrc Ino, nameSrc string, parentDst
 			if (dn.Flags&FlagAppend) != 0 || (dn.Flags&FlagImmutable) != 0 {
 				return syscall.EPERM
 			}
+			if dpn.Flags&FlagSkipTrash != 0 || dn.Flags&FlagSkipTrash != 0 {
```
jiefenghuang: dpn.Flags&FlagSkipTrash != 0 is not needed?

mcassaniti: Yep, fixed.

mcassaniti:
I'm thinking more an integration test rather than a unit test. That way the testing will cover multiple backends at once when run as a matrix of tests. I just don't know which workflow to call a script from.

mcassaniti:
I've just extended the documentation but I'll need some help with:

* Chinese translation
* Adding integration tests

jiefenghuang:
> I'm thinking more an integration test rather than a unit test. That way the testing will cover multiple backends at once when run as a matrix of tests. I just don't know which workflow to call a script from.

You can add a unit test in base_test.go’s testTrash function, which will be executed with multiple metadata engines. I will add the Chinese documentation later.

pkg/meta/base_test.go
```diff
@@ -2188,6 +2188,62 @@ func testTrash(t *testing.T, m Meta) {
...
+	if st := m.Mkdir(ctx, 1, "secrmd", 0755, 022, 0, &parent, attr); st != 0 {
+		t.Fatalf("mkdir secrmd: %s", st)
+	}
+	if st := m.SetAttr(ctx, parent, SetAttrFlag, 0, &Attr{Flags: FlagSkipTrash}); st != 0 {
+		t.Fatalf("setattr secrmd secrm: %s", st)
+	}
+	if st := m.Create(ctx, parent, "f1", 0644, 022, 0, &inode, attr); st != 0 {
+		t.Fatalf("create secrmd/f1: %s", st)
+	}
+	if st := m.GetAttr(ctx, inode, attr); st != 0 || (attr.Flags&FlagSkipTrash) != 0 {
```
jiefenghuang: hi @mcassaniti, unit test failed
should be ```(attr.Flags&FlagSkipTrash) == 0```

mcassaniti: Whoops.

pkg/meta/base_test.go
```diff
@@ -2188,6 +2188,62 @@ func testTrash(t *testing.T, m Meta) {
...
+	if st := m.Rename(ctx, parent, "f2", parent, "f3", 0, &inode, attr); st != 0 {
+		t.Fatalf("rename secrmd/f2 -> f3: %s", st)
+	}
+	if st := m.Readdir(ctx, TrashInode+1, 0, &entries); st != 0 {
+		t.Fatalf("readdir: %s", st)
+	}
+	if len(entries) != 12 {
+		t.Fatalf("entries: %d", len(entries))
+	}
+	entries = entries[:0]
```
jiefenghuang, line 2242: Unlink f3 first 
```golang
if st := m.Unlink(ctx, parent, "f3", true); st != 0 {
	t.Fatalf("unlink secrmd/f3: %s", st)
}
```

jiefenghuang: You can run TestSQLiteClient to verify the unit test.

mcassaniti: No. It should be removed but not there. f3 should bypass the trash. It should be unlinked before rmdir secrmd.

The current test shows that when we unlink secrmd/f1 it is going to the trash. Since the flag is set to skip the trash, I have a bug at least for Redis.

pkg/meta/redis.go
```diff
@@ -1475,12 +1478,15 @@ func (m *redisMeta) doUnlink(ctx Context, parent Ino, name string, attr *Attr, s
 			return st
 		}
 	}
-	if trash == 0 {
-		defer func() { m.of.InvalidateChunk(inode, invalidateAttrOnly) }()
-	}
 	if attr == nil {
 		attr = &Attr{}
 	}
+	if (attr.Flags&FlagSkipTrash) != 0 {
```
jiefenghuang: change line 1484-1489 to 
```golang
defer func() {
    if trash == 0 {
         m.of.InvalidateChunk(inode, invalidateAttrOnly)
    }
}()
```

pkg/meta/redis.go
```diff
@@ -1541,6 +1547,10 @@ func (m *redisMeta) doUnlink(ctx Context, parent Ino, name string, attr *Attr, s
 			if (attr.Flags&FlagAppend) != 0 || (attr.Flags&FlagImmutable) != 0 {
 				return syscall.EPERM
 			}
+			if (attr.Flags&FlagSkipTrash) != 0 && trash > 0 {
```
jiefenghuang, line 1547: change to 
```golang
if (attr.Flags&FlagSkipTrash) != 0 {
	trash = 0
}
```

pkg/meta/redis.go
mcassaniti: While I didn't make your requested changes, take a look now. It should still remove the redundant code.

mcassaniti:
Thank you so much for your patience with this.

PR #6362, opened 2025-09-15
meta/sql: simplify isTableNotExistError to use string matching

Xuyuchao-juice:
fix https://github.com/juicedata/juicefs/issues/6304


jiefenghuang:
They are all low-frequency operations, so the performance impact is minimal.
Doing error checks with string comparisons directly could lead to compatibility issues in future versions.

PR #6364, opened 2025-09-16
do partial read if cache is disabled

### LINKED ISSUE #6363: Random reads will cause read amplification even though caching and readahead are disabled.
**What happened**:

juicefs mount without cache and readahead.
```
juicefs mount ... --max-readahead=0  --cache-size=0
```

a doubling of read amplification:

Image

Reduce read amplification.

Image

**Environment**:
- Kernel (e.g. `uname -a`):
- Others:

nautaa:
fix #6363

### DIFF
--- a/pkg/chunk/cached_store.go
+++ b/pkg/chunk/cached_store.go
@@ -150,7 +150,8 @@ func (s *rSlice) ReadAt(ctx context.Context, page *Page, off int) (n int, err er
 	s.store.cacheMiss.Add(1)
 	s.store.cacheMissBytes.Add(float64(len(p)))
 
-	if s.store.seekable && boff > 0 && len(p) <= blockSize/4 {
+	if s.store.seekable &&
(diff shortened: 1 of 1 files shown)


pkg/chunk/cached_store.go
```diff
@@ -150,7 +150,8 @@ func (s *rSlice) ReadAt(ctx context.Context, page *Page, off int) (n int, err er
 	s.store.cacheMiss.Add(1)
 	s.store.cacheMissBytes.Add(float64(len(p)))
 
-	if s.store.seekable && boff > 0 && len(p) <= blockSize/4 {
+	if s.store.seekable &&
```
jiefenghuang, line 153:
```golang
if s.store.seekable &&
		(!s.store.conf.CacheEnabled() || (boff > 0 && len(p) <= blockSize/4))
```

jiefenghuang: hi @nautaa，we can simplify this condition.

nautaa: resolved

jiefenghuang (approved):

Issue #6370, opened 2025-09-18
add --skip-trash to mount cmd


sw-cho:
**What would you like to be added**:

add --skip-trash to mount cmd

**Why is this needed**:
.

It is better to skip trash per client.
for example in Kubernetes, when Pods share temporary data through a PVC, it is useful to skip  trash is usefule.

jiefenghuang:
hi @sw-cho  #5385 should meet your needs?

PR #6371, opened 2025-09-18
mount: add skip-trash option

sw-cho:


jiefenghuang:
PR #6359 should meet the requirements of #6370.

jiefenghuang:
Closed for now due to no response for a long time. We can revisit the necessity of this later. If it’s just for temporary files, we can use #6359.

PR #6391, opened 2025-09-30
Add cooldown time for new block cache

polyrabbit:
## Problem

Currently, new blocks get atime=now on write (`flush`/`stage`), which prevents the eviction algorithm from distinguishing recently written and recently read blocks. Under heavy write traffic, this leads to evicting hot read blocks and degrades cache hit rate.

## Solution

Introduce an atime cooldown period to newly written blocks, and only bump to `now` on later read. This biases eviction algorithms (100% for LRU) toward never-read, write-produced blocks, allowing users to enable `write-back` and `cache-large-write` with more confidence, without worrying about polluting the hot cache.

### DIFF
--- a/pkg/chunk/disk_cache.go
+++ b/pkg/chunk/disk_cache.go
@@ -93,6 +93,9 @@ type cacheStore struct {
 	rawFull   bool
 	checksum  string // checksum level
 	uploader  func(key, path string, force bool) bool
+	// newBlockCooldown reduces the initial access time for newly cached blocks.
+	// This helps prevent a surge of writes from evicting active read blocks.
+	newBlockCooldown time.Duration
 
 	opTs map[time.Duration]func() error
 	opMu sync.Mutex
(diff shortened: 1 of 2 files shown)


polyrabbit:
FYI - RocksDB also divides its LRU into two tiers to prevent heavy writes from evicting hot data.
https://github.com/facebook/rocksdb/blob/04c085a3faab360747615fb9975495bfa5a3360a/cache/lru_cache.cc#L281

Same idea here, it should be applied to all write parts, including flush and stage.

pkg/chunk/disk_cache.go
```diff
@@ -736,9 +740,7 @@ func (cache *cacheStore) add(key string, size int32, atime uint32) {
 			cache.used -= int64(iter.size + 4096)
 		}
 		iter.size = size
-		if atime > iter.atime {
-			iter.atime = atime
-		}
+		// atime should only be updated on `load`
```
jiefenghuang, line 743: 这里保留？scan的时候需要，其他地方也不影响

jiefenghuang:
> FYI - RocksDB also divides its LRU into two tiers to prevent heavy writes from evicting hot data. https://github.com/facebook/rocksdb/blob/04c085a3faab360747615fb9975495bfa5a3360a/cache/lru_cache.cc#L281
> 
> Same idea here, it should be applied to all write parts, including flush and stage.

写缓存没问题。不过flush也包含warmup之类的缓存，现在淘汰优先级也变高了

polyrabbit:
Closing as we have different priorities.

Issue #6408, opened 2025-10-16
`--update-fstab --attr-cache=1` creates wrong fstab line — option attr-cache requires a value


kuba-orlik:
**What happened**:

I mounted juicefs with

```
juicefs mount --update-fstab --attr-cache=1 --entry-cache=1 --dir-entry-cache=1 --negative-entry-cache=1 --buffer-size=1000 --prefetch=2 --writeback sqlite3:///var/lib/juicefs/metadata.db /mnt/juice
```

After reboot, the mount point was not available. The line in fstab was:

```
sqlite3:///var/lib/juicefs/metadata.db  /mnt/juice  juicefs  _netdev,attr-cache,buffer-size=1000,dir-entry-cache,entry-cache,negative-entry-cache,prefetch=2,writeback  0 0
```

And systemd reports an error:

```
juicefs[873] <FATAL>: option attr-cache requires a value [main@main.go:31]
mnt-juice.mount: Mount process exited, code=exited, status=1/FAILURE
mnt-juice.mount: Failed with result 'exit-code'.
Failed to mount mnt-juice.mount - /mnt/juice.
```

**What you expected to happen**:

I expected the mountpoint to be mounted upon reboot

**How to reproduce it (as minimally and precisely as possible)**:

See steps in the  "what happened" section

**Anything else we need to know?**

**Environment**:
- JuiceFS version (use `juicefs --version`) or Hadoop Java SDK version: juicefs version 1.3.0+2025-07-03.30190ca
- Cloud provider or hardware configuration running JuiceFS: Hetzner VPS
- OS (e.g `cat /etc/os-release`): Debian 13
- Kernel (e.g. `uname -a`): Llinux 6.12.48
- Object storage (cloud provider and region, or self maintained): I think it's not relevant here — but I can provide it if necessary
- Metadata engine info (version, cloud provider managed or self maintained): sqlite3
- Network connectivity (JuiceFS to metadata engine, JuiceFS to object storage): local sqlite3 to metadata, fiber connection to s3 provider

PR #6412, opened 2025-10-17
cmd/mount: preserve parameter values in fstab options

Xuyuchao-juice:
Fix attr-cache=1 becoming attr-cache in fstab entries

rel https://github.com/juicedata/juicefs/issues/6408

### DIFF
--- a/cmd/mount_unix.go
+++ b/cmd/mount_unix.go
@@ -673,10 +673,8 @@ func tellFstabOptions(c *cli.Context) string {
 		if paramName == "o" {
 			opts = append(opts, c.String(paramName))
 		} else if strings.Contains(s, "=") {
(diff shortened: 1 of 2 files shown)


cmd/mount_unix.go
```diff
@@ -666,25 +666,27 @@ func tellFstabOptions(c *cli.Context) string {
 			continue
 		}
 		s = strings.TrimLeft(s, "-")
-		s = strings.Split(s, "=")[0]
-		if !c.IsSet(s) || s == "update-fstab" || s == "background" || s == "d" {
+		paramName := strings.Split(s, "=")[0]
```
jiefenghuang, line 669: 问题应该是c.Bool()会把其他flag值为"1", "true" "false" "0"误判
这么写bool变量在fstab中也不正确了？

PR #6424, opened 2025-10-24
object/azure: add managed identity authentication

### LINKED ISSUE #6429: Support for Azure Managed Identity Authentication

Currently, the supported methods of authentication are:

1. **Connection String Authentication**
   - Environment variable: `AZURE_STORAGE_CONNECTION_STRING`
   - Format: `"DefaultEndpointsProtocol=https;AccountName=XXX;AccountKey=XXX;EndpointSuffix=core.windows.net"`
   - Uses: `azblob.NewClientFromConnectionString(connString, nil)`

2. **Shared Key Authentication**
   - Parameters: `--access-key` (account name), `--secret-key` (account key)
   - Uses: `azblob.NewSharedKeyCredential(accountName, accountKey)`
   - Then: `azblob.NewClientWithSharedKeyCredential(url, credential, nil)`

I would like to add support for a third authentication method using `azidentity.DefaultAzureCredential` to [newWasb](https://github.com/juicedata/juicefs/blob/f318fce5a877968ad7fb3aada91fe399087e478c/pkg/object/azure.go#L209). Something like:

```go
func newWasb(endpoint, accountName, accountKey, token string) (ObjectStorage, error) {
    // ... endpoint parsing ...
    
    // 1. CONNECTION STRING (existing - highest priority)
    if connString := os.Getenv("AZURE_STORAGE_CONNECTION_STRING"); connString != "" {
        // ... existing logic
        return &wasb{...}, nil
    }

    // 2. SHARED KEY (existing - when both provided)
    if accountName != "" && accountKey != "" {
        // ... existing shared key logic
        return &wasb{...}, nil
    }

    // 3. AZURE AD AUTHENTICATION (new - when accountName provided but no accountKey)
    if accountName != "" && accountKey == "" {
        credential, err := azidentity.NewDefaultAzureCredential(nil)
        if err != nil {
            return nil, fmt.Errorf("failed to create Azure credential: %v", err)
        }
        
        client, err := azblob.NewClient(
            fmt.Sprintf("%s://%s.%s", uri.Scheme, accountName, domain), 
            credential, nil
        )
        if err != nil {
            return nil, err
        }
        return &wasb{...}, nil
    }

    // 4. ERROR (when accountName is empty and no connection string)
    return nil, fmt.Errorf("accountName is required for Azure authentication")
}
```

**Why is this needed**:

Primarily to enable the use of workload identity in a kubernetes context, i.e. the process of obtaining credentials via metadata server, but potentially also to enable other common forms of authentication. This eliminates the need to explicitly manage storage account keys in configuration.

mt5225:
This commit implements support for Azure managed identity authentication, allowing JuiceFS to work seamlessly on Azure infrastructure (VMs, AKS, App Service, etc.) [without requiring explicit credentials](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-quickstart-blobs-go?toc=%2Fazure%2Fdeveloer%2Fgo%2Ftoc.json&bc=%2Fazure%2Fdeveloper%2Fbreadcrumb%2Ftoc.json&tabs=roles-azure-portal#authenticate-to-azure-and-authorize-access-to-blob-data).

Changes:
- Added azidentity package import for token-based authentication
- Implemented createAzureCredential() helper function that supports:
  * System-assigned managed identity (default)
  * User-assigned managed identity (via AZURE_CLIENT_ID env var)
  * DefaultAzureCredential fallback chain (MI → Azure CLI → etc.)
- Modified newWasb() to detect when no account key is provided and attempt token-based authentication before falling back to shared key
- Updated wasb struct to track credential type (sharedKeyCred field)
- Fixed Copy() method to work with both shared key and managed identity:
  * Uses SAS tokens when shared key is available
  * Uses direct blob URLs with token credentials for managed identity
- Simplified endpoint resolution by defaulting to blob.core.windows.net
- Added debug logging throughout authentication flow

Authentication priority:
1. Connection string (AZURE_STORAGE_CONNECTION_STRING)
2. Managed identity / token-based (when accountKey is empty)
3. Shared key (accountName + accountKey)

Backward compatibility: All existing authentication methods continue to work as before. This is purely additive functionality.

Check attachment for test report
[Azure_Managed_Identity_Test_Report.pdf](https://github.com/user-attachments/files/23112376/Azure_Managed_Identity_Test_Report.pdf)

close #6429

### DIFF
--- a/pkg/object/azure.go
+++ b/pkg/object/azure.go
@@ -42,10 +42,11 @@ import (
 
 type wasb struct {
 	DefaultObjectStorage
-	container *container.Client
-	azblobCli *azblob.Client
-	sc        string
-	cName     string
+	container    *container.Client
+	azblobCli    *azblob.Client
+	sc           string
+	cName        string
+	useTokenAuth bool // true when using managed identity/token-based auth, false for shared key/connection string
 }
 
 func (b *wasb) String() string {
@@ -118,11 +119,25 @@ func (b *wasb) Copy(ctx context.Context, dst, src string) error {
 	if b.sc != "" {
 		options.Tier = str2Tier(b.sc)
 	}
-	srcSASUrl, err := srcCli.GetSASURL(sas.BlobPermissions{Read: true}, time.Now().Add(10*time.Second), nil)
-	if err != nil {
-		return err
+
+	var srcURL string
+	var err error
+
+	if b.useTokenAuth {
+		// Token-based authentication: use direct blob URL
+		// Azure will authenticate using the OAuth token from the credential chain
+		srcURL = srcCli.URL()
+		logger.Debugf("Using token-based authentication for Copy operation (direct URL without SAS)")
+	} else {
+		// Shared key authentication: generate SAS token for source blob
+		srcURL, err = srcCli.GetSASURL(sas.BlobPermissions{Read: true}, time.Now().Add(10*time.Second), nil)
+		if err != nil {
+			return err
+		}
+		logger.Debugf("Using shared key authentication for Copy operation (SAS URL)")
 	}
-	_, err = dstCli.CopyFromURL(ctx, srcSASUrl, options)
+
+	_, err = dstCli.CopyFromURL(ctx, srcURL, options)
 	return err
 }
 
@@ -269,7 +284,7 @@ func newWasb(endpoint, accountName, accountKey, token string) (ObjectStorage, er
 		if client, err = azblob.NewClientFromConnectionString(connString, nil); err != nil {
 			return nil, err
 		}
-		return &wasb{container: client.ServiceClient().NewContainerClient(containerName), azblobCli: client, cName: containerName}, nil
+		return &wasb{container: client.ServiceClient().NewContainerClient(containerName), azblobCli: client, cName: containerName, useTokenAuth: false}, nil
 	}
 
 	// Priority 2: Try managed identity / token-based authentication if no account key provided
@@ -296,7 +311,7 @@ func newWasb(endpoint, accountName, accountKey, token string) (ObjectStorage, er
 			return nil, fmt.Errorf("Failed to create Azure blob client with token credential: %v", err)
 		}
 		logger.Debugf("Successfully authenticated using token-based credential")
-		return &wasb{container: client.ServiceClient().NewContainerClient(containerName), azblobCli: client, cName: containerName}, nil
+		return &wasb{container: client.ServiceClient().NewContainerClient(containerName), azblobCli: client, cName: containerName, useTokenAuth: true}, nil
 	}
 
 	// Priority 3: Shared key authentication (existing behavior)
@@ -320,7 +335,7 @@ func newWasb(endpoint, accountName, accountKey, token string) (ObjectStorage, er
 	if err != nil {
 		return nil, err
 	}
-	return &wasb{container: client.ServiceClient().NewContainerClient(containerName), azblobCli: client, cName: containerName}, nil
+	return &wasb{container: client.ServiceClient().NewContainerClient(containerName), azblobCli: client, cName: containerName, useTokenAuth: false}, nil
 }
 
 func init() {


jiefenghuang:
@mt5225 build failed.
need "go mod tidy" and check build again

pkg/object/azure.go
```diff
@@ -216,19 +258,74 @@ func newWasb(endpoint, accountName, accountKey, token string) (ObjectStorage, er
...
 	if connString := os.Getenv("AZURE_STORAGE_CONNECTION_STRING"); connString != "" {
+		logger.Debugf("Using Azure connection string authentication")
 		var client *azblob.Client
 		if client, err = azblob.NewClientFromConnectionString(connString, nil); err != nil {
 			return nil, err
 		}
-		return &wasb{container: client.ServiceClient().NewContainerClient(containerName), azblobCli: client, cName: containerName}, nil
+
+		// Parse connection string to extract credentials for SAS token generation in Copy operations
+		var sharedKeyCred *azblob.SharedKeyCredential
```
jiefenghuang: The sharedKeyCred variable is not actually used — it’s just for a check in Copy.
So for initialization we only need a boolean variable to judge whether it’s a SharedKeyCredential, no need the actual credential value, right?

pkg/object/azure.go
```diff
@@ -117,11 +119,22 @@ func (b *wasb) Copy(ctx context.Context, dst, src string) error {
...
+		// Use SAS token for shared key authentication
+		srcSASUrl, err := srcCli.GetSASURL(sas.BlobPermissions{Read: true}, time.Now().Add(10*time.Second), nil)
+		if err != nil {
+			return err
+		}
+		srcURL = srcSASUrl
+	} else {
+		// For managed identity, use the direct blob URL
+		// The copy operation will use the same credential that authenticated the client
+		srcURL = srcCli.URL()
```
zhijian-pro: It's best not to change the original logic.

pkg/object/azure.go
```diff
@@ -117,11 +119,22 @@ func (b *wasb) Copy(ctx context.Context, dst, src string) error {
...
 		options.Tier = str2Tier(b.sc)
 	}
-	srcSASUrl, err := srcCli.GetSASURL(sas.BlobPermissions{Read: true}, time.Now().Add(10*time.Second), nil)
-	if err != nil {
-		return err
+
+	var srcURL string
+	if b.sharedKeyCred != nil {
+		// Use SAS token for shared key authentication
+		srcSASUrl, err := srcCli.GetSASURL(sas.BlobPermissions{Read: true}, time.Now().Add(10*time.Second), nil)
```
zhijian-pro: This method is the one used before, so what's the difference between this change and the previous one

eakman-datadog: According to the documentation of `GetSASURL`:

> GetSASURL is a convenience method for generating a SAS token for the currently pointed at blob. **It can only be used if the credential supplied during creation was a SharedKeyCredential.**

So seems like the two auth methods need to be treated differently in the context of a server-side copy. Is that right @mt5225 ?

pkg/object/azure.go
```diff
@@ -216,19 +258,74 @@ func newWasb(endpoint, accountName, accountKey, token string) (ObjectStorage, er
...
+		var domain string
+		if len(hostParts) > 1 {
+			domain = hostParts[1]
+			if !strings.HasPrefix(hostParts[1], "blob") {
+				domain = fmt.Sprintf("blob.%s", hostParts[1])
+			}
+		} else {
+			// Default to public Azure cloud when using managed identity
+			domain = "blob.core.windows.net"
+		}
```
eakman-datadog: What's the thinking behind this default? Should we mimic what we do with `autoWasbEnpoint` for consistency? Curious to hear your thoughts.

pkg/object/azure.go
```diff
@@ -180,6 +193,35 @@ func (b *wasb) SetStorageClass(sc string) error {
...
+
+	// Use DefaultAzureCredential for flexible authentication
+	// This tries: Environment -> Workload Identity -> Managed Identity -> Azure CLI -> Azure Developer CLI
+	logger.Debugf("Attempting authentication with DefaultAzureCredential (managed identity, Azure CLI, etc.)")
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
+	if err != nil {
+		logger.Debugf("Failed to create DefaultAzureCredential: %v", err)
+		return nil, err
+	}
+	return cred, nil
```
eakman-datadog: I'm not an expert here, but my gut says we should just use `azidentity.NewDefaultAzureCredential` since this handles the managed identity use case already. FWIW, for my use case (https://github.com/juicedata/juicefs/issues/6429), I used the term _managed identity_, but actually _workload identity_ is really what I meant--sorry for the lack of precision there.

Also if a user wanted to provide credentials via environment variables  (e.g.`AZURE_TENANT_ID`, `AZURE_CLIENT_ID`, `AZURE_CLIENT_SECRET`) this would not work if I'm understanding correctly since the presence of `AZURE_CLIENT_ID` is interpreted as meaning the user intends to use managed identity.
  
  Any reason not to just use `NewDefaultAzureCredential`? Is there a specific use case you had in mind?

eakman-datadog (commented):
Looks pretty good to me for the most part. Added a few comments/questions. Might not be a bad idea to add a test like [this one](https://github.com/juicedata/juicefs/blob/a1b0b093b89e24444fe85b85809e02f61647f84b/pkg/object/object_storage_test.go#L764) for integration testing i.e.:

```go
// existing test
func TestAzure(t *testing.T) { //skip mutate
	if os.Getenv("AZURE_STORAGE_ACCOUNT") == "" {
		t.SkipNow()
	}
	//https://containersName.core.windows.net
	abs, _ := newWasb(os.Getenv("AZURE_ENDPOINT"),
		os.Getenv("AZURE_STORAGE_ACCOUNT"), os.Getenv("AZURE_STORAGE_KEY"), "")
	testStorage(t, abs)
}

// new test
func TestAzureManagedIdentity(t *testing.T) { //skip mutate
	if os.Getenv("AZURE_CLIENT_ID") == "" {
		t.SkipNow()
	}
	//https://containersName.core.windows.net
	abs, _ := newWasb(os.Getenv("AZURE_ENDPOINT"), "", "", "")
	testStorage(t, abs)
}

// etc...
```

eakman-datadog (approved):
🚀 🚀

eakman-datadog (commented):
Actually, one more question. Does `Copy()` still work after your latest change with managed identity?

mt5225:
you are right, current Copy() use [SAS](https://github.com/juicedata/juicefs/blob/main/pkg/object/azure.go#L120),  according to the [doc](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob#Client.GetSASURL) , it only works with SharedKeyCredential.

but i would like to minimize changes to original code in the PR, and Copy() is a base function(?), may create a new PR with proper testing to fix the `Copy()`

eakman-datadog:
> you are right, current Copy() use [SAS](https://github.com/juicedata/juicefs/blob/main/pkg/object/azure.go#L120), according to the [doc](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob#Client.GetSASURL) , it only works with SharedKeyCredential.
> 
> but i would like to minimize changes to original code in the PR, and Copy() is a base function(?), may create a new PR with proper testing to fix the `Copy()`

Ok, I'll follow your lead. FWIW, I think what you had before wasn't bad--I would just switch to using a boolean rather than storing the shared key credential on the `wasb` struct as @jiefenghuang [pointed out](https://github.com/juicedata/juicefs/pull/6424#discussion_r2468113252).

mt5225:
will continue this PR to update Copy() to work with the identity-base auth

mt5225:
updated Copy() attached test result
[testresults_summary.pdf](https://github.com/user-attachments/files/23222991/testresults_summary.pdf)

jiefenghuang (approved):

PR #6431, opened 2025-10-29
cmd/sync: refactor some check

jiefenghuang:
1. "obj.IsSymlink() && config.Links" not works
2. produceSingleObject should return nil when object is dir and config.Dirs == false (There are currently only two places that call it, so this return value doesn’t matter.)

### DIFF
--- a/pkg/sync/config.go
+++ b/pkg/sync/config.go
@@ -225,12 +225,6 @@ func NewConfigFromCli(c *cli.Context) *Config {
 		logger.Warnf("threads should be larger than 0, reset it to 1")
 		cfg.Threads = 1
 	}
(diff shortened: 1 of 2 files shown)


pkg/sync/config.go
```diff
@@ -225,6 +225,12 @@ func NewConfigFromCli(c *cli.Context) *Config {
 		logger.Warnf("threads should be larger than 0, reset it to 1")
 		cfg.Threads = 1
 	}
+	if len(cfg.FilesFrom) > 0 {
+		if !cfg.Dirs {
```
zhijian-pro: 这里会影响后续的子目录的行为

PR #6435, opened 2025-10-31
cache/stage: add cooldown time for cache

jiefenghuang:
take over #6391
thanks to @polyrabbit


polyrabbit:
But a periodic scan will override those biased atimes.

Xuyuchao-juice (approved):

PR #6437, opened 2025-10-31
code: remove unused config arg

201341:


pkg/meta/sql.go
```diff
@@ -5097,12 +5106,18 @@ func (m *dbMeta) getDirFetcher() dirFetcher {
...
 			return nil
 		})
+
 		if err != nil {
 			return nil, nil, err
 		}
 		if len(entries) == 0 {
 			return nil, nil, nil
 		}
+		if m.conf.SortDir {
```
jiefenghuang: The SQL query already orders by name at the front, don’t need it here.

pkg/meta/redis.go
```diff
@@ -2205,6 +2205,11 @@ func (m *redisMeta) doReaddir(ctx Context, inode Ino, plus uint8, entries *[]*En
 	if err != nil {
 		return errno(err)
 	}
+	if m.conf.SortDir {
```
jiefenghuang: Why this change is needed? FUSE isn’t involved here — other internal functions currently don’t need to be ordered. If some other SDK needs ordering, it should be done in the SDK. Let’s move this to another PR.

201341: ok

PR #6442, opened 2025-11-03
fs: return sort entries when sort-dir is true

201341:


pkg/fs/fs.go
```diff
@@ -1398,6 +1399,11 @@ func (f *File) Readdir(ctx meta.Context, count int) (fi []os.FileInfo, err sysca
 		if err != 0 {
 			return
 		}
+		if f.fs.conf.Meta.SortDir {
```
jiefenghuang, line 1402: skip . and .. first

201341: `sort.Slice(inodes[2:], func(i, j int) bool {`  the code has skip . and ..

jiefenghuang: check line 1408

jiefenghuang (approved):

PR #6451, opened 2025-11-07
Implement delete option in fsck command

### LINKED ISSUE #6441: [cli] fsck --path not working as expected
cli version: 1.3.0

`fsck <metaurl>`

`fsck <metaurl> --path <path> -r`

I found that fsck with path or without path works differently. When without `--path` fsck check entire my storage and checking missing slices, but if I know which file is problematic and use `--path` to check it directly, it does not really do the job.

Is there any reason that `--path` cannot checking the missing slices? I certainly don't want to fsck entire storage every time when I just want to check a few directories.

marios1861:
Added delete flag to fsck command to remove lost files in the object store

Partially fixes #6441

### DIFF
--- a/cmd/fsck.go
+++ b/cmd/fsck.go
@@ -65,6 +65,10 @@ $ juicefs fsck redis://localhost --path /d1/d2 --recursive`,
 				Name:  "sync-dir-stat",
 				Usage: "sync stat of all directories, even if they are existed and not broken (NOTE: it may take a long time for huge trees)",
 			},
+			&cli.BoolFlag{
+				Name:  "delete",
+				Usage: "delete lost files",
+			},
 		},
 	}
 }
@@ -205,7 +209,17 @@ func fsck(ctx *cli.Context) error {
 		}
 		sort.Strings(fileList)
 		msg += strings.Join(fileList, "\n")
-		logger.Fatal(msg)
+		if ctx.Bool("delete") {
+			logger.Infof("Cleaning up broken files")
+			for i, p := range brokens {
+				if strings.HasPrefix(p, "inode:") {
+					m.DeleteFileData(i, 0, true)
+					logger.Infof("Deleted data of inode %d", i)
+				}
+			}
+		} else {
+			logger.Fatal(msg)
+		}
 	}
 
 	return nil
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -2479,6 +2479,10 @@ func (m *baseMeta) fileDeleted(opened, force bool, inode Ino, length uint64) {
 	}
 }
 
+func (m *baseMeta) DeleteFileData(inode Ino, length uint64, force bool) {
+	m.en.doDeleteFileData(inode, length)
+}
+
 func (m *baseMeta) tryDeleteFileData(inode Ino, length uint64, force bool) {
 	if force {
 		m.maxDeleting <- struct{}{}
--- a/pkg/meta/interface.go
+++ b/pkg/meta/interface.go
@@ -507,6 +507,7 @@ type Meta interface {
 	// Get a copy of the current format
 	GetFormat() Format
 
+	DeleteFileData(inode Ino, length uint64, force bool)
 	// OnMsg add a callback for the given message type.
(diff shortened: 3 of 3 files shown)


jiefenghuang:
Hi @marios1861, the log had include all lost files, you can delete them directly.

marios1861:
Hi @jiefenghuang. By deleting the lost files listed in the output of `juicefs fsck`, purging trash, and compacting with `juicefs gc --compact` `juicefs gc --delete`, the problem remains. It seems that by removings the files, juicefs still tries to find the missing blocks that don't actually exist in the object store, and that don't point to any file any more.

```
2025/11/07 12:49:02.936363 juicefs[368718] <WARNING>: Get directory path of 9223372032828244098: getattr inode 9223372032828244098: no such file or directory [GetPaths@base.go:1899]
2025/11/07 12:49:02.936429 juicefs[368718] <ERROR>: can't find block 1/1773/1773120_0_1466794 for file inode:1504122: file does not exist [fsck@fsck.go:187]
2025/11/07 12:49:03.054603 juicefs[368718] <WARNING>: Get directory path of 9223372032828244096: getattr inode 9223372032828244096: no such file or directory [GetPaths@base.go:1899]
2025/11/07 12:49:03.054678 juicefs[368718] <ERROR>: can't find block 1/1610/1610383_1_4194304 for file inode:8471: file does not exist [fsck@fsck.go:187]
2025/11/07 12:49:03.070134 juicefs[368718] <ERROR>: can't find block 1/1610/1610383_7_4194304 for file inode:8471: file does not exist [fsck@fsck.go:187]
2025/11/07 12:49:03.081944 juicefs[368718] <ERROR>: can't find block 1/1610/1610383_12_4194304 for file inode:8471: file does not exist [fsck@fsck.go:187]
2025/11/07 12:49:03.089229 juicefs[368718] <ERROR>: can't find block 1/1610/1610383_15_4194304 for file inode:8471: file does not exist [fsck@fsck.go:187]
2025/11/07 12:49:03.104766 juicefs[368718] <ERROR>: can't find block 1/1610/1610384_6_4194304 for file inode:8471: file does not exist [fsck@fsck.go:187]
2025/11/07 12:49:03.115867 juicefs[368718] <ERROR>: can't find block 1/1610/1610384_12_4194304 for file inode:8471: file does not exist [fsck@fsck.go:187]
2025/11/07 12:49:03.128584 juicefs[368718] <ERROR>: can't find block 1/1610/1610385_4_4194304 for file inode:8471: file does not exist [fsck@fsck.go:187]
2025/11/07 12:49:03.150288 juicefs[368718] <ERROR>: can't find block 1/1610/1610386_1_4194304 for file inode:8471: file does not exist [fsck@fsck.go:187]
2025/11/07 12:49:03.167168 juicefs[368718] <ERROR>: can't find block 1/1610/1610386_9_4194304 for file inode:8471: file does not exist [fsck@fsck.go:187]
```

These are all detached inodes i'm assuming, but `juicefs gc` isn't removing them.

jiefenghuang:
1. The issue you reported initially — that a block in object storage was lost, and you need to clear the corresponding file in the filesystem — you can simply delete the file from the filesystem and wait for the background process to automatically reclaim it. That does not require this PR.
2. The log you post above is another issue — it appears to be an abnormal trash directory deletion. This should not happen. We should identify the root cause of this problem, rather than introducing this PR. Are you able to reproduce it? Please add information back into the original issue description. Thank you.

marios1861:
@jiefenghuang I was struggling the past days to remove those ghost inodes and the `deleteSlice` wasn't helping. The issue was that the chunk metadata remained, so even if slices and inodes were removed from metadata, since (in the case of sql engine) `listSlices` uses the chunk metadata, the false inodes were showing up. So I used the `doDeleteFileData` and it worked, but only because in the case of sql the `length` argument isn't actually used at all. I don't think this is a useful pr so I am closing it, but it seems to me that the background cleanup task tried to remove the slices from the object store, and because the object store was not functioning properly, it returned response 500. The cleanup task incorrectly removed these slices from the pending slices for some odd reason. I couldn't do much else other than manual fix this specific issue for me.

PR #6469, opened 2025-11-14
meta: Replace doReaddir with DirHandler for batch fetching in cloneEntry

### LINKED ISSUE #4806: clone on a big dir is slow and progress display incorrect.
**What happened**:
1. the meta op is too slow.
2. ETA in progress is a minus number.
3. at the beginning of clone, the progress is 0 for a long time.

image

image

**Environment**:
- Kernel (e.g. `uname -a`):
- Others:

Xuyuchao-juice:
close #4806
Why this change:

The original implementation used doReaddir to load all directory entries at once, which causes performance issues and high memory pressure when cloning large directories. This leads to slow metadata operations, negative ETA values, and progress staying at 0 for extended periods.

How it's implemented:

Replace the single doReaddir call with DirHandler's List method to fetch directory entries in batches. The implementation maintains the directory-first processing strategy (directories before files) to maximize parallelism, while using incremental offset-based fetching to reduce per-query overhead and improve overall performance for large directories.

Test result:
image

### DIFF
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -2929,70 +2929,26 @@ func (m *baseMeta) cloneEntry(ctx Context, srcIno Ino, parent Ino, name string, 
 		return eno
 	}
 
-	// Process entries in batches: directories first, then files
 	offset := 0
-	var dirEntries []*Entry
-	var fileEntries []*Entry
 LOOP:
 	for {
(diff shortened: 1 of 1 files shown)


jiefenghuang:
Need to fix unit tests.

jiefenghuang (approved):

PR #6473, opened 2025-11-18
vfs: merge getattr and lookup requests

### LINKED ISSUE #6322: Consider adding single flight optimization for some simple requests?
When multiple processes access the same mount point, they may issue the same requests for the same inode/entry simultaneously, such as `getattr`/`lookup`:

Image

Image

This situation is more likely to occur in a Kubernetes runtime environment, as identical images generate identical containers. These simultaneous requests cannot be intercepted by a cache, but with a single flight optimization in the daemon, the pressure to meta engine in the above scenario could be reduced by 90%.

jiefenghuang:
close #6322

### DIFF
--- a/pkg/vfs/handle.go
+++ b/pkg/vfs/handle.go
@@ -19,6 +19,7 @@ package vfs
 import (
 	"encoding/hex"
 	"encoding/json"
+	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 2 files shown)


SandyXSD:
IMHO the scenario for this optimization is very limited and unnecessary.

Issue #4599, opened 2024-03-28
Expose trash usage metrics


polyrabbit:
**What would you like to be added**:
Just like `used_space` and `used_inodes`, it will be very helpful to spot capacity issues when we have similar metrics on trash.

Also it will be great if we have metrics on `DelayedSlices` 😉.

**Why is this needed**:
To spot capacity issues.

jiefenghuang:
Currently, after DirStat is enabled, the trash directory is also counted in real time and can be viewed through the `summary` without requiring a traversal.

PR #6482, opened 2025-11-21
meta/metrics: Add Prometheus metrics for trash and delayed slices

Xuyuchao-juice:
rel https://github.com/juicedata/juicefs/issues/4599

Test result：
image

### DIFF
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -255,6 +255,12 @@ type baseMeta struct {
 	fsStatsLock sync.Mutex
 	*fsStat
 
+	trashMetricsLock      sync.Mutex
+	trashSpace            int64
+	trashInodes           uint64
+	delayedSlicesSpace    uint64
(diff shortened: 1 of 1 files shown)


pkg/meta/base.go
```diff
@@ -375,6 +398,8 @@ func (m *baseMeta) InitMetrics(reg prometheus.Registerer) {
 				m.totalSpaceG.Set(float64(totalSpace))
 				m.totalInodesG.Set(float64(iused + iavail))
 			}
+			m.updateTrashMetrics(Background())
```
zhijian-pro: The cost of scanning every 10 seconds is too high. It is necessary to internally record a total and continuously update it, feeding back this indicator every 10 seconds

jiefenghuang:
Global scanning is too costly — we should instead record the delta. Let’s close this for now.

PR #6493, opened 2025-11-26
fix: correct object name assignment in getObjectInfoNoFSLock

zhijian-pro:


pkg/gateway/gateway.go
```diff
@@ -1488,7 +1488,7 @@ func (n *jfsObjects) getObjectInfoNoFSLock(ctx context.Context, bucket, object s
 		fi := info.(*fs.FileStat)
 		oi = minio.ObjectInfo{
 			Bucket:   bucket,
-			Name:     fi.Name(),
```
zhitaoli-6, line 1491: Hi, @zhijian-pro  is there any difference between FileStat.Name() and object name in getObjectInfo function? 

I find that Entry.Name is filled with  FileStat.Name() in listDirFactory() function(juicefs pkg/gateway/gateway.go) and then used as object name in getObjectInfo function(minio cmd/object-api-common.go). 

Please tell me if I misunderstand it.

zhitaoli-6: cc @Xuyuchao-juice

PR #6495, opened 2025-11-26
meta/redis: support client cache

### LINKED ISSUE #97: Enable Redis client cache

jiefenghuang:
take over #6272
close #97
thanks to @valandreev

### DIFF
--- a/pkg/meta/redis_csc.go
+++ b/pkg/meta/redis_csc.go
@@ -269,7 +269,9 @@ func (c *redisCache) close() {
 		}
 		c.subscription = nil
 	}
-	c.cli.Options().OnConnect = nil
+	if c.cli != nil {
+		c.cli.Options().OnConnect = nil
+	}
 	c.cli = nil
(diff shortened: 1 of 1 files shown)


pkg/meta/redis_csc.go
```diff
@@ -0,0 +1,327 @@
...
+		return keyTypInode
+	}
+	if strings.HasPrefix(key, c.prefix+"d") {
+		return keyTypEntry
+	}
+	return keyTypOther
+}
+
+func (c *redisCache) entryName(parent Ino, name string) string {
+	return fmt.Sprintf("%d%d%s", parent, os.PathSeparator, name)
```
SandyXSD, line 116:
```suggestion
	return fmt.Sprintf("%d%s%s", parent, os.PathSeparator, name)

```
?

PR #6532, opened 2025-12-05
decode based on the encoding type returned by the server

zhijian-pro:


zhijian-pro:
Not all servers (ex. Huawei OceanStor) support the "EncodingType" parameter, so we should decide whether to decode based on whether the server actually encoded it.

jiefenghuang (approved):

Issue #5745, opened 2025-03-07
cmd/gc: enhance progress display


jiefenghuang:
**What would you like to be added**:
gc progress detail when doing ListSlices

**Why is this needed**:
Currently, in ListSlice, functions like cleanupLeakedInodes take a considerable amount of time to iterate without displaying progress, which may cause confusion for users.

<img width="567" alt="Image" src="https://github.com/user-attachments/assets/8ff77a81-e561-4238-907a-d7040ca46a7e" />

![Image](https://github.com/user-attachments/assets/9ae92017-5c87-4b34-84ae-b444355cd164)

bobqian1007:
seems ListNode function already have showProgress func as the parameter to trace the progress? do you know we also need to add the progress bar in the pre-cleanup stage? Like procedure cleanupLeakedInodes,cleanupLeakedChunks,cleanupOldSliceRefs,doCleanupSlices?

jiefenghuang:
which ListNode? Currently, cleanupLeakedInodes is only added in the Redis engine, and there is no progress display for the current scanning process. I think it’s necessary to show the progress of time-consuming operations during the GC process, otherwise, users won’t know whether it’s working normally.

bobqian1007:
> which ListNode? Currently, cleanupLeakedInodes is only added in the Redis engine, and there is no progress display for the current scanning process. I think it’s necessary to show the progress of time-consuming operations during the GC process, otherwise, users won’t know whether it’s working normally.

I got it. Btw I am new to this project and would like to know what I should do if I want to take this task. Do I need to write a design document first?

jiefenghuang:
For this issue, a simple description is sufficient, but we need to consider the implementation of multiple engines. 

Going further, the GC currently scans two main categories: metadata and data. Within these, there are various scenarios, such as residues after compaction, records of asynchronously deleted files, residues caused by cloning, dangling inodes or edges, etc. It would be helpful to categorize the scanned data and ultimately produce a clearer statistical list. @SandyXSD @davies is this improvement needed?

bobqian1007:
> For this issue, a simple description is sufficient, but we need to consider the implementation of multiple engines.
> 
[...]

PR #6538, opened 2025-12-08
log: add debug log to enhance GC

hegdeadithyak:
image
Solves #5745

If this change goes beyond what was intended, please let me know. This is my first contribution to this repository, so any feedback or guidance is appreciated. @jiefenghuang

### DIFF
--- a/pkg/meta/redis.go
+++ b/pkg/meta/redis.go
@@ -2883,7 +2883,6 @@ func (m *redisMeta) doFindDeletedFiles(ts int64, limit int) (map[Ino]uint64, err
 }
 
 func (m *redisMeta) doCleanupSlices(ctx Context) {
-	logger.Infof("Cleaning up slices with negative reference counts...")
 	_ = m.hscan(ctx, m.sliceRefs(), func(keys []string) error {
 		for i := 0; i < len(keys); i += 2 {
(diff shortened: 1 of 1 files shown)


pkg/meta/redis.go
```diff
@@ -3349,12 +3356,18 @@ func (m *redisMeta) hscan(ctx context.Context, key string, f func([]string) erro
 }
 
 func (m *redisMeta) ListSlices(ctx Context, slices map[Ino][]Slice, scanPending, delete bool, showProgress func()) syscall.Errno {
+	logger.Infof("GC: Starting pre-cleanup operations (leaked inodes, chunks, and slice references)...")
```
jiefenghuang: This should be enough, no other modifications are required.
```go
	logger.Debugf("start cleanup...")
	m.cleanupLeakedInodes(delete)
	m.cleanupLeakedChunks(delete)
	m.cleanupOldSliceRefs(delete)
	if delete {
		m.doCleanupSlices(ctx)
	}
	logger.Debugf("start listing slices...")
```

hegdeadithyak: Got it — I suspected I might have overdone it. Thanks for pointing it out, I’ve refined the changes now. @jiefenghuang

PR #6545, opened 2025-12-16
vfs: fix append write offset

jiefenghuang:
When alternately appending writes, file1 and file2 are the same file on two different mount points, and data may be lost.
```shell
for ((i=1; i<=COUNT; i++)); do
    echo "$FILE1:$i" >> "$FILE1"    
    echo "$FILE2:$i" >> "$FILE2"
done
```

### DIFF
--- a/pkg/vfs/vfs.go
+++ b/pkg/vfs/vfs.go
@@ -848,6 +848,10 @@ func (v *VFS) Write(ctx Context, ino Ino, buf []byte, off, fh uint64) (err sysca
 	}
 	defer h.Wunlock()
 
+	// kernel might return a stale length from attr-cache before open as the offset.
+	if h.flags&uint32(os.O_APPEND) != 0 {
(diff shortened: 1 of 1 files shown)


pkg/vfs/vfs.go
```diff
@@ -848,6 +848,10 @@ func (v *VFS) Write(ctx Context, ino Ino, buf []byte, off, fh uint64) (err sysca
 	}
 	defer h.Wunlock()
 
+	// kernel might return a stale length from attr-cache before open as the offset.
+	if h.flags&uint32(os.O_APPEND) != 0 {
+		off = h.writer.GetLength()
```
davies, line 853: Should we do this in meta transaction?

jiefenghuang: 我们只要保证CTO，拿到Open之后的attr length就可以，handler的length是Open之后更新的了

PR #6548, opened 2025-12-19
utils: add utils.GetFileInodeNotFollow()

chenjie4255:
* added utils.GetFileInodeNotFollow() methods

### DIFF
--- a/pkg/utils/utils_windows.go
+++ b/pkg/utils/utils_windows.go
@@ -24,12 +24,16 @@ import (
 	"golang.org/x/sys/windows"
 )
 
-func GetFileInode(path string) (uint64, error) {
+func getFileInode(path string, follow bool) (uint64, error) {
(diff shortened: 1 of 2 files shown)


pkg/utils/utils_windows.go
```diff
@@ -42,6 +46,14 @@ func GetFileInode(path string) (uint64, error) {
 	return uint64(data.FileIndexHigh)<<32 + uint64(data.FileIndexLow), nil
 }
 
+func GetFileInode(path string) (uint64, error) {
+	return getFileInode(path, true)
+}
+
+func GetFileInodeNotFollow(path string) (uint64, error) {
```
zhijian-pro, line 53: Is there no one using this function currently?

chenjie4255: No, not for this project, since this is in a utility package and it will be used by other projects.

PR #6560, opened 2025-12-25
cmd/gateway: improve the performance of batch deletion of objects

### LINKED ISSUE #5094: Improve the performance of gateway delete objects interface
The delete object interface of the gateway can use the asynchronous batch delete method to improve performance

zhijian-pro:
close #5094

### DIFF
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -297,6 +297,9 @@ type baseMeta struct {
 	opCount      *prometheus.CounterVec
 	opDuration   *prometheus.CounterVec
 
+	// Subdir info metric
+	subdirInfoG *prometheus.GaugeVec
+
 	// Quota metrics
 	dirQuotaMaxSpaceG   *prometheus.GaugeVec
 	dirQuotaMaxInodesG  *prometheus.GaugeVec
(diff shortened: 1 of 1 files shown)


pkg/gateway/gateway.go
```diff
@@ -507,14 +507,21 @@ func (n *jfsObjects) DeleteObject(ctx context.Context, bucket, object string, op
...
-		_, errs[idx] = n.DeleteObject(ctx, bucket, object.ObjectName, options)
-		if errs[idx] == nil {
-			objs[idx] = minio.DeletedObject{
-				ObjectName: object.ObjectName,
+	var g errgroup.Group
+	g.SetLimit(10)
+	for i := 0; i < len(objects); i++ {
+		i := i
+		g.Go(func() error {
+			_, errs[i] = n.DeleteObject(ctx, bucket, objects[i].ObjectName, options)
```
jiefenghuang, line 515: A common case is batch deletion under the same directory（mc rm -r）. Because the metadata has a directory lock, concurrency doesn’t improve much.
We can sort them, use the newly BatchUnlink API for files in the same directory, and run concurrently for different directories.

PR #6561, opened 2025-12-26
cmd/fsck: allow specifying dir permissions for repair

ndeepak-baseten:
Juicefs-fsck repair sets directory permissions to 0755 by default. This can be quite rigid.

If we can make this configurable, it will allow us to repair the directories to our specific permission bits.

In this PR:
* add a cmdline arg to take in an expected repair mode, with 0755 as the default
* use arg value when repairing dir

I have added a unit test, but could not run it locally as it requires MacFUSE, which I don't have.

### DIFF
--- a/pkg/meta/base_test.go
+++ b/pkg/meta/base_test.go
@@ -2714,7 +2714,10 @@ func testCheckAndRepair(t *testing.T, m Meta) {
 
 	showProgress := func(n int) {}
 	slices := make(map[Ino][]Slice)
-	if err := m.Check(Background(), "/check", false, false, false, showProgress, slices); err == nil {
+	if err := m.Check(Background(), "/check", &CheckOpt{
+		ShowProgress: showProgress,
+		Slices:       slices,
+	}); err == nil {
 		t.Fatal("check should fail")
 	}
 	if st := m.GetAttr(Background(), checkInode, dirAttr); st != 0 {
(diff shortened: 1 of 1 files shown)


jiefenghuang:
Hi @ndeepak-baseten, thanks for your feedback. 
However, this permission setting can be modified with chmod after the user confirms the repair results with fsck. I think it’s unnecessary to add an extra parameter for configuration.

ndeepak-baseten:
@jiefenghuang Thank you for your review. We are doing that (post-repair pass) now, but two problems:
* it is not an atomic operation
* it is intensive to run recursive chmod's for large JuiceFS deployments

If a CLI argument is too much, we could use an environment variable as well, perhaps `JUICEFS_FSCK_DIR_PERMS`. That's an alternative.

What do you think?

pkg/meta/interface.go
```diff
@@ -499,7 +499,7 @@ type Meta interface {
 	// GetPaths returns all paths of an inode
 	GetPaths(ctx Context, inode Ino) []string
 	// Check integrity of an absolute path and repair it if asked
-	Check(ctx Context, fpath string, repair bool, recursive bool, statAll bool, showProgress func(n int), slices map[Ino][]Slice) error
+	Check(ctx Context, fpath string, repair bool, recursive bool, statAll bool, repairDirMode uint16, showProgress func(n int), slices map[Ino][]Slice) error
```
jiefenghuang:
```golang
Check(ctx Context, fpath string, repair bool, recursive bool, statAll bool, repairDirMode uint16, showProgress func(n int), slices map[Ino][]Slice) error

// too much options, we can change to this
Check(ctx Context, fpath string, opt *CheckOpt) error

type CheckOpt struct {
    Repair bool
    Recursive bool
    ...
}
```

ndeepak-baseten: @jiefenghuang Thanks! Expanded the PR, but good time to factor this!

jiefenghuang:
Hi @ndeepak-baseten, please sign the cla.

ndeepak-baseten:
> Hi @ndeepak-baseten, please sign the cla.

@jiefenghuang Just done, and thanks for the UT fix!

Issue #6576, opened 2026-01-01
"testing/" directory remains after formatting volume with shards


Light-City:
**What happened**:

When formatting a new volume with --shards > 0 (e.g., --shards=10 --bucket=s3://mybucket/data-%d), the test directory testing/ is not cleaned up from the object storage.

The root cause is that sharded.pick(key) hashes the entire key string. During the format test:
A test object is created with a key like `testing/abcdefg`. It is hashed and routed to Shard A.
After the test, the command attempts to delete the prefix `testing/`. However, the string `testing/` hashes to a different value, routing the request to Shard B.

**What you expected to happen**:

should delete testing dir.

**How to reproduce it (as minimally and precisely as possible)**:

cmd is:
```
./juicefs format redis://127.0.0.1:6379 myjfs --shards=120 --bucket=metadata/juicefs-%d
```

**Anything else we need to know?**

**Environment**:
- JuiceFS version (use `juicefs --version`) or Hadoop Java SDK version: 1.4.0-dev+2026-01-01.f1701da0

jiefenghuang:
The current issue uses sharding, and the backend storage is a filesystem type (this problem doesn’t occur with S3 type). It leaves behind an empty directory (possibly with multiple levels).
When deleting a prefix, we need to traverse all filesystem-type backend stores to delete the prefix directory. The fix in #6577 does not completely solve the problem.

I think we can keep the current state for now since it doesn’t have a major impact. Later we will consider whether it’s necessary to clean up the empty directories.

Light-City:
> The current issue uses sharding, and the backend storage is a filesystem type (this problem doesn’t occur with S3 type). It leaves behind an empty directory (possibly with multiple levels). When deleting a prefix, we need to traverse all filesystem-type backend stores to delete the prefix directory. The fix in [#6577](https://github.com/juicedata/juicefs/pull/6577) does not completely solve the problem.
> 
> I think we can keep the current state for now since it doesn’t have a major impact. Later we will consider whether it’s necessary to clean up the empty directories.

“When deleting a prefix, we need to traverse all filesystem-type backend stores to delete the prefix directory.”  Does this sentence mean that all buckets need to be deleted?
For example, if the "testing" subdirectory exists in bucket1, bucket2, etc., does it mean that the "testing" subdirectory in all buckets should be deleted?

[...]

PR #6577, opened 2026-01-01
cmd/object: fix "testing/" directory remains after formatting volume with shards

Light-City:
related issue: https://github.com/juicedata/juicefs/issues/6576

**Summary:** This PR fixes a bug where test data is left behind in object storage when formatting a volume with sharding enabled.

**Technical Details**: Since sharded storage relies on FNV hashing of the key, a directory prefix like `testing/` and an object within it like` testing/abc` will likely land on different shards. By the time we try to clean up the prefix, the sharded implementation routes the request to the wrong bucket.

I implemented a ShardHint mechanism via context.Context. This allows the caller (in this case, the format command) to "hint" to the sharded storage which shard it should target, based on a previously used key. This approach is non-intrusive to the ObjectStorage interface and is a common pattern in Go for metadata propagatio.

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -299,7 +299,7 @@ func config(ctx *cli.Context) error {
 			if err != nil {
 				return err
 			}
-			if err = test(blob); err != nil {
+			if err = test(blob, format.Name); err != nil {
 				return err
 			}
 		}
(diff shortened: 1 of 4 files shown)


jiefenghuang (commented):
Based on the discussion on 6575, we may close this PR for now.

PR #6582, opened 2026-01-06
meta/tkv: also support cancelling txn when request is interrupted

polyrabbit:
In a busy mount point, fuse requests being canceled (interrupted) happens frequently. When a request is canceled, the corresponding inflight metadata and data requests should also support cancellation. In the current redis engine, txn already supports cancellation. This PR completes the cancellation capability for tkv.

### DIFF
--- a/pkg/meta/redis.go
+++ b/pkg/meta/redis.go
@@ -1131,14 +1131,15 @@ func (m *redisMeta) txn(ctx Context, txf func(tx *redis.Tx) error, keys ...strin
 
 	m.txLock(h)
 	defer m.txUnlock(h)
-	// TODO: enable retry for some of idempodent transactions
+	// TODO: enable retry for some of idempotent transactions
 	var (
-		retryOnFailture = false
-		lastErr         error
-		method          string
+		retryOnFailure = false
+		lastErr        error
+		method         txMethod
 	)
 	for i := 0; i < 50; i++ {
 		if ctx.Canceled() {
+			logger.Warnf("Transaction %s interrupted after %s, tried %d, keys: %v", method.name(ctx), time.Since(start), i+1, keys)
 			return syscall.EINTR
 		}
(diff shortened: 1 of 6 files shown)


pkg/meta/tkv_bak.go
```diff
@@ -297,7 +296,7 @@ func (m *kvMeta) dumpMix(ctx Context, opt *DumpOption, ch chan<- *dumpedResult)
 		}
 		logger.Debugf("range: %v-%v", start, end)
 		eg.Go(func() error {
-			return m.txn(egCtx, func(tx *kvTxn) error {
+			return m.txn(WrapContext(ctx), func(tx *kvTxn) error {
```
jiefenghuang: WrapContext(egCtx)

pkg/meta/redis.go
```diff
@@ -1125,6 +1125,7 @@ func (m *redisMeta) txn(ctx Context, txf func(tx *redis.Tx) error, keys ...strin
 	var khash = fnv.New32()
 	_, _ = khash.Write([]byte(keys[0]))
 	h := uint(khash.Sum32())
+	method := callerName(ctx)
```
jiefenghuang: lazy evaluation ?

polyrabbit: Then these 3 lines of code need to be repeated at lease twice in each `txn` function, feels not so elegant?
```
if method == "" {
	method = callerName(ctx) // lazy evaluation
}
```

pkg/meta/tkv_bak.go
polyrabbit: Right, updated.

pkg/meta/redis.go
jiefenghuang: I think it’s fine.

polyrabbit: OK, updated with a cleaner way

jiefenghuang (approved):

PR #6596, opened 2026-01-12
Revert "vfs: fix append write offset"

jiefenghuang:
Reverts juicedata/juicefs#6545
issue #6594
Under the [writeback_cache] mode, the kernel may first read pages and then complete the write before flushing them. In JuiceFS user-space this appears like an overwrite behavior, so using the file size directly as the offset (as in #6545) is incorrect.

The issue of concurrent appends potentially overwriting each other currently doesn’t have a good solution. It will be documented under the consistency exceptions section.

### DIFF
--- a/pkg/vfs/vfs.go
+++ b/pkg/vfs/vfs.go
@@ -848,10 +848,6 @@ func (v *VFS) Write(ctx Context, ino Ino, buf []byte, off, fh uint64) (err sysca
 	}
 	defer h.Wunlock()
 
-	// kernel might return a stale length from attr-cache before open as the offset.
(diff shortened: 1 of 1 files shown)


SandyXSD:
If the previous fix works fine for normal cases, and breaks the pages only under the FUSE `writeback_cache` mode, we can adjust the behavior based on the FUSE option.

jiefenghuang:
Another consideration is that it’s unclear whether the kernel may apply special handling to pages in other scenarios, which could cause the offset used for append writes to differ from the file end we expect. Therefore, we have decided to roll back for now.

The original issue can be addressed by disabling the kernel entry-cache to achieve consistency.

PR #6605, opened 2026-01-14
cmd/gateway:improve the performance of batch deletion of objects

zhijian-pro:


zhijian-pro:
close https://github.com/juicedata/juicefs/issues/5094

zhijian-pro:
Performance Test
Using Redis as the engine, delete 10,000 files under the same prefix

old: delete elapsed: 42.198010968s
new: delete elapsed: 4.307741534s
There is a 10x performance improvement.

pkg/gateway/gateway.go
```diff
@@ -507,14 +507,38 @@ func (n *jfsObjects) DeleteObject(ctx context.Context, bucket, object string, op
 func (n *jfsObjects) DeleteObjects(ctx context.Context, bucket string, objects []minio.ObjectToDelete, options minio.ObjectOptions) (objs []minio.DeletedObject, errs []error) {
 	objs = make([]minio.DeletedObject, len(objects))
```
jiefenghuang, line 515: DeleteObject的检测是否挪到这边，做一次检测。
if err = n.checkBucket(ctx, bucket); err != nil {
		return
}

pkg/gateway/gateway.go
```diff
@@ -507,14 +507,38 @@ func (n *jfsObjects) DeleteObject(ctx context.Context, bucket, object string, op
...
-			objs[idx] = minio.DeletedObject{
-				ObjectName: object.ObjectName,
-			}
+	delMap := make(map[string][]int)
+	for idx, o := range objects {
+		p := path.Dir(path.Clean(n.path(bucket, o.ObjectName)))
+		delMap[p] = append(delMap[p], idx)
+	}
+	var g errgroup.Group
+	g.SetLimit(10)
```
jiefenghuang: 改成核数

Issue #6607, opened 2026-01-15
Support multiple subdir for hadoop-sdk


kyungwan-nam:
**What would you like to be added**:

`juicefs.subdir` was introduced in https://github.com/juicedata/juicefs/issues/6096
It is not possible to access the multiple paths.

**Why is this needed**:

kyungwan-nam:
PR: https://github.com/juicedata/juicefs/pull/6608

tangyoupeng:
Hi, why multiple subdir is needed in hadoop sdk since the fuse mount only supports one sub.

kyungwan-nam:
Hello, Thank you for taking your time.

it's required in the following cases:
We use juicefs separately for each user in k8s. e.g. /user1, /user2
and we also analyze data from multiple users with hive.
In situations where hiveserver has to process data from multiple users in Hadoop, It should allow access to the /user1 and /user2 paths.
I could allow all paths without `subdir`, but I'd like to minimize the number of access paths to avoid wrong access.

PR #6608, opened 2026-01-15
hadoop: support multiple subdir

kyungwan-nam:
https://github.com/juicedata/juicefs/issues/6607


kyungwan-nam:
@tangyoupeng 
Thank you for review and merge.

PR #6611, opened 2026-01-16
feat: improve object Get method

201341:


pkg/object/hdfs.go
```diff
@@ -112,7 +112,7 @@ func (h *hdfsclient) Get(ctx context.Context, key string, off, limit int64, gett
 	}
 
 	finfo := f.Stat()
-	if finfo.IsDir() {
+	if finfo.IsDir() || off >= finfo.Size() {
```
jiefenghuang, line 115: _ = f.Close()

201341: fixed it.

pkg/object/tos.go
```diff
@@ -87,6 +83,11 @@ func (t *tosClient) Get(ctx context.Context, key string, off, limit int64, gette
 		v, _ := resp.Meta.Get(checksumAlgr)
 		resp.Content = verifyChecksum(resp.Content, v, resp.ContentLength)
 	}
+	if resp != nil {
```
zhijian-pro: When GetObjectV2 fails, this line of code will not be executed, so the requestId will be lost, which does not meet the expectation.

pkg/object/file.go
```diff
@@ -131,7 +131,7 @@ func (d *filestore) Get(ctx context.Context, key string, off, limit int64, gette
 		_ = f.Close()
 		return nil, err
 	}
-	if finfo.IsDir() || off > finfo.Size() {
+	if finfo.IsDir() || off >= finfo.Size() {
```
zhijian-pro, line 134: You can check the implementations of other objects. There are still some remaining areas that haven't been changed, such as cifs

pkg/object/tos.go
jiefenghuang: This part does not need to be modified, response like RequestID can still be obtained even when an error occurs.

pkg/object/file.go
201341: fix

PR #6626, opened 2026-01-23
cmd/sync: add global traffic control for sync

zhijian-pro:


cmd/sync.go
```diff
@@ -265,6 +265,10 @@ func syncStorageFlags() []cli.Flag {
 			Name:  "bwlimit",
 			Usage: "limit bandwidth in Mbps (0 means unlimited)",
 		},
+		&cli.StringFlag{
+			Name:  "traffic-control-addr",
+			Usage: "the address of the traffic control",
```
davies: `url` or `uri` could be better than `address`

pkg/sync/sync.go
```diff
@@ -66,9 +68,125 @@ var (
...
+		// request credit for other waiters together
+		ask := l.need - l.balance
+		if ask >= bytes*10 {
+			// don't wait for too long
+			ask = bytes * 10
+		}
+		l.Unlock()
+		granted, expire, err := l.request(ask)
+		l.Lock()
+		if err == nil {
```
jiefenghuang, line 159: mark：server-side exception fault tolerance

zhijian-pro: Let's do it after receiving customer feedback

PR #6639, opened 2026-01-27
CI: fix sync test timeout

zhoucheng361:


zhoucheng361:
ref: https://github.com/juicedata/juicefs/actions/runs/21366813551

PR #6642, opened 2026-01-27
Windows: add create-perm mount option

chenjie4255:


cmd/mount_windows.go
```diff
@@ -84,6 +85,19 @@ func mountFlags() []cli.Flag {
 			Name:  "report-case",
 			Usage: "If set, juicefs will report the correct case of a file path for a case-insensitive filesystem. (May incur a performance lost)",
 		},
+		&cli.StringFlag{
+			Name:  "create-perm",
```
jiefenghuang, line 89: Unix has a "umask" flag, we can use the same name

PR #6643, opened 2026-01-27
Windows: add readdir batch size option

chenjie4255:
* This pr decreased the default readdir batch size from 10000 to 1000 on the windows platform.

### DIFF
--- a/pkg/winfsp/winfs.go
+++ b/pkg/winfsp/winfs.go
@@ -793,6 +793,13 @@ func (j *juice) Readdir(path string,
 			e = errorconv(err)
 			return
 		}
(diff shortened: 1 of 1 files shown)


pkg/winfsp/winfs.go
```diff
@@ -824,6 +821,11 @@ func (j *juice) Readdir(path string,
 				break
 			}
 		}
+
+		if len(entries) < j.readdirBatchSize {
```
jiefenghuang: Readdir的返回值,len(entries) 可能小于传入的readdirBatchSize, ==0才代表结束

PR #6645, opened 2026-01-27
Windows: add admin-as-root option for user permissions

chenjie4255:
This PR introduces a new --admin-as-root option (default: true) that allows treating the Windows built-in Administrator user as root (UID 0) on Linux, improving cross-platform permission compatibility.

## Changes

* New --admin-as-root mount option: Controls whether Windows Administrator should be mapped to UID 0
  * Default: true for better compatibility with Unix permission models
  * When enabled: Administrator (SID S-1-5-21-...-500) → UID 0
  * When disabled: Uses System user mapping (UID 18) (not good, in most of case linux system cannot handle this )
* Windows SID to UID/GID conversion
  * Implements Cygwin-compatible SID-to-POSIX mapping algorithm
  * Handles local accounts, domain accounts, and well-known SIDs
  * Supports trusted domain accounts with LDAP-based trustPosixOffset queries
  * Distinguished users:
     * Administrator: UID 197108 (0x30000 + 500)
     * Administrators group: GID 544
     * System: UID 18
* LDAP integration: 
   * Connects to Active Directory via LDAP
   * Queries trustPosixOffset for trusted domains
   * Enables proper UID/GID mapping for cross-domain accounts

### DIFF
--- a/pkg/win/ldap.go
+++ b/pkg/win/ldap.go
@@ -1,6 +1,22 @@
 //go:build windows
 // +build windows
 
+/*
+ * JuiceFS, Copyright 2026 Juicedata, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package win
 
 import (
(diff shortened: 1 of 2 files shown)


pkg/win/ldap.go
```diff
@@ -0,0 +1,159 @@
...
+import (
+	"fmt"
+	"strings"
+	"unsafe"
+
+	"golang.org/x/sys/windows"
+)
+
+var (
+	modWldap32           = windows.NewLazySystemDLL("wldap32.dll")
```
jiefenghuang, line 31: some questions:
1. Are all Windows versions had the same dll?
2. Is the dll included by default? Should this function have error-handling/fallback logic?
3. Are constants the same across all Windows versions?

chenjie4255: 1. both dll inlcuded in this pr should be included in all major windows version as they are the core part of the windows sytem. 
2. Dynamic addressing is used instead of absolute addressing. Currently, these called functions should all be included within these DLLs.
3. yes they are defined in the windows header file.

Issue #6640, opened 2026-01-27
CI: panic in pkg/meta.(*dbMeta).doBatchUnlink.func1


zhoucheng361:
**What happened**:
https://github.com/juicedata/juicefs/actions/runs/21373202309/job/61522822003

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x2cf14d7]

goroutine 1 [running]:
github.com/juicedata/juicefs/pkg/meta.(*dbMeta).doBatchUnlink.func1(0xc001b425b0)
	/home/runner/work/juicefs/juicefs/pkg/meta/sql.go:2788 +0x1c17
github.com/juicedata/juicefs/pkg/meta.(*dbMeta).txn.func2(0xc001b425b0?)
	/home/runner/work/juicefs/juicefs/pkg/meta/sql.go:1065 +0x22
xorm.io/xorm.(*Engine).Transaction(0x8fb?, 0xc001008248, {0x0, 0x0, 0x420a31?})
	/home/runner/go/pkg/mod/gitea.com/davies/xorm@v1.0.8-0.20220528043536-552d84d1b34a/engine.go:1312 +0xac
github.com/juicedata/juicefs/pkg/meta.(*dbMeta).txn(0xc000683c80, 0xc001008430, {0x0, 0x0, 0x0})
	/home/runner/work/juicefs/juicefs/pkg/meta/sql.go:1064 +0x3a6
github.com/juicedata/juicefs/pkg/meta.(*dbMeta).doBatchUnlink(0xc000683c80, {0x509cfb8, 0xc0005cb9c0}, 0x7fffffff10000001, {0xc00134e008, 0x130, 0x25f}, 0xc001b46798, 0xc001b467a0, 0xc001b467a8, ...)
	/home/runner/work/juicefs/juicefs/pkg/meta/sql.go:2636 +0x565
github.com/juicedata/juicefs/pkg/meta.(*baseMeta).BatchUnlink(0xc00062b908, {0x509cfb8, 0xc0005cb9c0}, 0x7fffffff10000001, {0xc00134e008, 0x130, 0x25f}, 0xc000857b40, 0x1)
	/home/runner/work/juicefs/juicefs/pkg/meta/base.go:1708 +0x203
[...]

PR #6646, opened 2026-01-27
meta/sql: fix nil pointer in doBatchUnlink

Xuyuchao-juice:
close https://github.com/juicedata/juicefs/issues/6640


jiefenghuang (changes requested):
m.sid影响opened的逻辑要修改

PR #6652, opened 2026-01-30
cmd/sync: limit the number of entries retained in each directory level

zhijian-pro:


pkg/object/object_storage.go
```diff
@@ -219,7 +219,7 @@ func ListAllWithDelimiter(ctx context.Context, store ObjectStorage, prefix, star
 					if !entries[i].IsDir() || key == prefix {
 						continue
 					}
-					t.entries, t.hasMore, t.nextToken, t.err = store.List(ctx, key, "\x00", t.nextToken, "/", 1e9, followLink) // exclude itself
+					t.entries, t.hasMore, t.nextToken, t.err = store.List(ctx, key, "\x00", t.nextToken, "/", 1000, followLink) // exclude itself
```
jiefenghuang, line 229: mark: the peak memory usage inside readdir is still high; it’s freed after the call returns.

pkg/object/object_storage.go
```diff
@@ -259,6 +259,17 @@ func ListAllWithDelimiter(ctx context.Context, store ObjectStorage, prefix, star
...
+				if t.err != nil {
+					err = t.err
+					t.Unlock()
+					return err
+				}
+				t.entries = append(t.entries, more...)
+			}
 			t.ready = false
 			t.cond.Signal()
 			children := t.entries
```
davies, line 282: reset t.entries to release the memory earlier

pkg/object/object_storage.go
```diff
@@ -259,9 +259,21 @@ func ListAllWithDelimiter(ctx context.Context, store ObjectStorage, prefix, star
 				t.Unlock()
 				return err
 			}
+			for t.hasMore {
```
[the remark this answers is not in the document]
zhijian-pro: This situation will not occur.

PR #6665, opened 2026-02-04
cmd/sync: change tmp file name

zhijian-pro:
…rom internal files


cmd/object.go
```diff
@@ -139,7 +139,7 @@ func (j *juiceFS) Put(rCtx context.Context, key string, in io.Reader, getters ..
 		if len(name) > 200 {
 			name = name[:200]
 		}
-		tmp = path.Join(path.Dir(p), fmt.Sprintf(".%s.tmp.%d", name, rand.Int()))
+		tmp = path.Join(path.Dir(p), fmt.Sprintf(".jfs.%s.tmp.%d", name, rand.Int()))
```
jiefenghuang: we should change other filesystem backends like cifs/nfs...
Add a common function so that multiple backends remain consistent?

zhijian-pro: ok

jiefenghuang (approved):

PR #6668, opened 2026-02-05
meta: add delfile_remain metric to track pending files

Xuyuchao-juice:


jiefenghuang:
The job status and the processed count are enough

Issue #6681, opened 2026-02-25
`ls .trash`: always stale .trash readdir entries with readdir-cache enabled


zhitaoli-6:
**What happened**:
1. `juicefs mount` is started with readdir-cache enabled and trash-days 1, 
2. `ls .trash` see no entries.
3. delete some file manully.
4. `ls .trash` still see no entries. `stat .trash/<date-hour>` succeeds with the attributes. 
5. `ls .trash` still see no entries, even after 12hours. It seems we will always read stale data in this case.

**What you expected to happen**:
`ls .trash` can see the entries in time, or after some timeout.

**How to reproduce it (as minimally and precisely as possible)**:
As above.

**Anything else we need to know?**

With readdir-cache enabled, the kernel uses attr.mtime to check whether the readdir cache is valid after timeout. However, the attr.mtime of .trash is hard-coded and doesn't change after creation. So the kernel never invalidates readdir cache and `ls .trash` see  always stale entries.

We can simply fix the issue by disabling readdir-cache in special inodes.
I have tested the fix in our internal environment and it does work.

<img width="1377" height="260" alt="Image" src="https://github.com/user-attachments/assets/2c75cbce-e5ad-4870-b986-37a3412ca4c1" />

**Environment**:
- JuiceFS version (use `juicefs --version`) or Hadoop Java SDK version: 1.4.0
- Cloud provider or hardware configuration running JuiceFS: OSS inside
- OS (e.g `cat /etc/os-release`):Ubuntu 22.04.5 LTS
- Kernel (e.g. `uname -a`): 5.10.25 
- Object storage (cloud provider and region, or self maintained):OSS inside
- Metadata engine info (version, cloud provider managed or self maintained): TiKV
- Network connectivity (JuiceFS to metadata engine, JuiceFS to object storage): Reachable
- Others:

zhitaoli-6:
I'm coming here check whether it is a bug and whether the solution is acceptable to our community.

jiefenghuang:
Thank you for your feedback — you’re right, this issue exists.

zhitaoli-6:
@jiefenghuang Are you a member of JuiceFS?  Could I submit a commit to fix this issue? It will be a small milestone for me to contribute to this project :)

jiefenghuang:
Sure, prs are welcome

PR #6690, opened 2026-02-27
fuse: disable readdir-cache for special inodes like `.trash`

zhitaoli-6:
The attr of .trash doesn't change after creation. If readdir-cache is enabled, we always see stale entries by `ls`.

It fixes https://github.com/juicedata/juicefs/issues/6681

### DIFF
--- a/pkg/fuse/fuse.go
+++ b/pkg/fuse/fuse.go
@@ -362,9 +362,10 @@ func (fs *fileSystem) Flock(cancel <-chan struct{}, in *fuse.LkIn, block bool) (
 func (fs *fileSystem) OpenDir(cancel <-chan struct{}, in *fuse.OpenIn, out *fuse.OpenOut) (status fuse.Status) {
 	ctx := fs.newContext(cancel, &in.InHeader)
 	defer releaseContext(ctx)
-	fh, err := fs.v.Opendir(ctx, Ino(in.NodeId), in.Flags)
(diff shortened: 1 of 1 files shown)


pkg/fuse/fuse.go
```diff
@@ -362,9 +362,11 @@ func (fs *fileSystem) Flock(cancel <-chan struct{}, in *fuse.LkIn, block bool) (
 func (fs *fileSystem) OpenDir(cancel <-chan struct{}, in *fuse.OpenIn, out *fuse.OpenOut) (status fuse.Status) {
 	ctx := fs.newContext(cancel, &in.InHeader)
 	defer releaseContext(ctx)
-	fh, err := fs.v.Opendir(ctx, Ino(in.NodeId), in.Flags)
+	ino := Ino(in.NodeId)
+	fh, err := fs.v.Opendir(ctx, ino, in.Flags)
 	out.Fh = fh
-	if fs.conf.ReaddirCache {
+	// The attr of special directories like .trash doesn't change after creation. So we disable the readdir-cache for these directories to avoid stale data.
```
jiefenghuang: Since the code is already fairly clear, and there are a PR and an issue tracking the reason for this commit, we can remove this code comment.

zhitaoli-6: Thanks. Resolved.

PR #6701, opened 2026-03-03
vfs: add path protection feature with readonly and deny modes

shenchucheng:
Add path-based access control to the VFS layer, allowing users to protect specific paths via regex patterns with --path-protection flag. Supports "readonly" (block writes, allow reads) and "deny" (block all) modes. Protection checks are enforced on Lookup, Open, Read, Write, Truncate, Create, Mknod, Unlink, Mkdir, Rmdir, Symlink, Rename, and Link operations. Invalid config causes mount to fail immediately rather than silently degrading. Regex patterns recommend ^ anchoring to avoid unintended substring matches.


jiefenghuang:
Hi @shenchucheng , this can likely be addressed using filesystem UGO or ACLs. Are there any specific use cases that cannot be handled this way?

PR #6720, opened 2026-03-09
object/cos: Add mds for upload part

201341:
when set md5, cos server will check the consistent https://cloud.tencent.com/document/product/436/36427

### DIFF
--- a/pkg/object/cos.go
+++ b/pkg/object/cos.go
@@ -22,6 +22,8 @@ package object
 import (
 	"bytes"
 	"context"
+	"crypto/md5"
+	"encoding/base64"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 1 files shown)


jiefenghuang:
EnableCRC is already enabled by default, so MD5 should not be necessary.

201341:
> EnableCRC is already enabled by default, so MD5 should not be necessary.

EnableCRC just for Put interface. For Upload interface, set md5 is necessary.

jiefenghuang:
> > EnableCRC is already enabled by default, so MD5 should not be necessary.
> 
> EnableCRC just for Put interface. For Upload interface, set md5 is necessary.

SDK will validates the server’s response for UploadPart too, in **(c *Client) doAPI**

image

201341:
> > > EnableCRC is already enabled by default, so MD5 should not be necessary.
> > 
> > 
> > EnableCRC just for Put interface. For Upload interface, set md5 is necessary.
> 
> SDK will validates the server’s response for UploadPart too, in *_(c _Client) doAPI__
> 
> image

Got it

PR #6725, opened 2026-03-10
meta/badger: improve reliability and resource management

solracsf:
### Summary
This PR addresses several reliability issues in the Badger metadata backend that could cause silent data corruption, resource leaks, and unexpected crashes under certain conditions.

### Changes
- Fixed a case where reading keys during iteration could return corrupted data
- Fixed a goroutine that would leak memory after the database was closed
- Fixed a transaction resource that could be left unreleased on failure
- Fixed operations that would crash instead of recovering gracefully when a transaction grew too large
- Fixed scan operations that were doing unnecessary work when only keys were needed

### DIFF
--- a/pkg/meta/tkv_badger.go
+++ b/pkg/meta/tkv_badger.go
@@ -104,16 +104,7 @@ func (tx *badgerTxn) exist(prefix []byte) bool {
 }
 
 func (tx *badgerTxn) set(key, value []byte) {
-	err := tx.t.Set(key, value)
-	if err == badger.ErrTxnTooBig {
-		logger.Warn("Current transaction is too big, commit it")
-		if er := tx.t.Commit(); er != nil {
-			panic(er)
-		}
-		tx.t = tx.c.NewTransaction(true)
-		err = tx.t.Set(key, value)
-	}
-	if err != nil {
+	if err := tx.t.Set(key, value); err != nil {
 		panic(err)
 	}
 }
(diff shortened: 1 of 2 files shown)


pkg/meta/tkv_badger.go
```diff
@@ -126,14 +134,24 @@ func (tx *badgerTxn) incrBy(key []byte, value int64) int64 {
...
 
 func (tx *badgerTxn) delete(key []byte) {
-	if err := tx.t.Delete(key); err != nil {
+	err := tx.t.Delete(key)
+	if err == badger.ErrTxnTooBig {
+		logger.Warn("Current transaction is too big on delete, commit it")
+		if er := tx.t.Commit(); er != nil {
+			panic(er)
+		}
+		tx.t = tx.c.NewTransaction(true)
```
jiefenghuang: This breaks the transactional guarantees.

solracsf: In this case `set()` should also be fixed the same way, as it also introduces a silent atomicity violation. Right?

jiefenghuang: Oh yes. Normal operations such as attribute access, reads, and writes are unlikely to cause ErrTxnTooBig. Only some batch operations (whose transactional impact is relatively small) might trigger it. 

So it’s better to keep both for now to avoid some potential panics.

pkg/meta/tkv_test.go
```diff
@@ -284,3 +288,119 @@ func TestMemKV(t *testing.T) {
...
+func TestBadgerCloseExitsGCGoroutine(t *testing.T) {
+	before := runtime.NumGoroutine()
+
+	c, err := newBadgerClient(t.TempDir())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Allow the GC goroutine to start
+	time.Sleep(20 * time.Millisecond)
```
jiefenghuang: this maybe unstable？Since the added logic is straightforward, we can skip this unit test.

jiefenghuang (approved):

jiefenghuang:
@solracsf Thanks for your contribution.

PR #6724, opened 2026-03-10
CI:test gid quota

YunhuiChen:

PR #6743, opened 2026-03-15
meta/quota: fix create missing user/group quota records in doFlushQuotas

Xuyuchao-juice:
Previously, the doFlushQuotas function only updated the existing user/group quotas. This could lead to inaccurate usage statistics. Now, it is necessary to support the creation of new quotas during the flush process.

rel ci test: https://github.com/juicedata/juicefs/pull/6724

### DIFF
--- a/pkg/meta/sql.go
+++ b/pkg/meta/sql.go
@@ -4145,9 +4145,6 @@ func (m *dbMeta) doLoadQuotas(ctx Context) (map[uint64]*Quota, map[uint64]*Quota
 
 	// Load user and group quotas
 	for _, q := range userGroupQuotasList {
-		if q.MaxSpace < 0 && q.MaxInodes < 0 {
-			continue
-		}
 		quota := &Quota{
 			MaxSpace:   q.MaxSpace,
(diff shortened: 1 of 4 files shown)


pkg/meta/sql.go
```diff
@@ -4171,18 +4168,34 @@ func (m *dbMeta) doFlushQuotas(ctx Context, quotas []*iQuota) error {
...
 					return err
 				}
 			} else {
-				_, err := s.Exec(m.sqlConv("update user_group_quota set used_space=used_space+?, used_inodes=used_inodes+? where qtype=? and qkey=?"),
+				ret, err := s.Exec(m.sqlConv("update user_group_quota set used_space=used_space+?, used_inodes=used_inodes+? where qtype=? and qkey=?"),
 					q.quota.newSpace, q.quota.newInodes, q.qtype, q.qkey)
 				if err != nil {
 					return err
 				}
+				affected, err := ret.RowsAffected()
```
jiefenghuang, line 4182: TODO: upsert

Issue #6729, opened 2026-03-11
quota: The quota is not reclaimed after deleting data


YunhuiChen:
**What happened**:
```
root@chaos-k8s-001:~/chenyunhui/juicefs# ./juicefs quota set redis://127.0.0.1:6379/1 --uid 2001 --capacity 1 --inodes 5
2026/03/11 15:16:35.640359 juicefs[4049647] <INFO>: Meta address: redis://127.0.0.1:6379/1 [NewClient@interface.go:627]
2026/03/11 15:16:35.640793 juicefs[4049647] <WARNING>: AOF is not enabled, you may lose data if Redis is not shutdown properly. [checkRedisInfo@info.go:84]
2026/03/11 15:16:35.640850 juicefs[4049647] <INFO>: Ping redis latency: 20.67µs [checkServerConfig@redis.go:4304]
+----------+---------+------+------+--------+-------+-------+
|  User ID |   Size  | Used | Use% | Inodes | IUsed | IUse% |
+----------+---------+------+------+--------+-------+-------+
| UID:2001 | 1.0 GiB |  0 B |   0% |      5 |     0 |    0% |
+----------+---------+------+------+--------+-------+-------+

root@chaos-k8s-001:~/chenyunhui/juicefs# dd if=/dev/zero of=/juicefs2/2001 bs=1G count=1
1+0 records in
1+0 records out
1073741824 bytes (1.1 GB, 1.0 GiB) copied, 4.98064 s, 216 MB/s

root@chaos-k8s-001:~/chenyunhui/juicefs# chown 2001:2001 /juicefs2/2001

root@chaos-k8s-001:~/chenyunhui/juicefs# ./juicefs quota list redis://127.0.0.1:6379/1
2026/03/11 15:17:42.176569 juicefs[4050141] <INFO>: Meta address: redis://127.0.0.1:6379/1 [NewClient@interface.go:627]
2026/03/11 15:17:42.176993 juicefs[4050141] <WARNING>: AOF is not enabled, you may lose data if Redis is not shutdown properly. [checkRedisInfo@info.go:84]
2026/03/11 15:17:42.177067 juicefs[4050141] <INFO>: Ping redis latency: 43.77µs [checkServerConfig@redis.go:4304]
2026/03/11 15:17:42.177574 juicefs[4050141] <WARNING>: Used inodes of uid:1000 is negative (-330), please run `juicefs quota check` to fix it [quota@quota.go:232]
[...]

PR #6744, opened 2026-03-15
meta/quota : make trash files count towards user/group quota usage

Xuyuchao-juice:
Previously, files in trash did not count towards user/group quota usage,
which was inconsistent with the file system total quota behavior.

This change makes trash files also count towards user/group quota,
simplifying the implementation by aligning with the total quota strategy:

- Files moved to trash: no longer reduce user/group quota
- Files deleted from trash: reduce user/group quota
- Files restored from trash: no longer increase user/group quota
- Initial quota scan: now includes trash directory

The quota update logic is moved to the engine layer (sql.go, redis.go, tkv.go)
and follows the same pattern as `updateStats`, which only updates when
`trash == 0` (i.e., files are permanently deleted).

close https://github.com/juicedata/juicefs/issues/6729

### DIFF
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -189,39 +189,37 @@ type symlinkCache struct {
 	cap  int32
 }
 
-// userGroupQuotaDelta represents quota changes for a specific user and group.
-type userGroupQuotaDelta struct {
+// ugQuotaDelta represents quota changes for a specific user and group.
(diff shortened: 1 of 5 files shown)


jiefenghuang:
- redis cfr stat miss
- rename stat
- batchunlink stat in outer layer

jiefenghuang:
TODO: CI

PR #6750, opened 2026-03-18
ref #6749 pkg/object/restful.go prefer ipv6 connection over ipv4

### LINKED ISSUE #6749: Broken HTTP Dial for Object-Storage
That the Dialer who does the http calls to the Object Storage backend prefers IPv6 over IPv4 and disables either IPv4 or IPv6 when the protocol isnt setup.

This behaviour should be somewhat inline to https://www.rfc-editor.org/rfc/rfc8305#section-3

```
The algorithm proceeds as follows: if a positive AAAA response (a
   response with at least one valid AAAA record) is received first, the
   first IPv6 connection attempt is immediately started.  If a positive
   A response is received first due to reordering, the client SHOULD
   wait a short time for the AAAA response to ensure that preference is
   given to IPv6 (it is common for the AAAA response to follow the A
   response by a few milliseconds).  This delay will be referred to as
   the "Resolution Delay".  The recommended value for the Resolution
   Delay is 50 milliseconds.  If a positive AAAA response is received
within the Resolution Delay period, the client immediately starts the
   IPv6 connection attempt.  If a negative AAAA response (no error, no
   data) is received within the Resolution Delay period or the AAAA
   response has not been received by the end of the Resolution Delay
   period, the client SHOULD proceed to sorting addresses (see
   [Section 4](https://www.rfc-editor.org/rfc/rfc8305#section-4)) and staggered connection attempts (see [Section 5](https://www.rfc-editor.org/rfc/rfc8305#section-5)) using
   any IPv4 addresses returned so far.  If the AAAA response arrives
   while these connection attempts are in progress but before any
   connection has been established, then the newly received IPv6
   addresses are incorporated into the list of available candidate
   addresses (see [Section 6](https://www.rfc-editor.org/rfc/rfc8305#section-6)) and the process of connection attempts will
   continue with the IPv6
addresses added, until one connection is
   established.
```

**Why is this needed**:

The behavior is more expected and it fixes random Timeouts.

The Dialer  (pkg/object/restful.go) in its current state is more or less "broken".
Currently it resolves all of the IPs of a host, including both v4 and v6, and picks one randomly in a loop until we have a connection.

On either v4 only and v6 only hosts, when the Host that we want to dial has dual stack enabled, will the Dialer try randomly to connect to one of the given IPs. 
This leads to frequent timeout, although the network side is configured correctly.

scrapes:
Prefers IPv6 IP Pools over IPV4 Pools when Dialing via HTTP for the object backend.

Should close #6749

### DIFF
--- a/pkg/object/restful.go
+++ b/pkg/object/restful.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"io"
 	"math/rand"
@@ -27,6 +28,7 @@ import (
 	"net/http"
 	"os"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/viki-org/dnscache"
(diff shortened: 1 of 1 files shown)


scrapes:
Please comment if the error check for syscall.ENETUNREACH should stay.
The error behavior of the Dialer is not really documented so this is more or less a best effort check to skip further ipv6 dial calls if the network is not reachable. The intention behind this is to minimize connection delays with the cost of possibly skipping reachable servers; although i think in real world applications the scenario of having multiple ipv6 hosts in your DNS where only a subset has a reachable Route is not really common.

pkg/object/restful.go
```diff
@@ -35,7 +37,50 @@ import (
...
+	first := rand.Intn(n)
+	var conn net.Conn
+	var err error
+	for i := 0; i < n; i++ {
+		ip := ips[(first+i)%n]
+		conn, err = dialer.DialContext(ctx, network, net.JoinHostPort(ip.String(), port))
+		if err == nil {
+			return conn, nil
+		}
+		if isNoRouteError(err) {
```
scrapes, line 67: See comment in PR

pkg/object/restful.go
```diff
@@ -35,7 +37,50 @@ import (
...
+		}
+	}
+	return nil, err
+}
+
+func isNoRouteError(err error) bool {
+	if err == nil {
+		return false
+	}
+	return errors.Is(err, syscall.ENETUNREACH)
```
solracsf, line 78: Is this Windows friendly?
It is also missing `EHOSTUNREACH`

pkg/object/restful.go
```diff
@@ -35,7 +37,50 @@ import (
...
+}
+
 func init() {
+	dialer := &net.Dialer{Timeout: 10 * time.Second}
```
solracsf, line 82: Isn't this too high?

jiefenghuang:
Hi @scrapes, pls take a look at #6753
I think using dialParallel is better, especially for latency. 
However, Go currently doesn’t allow hooking into the process, so I copy and modify the implementation.

jiefenghuang:
close by #6753

pkg/object/restful.go
scrapes: The logic is that when no Route to host exists we assume no connectivity for that IP Version so that we skip the pool because we will most likely do not have Routes to the other hosts.
EHOSTUNREACH is more broad if the Host specifically not reachable.

This should be posix only and not Windows friendly.

PR #6753, opened 2026-03-19
object: fix client dial delay and prefer ipv6

### LINKED ISSUE #6749: Broken HTTP Dial for Object-Storage

jiefenghuang:
- take over #6750
- close #6749

### DIFF
--- a/pkg/object/restful.go
+++ b/pkg/object/restful.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"io"
 	"math/rand"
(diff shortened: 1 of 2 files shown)


pkg/object/restful.go
```diff
@@ -46,32 +155,23 @@ func init() {
...
+				dialer := &net.Dialer{Timeout: time.Second * 10}
+				if ip := net.ParseIP(host); ip != nil {
+					return dialer.DialContext(ctx, network, net.JoinHostPort(ip.String(), port))
+				}
 				ips, err := resolver.Fetch(host)
 				if err != nil {
 					return nil, err
 				}
 				if len(ips) == 0 {
 					return nil, fmt.Errorf("No such host: %s", host)
```
solracsf:
```suggestion
					return nil, &net.DNSError{Err: "no such host", Name: host, IsNotFound: true}
```

pkg/object/restful.go
```diff
@@ -35,6 +35,115 @@ import (
...
+			if res.error == nil {
+				return res.Conn, nil
+			}
+			if res.primary {
+				primary = res
+			} else {
+				fallback = res
+			}
+			if primary.done && fallback.done {
+				return nil, primary.error
```
solracsf:
```suggestion
				return nil, errors.Join(primary.error, fallback.error)
```

pkg/object/restful.go
```diff
@@ -46,32 +155,23 @@ func init() {
...
 			WriteBufferSize:       32 << 10,
 			DialContext: func(ctx context.Context, network string, address string) (net.Conn, error) {
-				separator := strings.LastIndex(address, ":")
-				host := address[:separator]
-				port := address[separator:]
+				host, port, err := net.SplitHostPort(address)
+				if err != nil {
+					return nil, err
+				}
+				dialer := &net.Dialer{Timeout: time.Second * 10}
```
solracsf: Move dialer instantiation outside the DialContext closure?

scrapes:
Looks good. 
In the dialRandom i would setup a check on len == 1 to skip directly to dialing the single IP.

The DNS Cache Used is pretty old and looks unmaintained ( Last commit 13 years ago). Its only a very small project with a map from Host to addresses. Maybe its easier to just fork it to have more control over it or copy it into the codebase.

The reason i am mention this is to add a Remove function to the dns cache wich will remove single IPs from the dns cache. Or we could implement a super simple negative cache on top of it to skip the dialing of known bad connections.

jiefenghuang:
"In the dialRandom i would setup a check on len == 1 to skip directly to dialing the single IP."
I think there’s no need to add extra logic for this.

"The reason i am mention this is to add a Remove function to the dns cache wich will remove single IPs from the dns cache. Or we could implement a super simple negative cache on top of it to skip the dialing of known bad connections."
It might be caused by a temporary network issue, so let’s keep the existing logic for now.

scrapes:
> "In the dialRandom i would setup a check on len == 1 to skip directly to dialing the single IP." I think there’s no need to add extra logic for this.
> 
> "The reason i am mention this is to add a Remove function to the dns cache wich will remove single IPs from the dns cache. Or we could implement a super simple negative cache on top of it to skip the dialing of known bad connections." It might be caused by a temporary network issue, so let’s keep the existing logic for now.

Yes but the Address would be back in cache after the next cache refresh.
But i agree, seperate issue

PR #6771, opened 2026-03-24
cmd/quota: fix user/group quota keys display

Xuyuchao-juice:
Before fix
```
root@chaos-k8s-001: ./juicefs quota list redis://127.0.0.1/1
+---------+---------+------+------+-----------+-------+-------+
| Path/ID |   Size  | Used | Use% |   Inodes  | IUsed | IUse% |
+---------+---------+------+------+-----------+-------+-------+
| 1000    | 1.0 GiB |  0 B |   0% | unchanged |     0 |       |
+---------+---------+------+------+-----------+-------+-------+
```
After fix
```
+---------+---------+------+------+-----------+-------+-------+
| Path/ID |   Size  | Used | Use% |   Inodes  | IUsed | IUse% |
+---------+---------+------+------+-----------+-------+-------+
| uid:1000| 1.0 GiB |  0 B |   0% | unchanged |     0 |       |
+---------+---------+------+------+-----------+-------+-------+
```

### DIFF
--- a/pkg/meta/quota.go
+++ b/pkg/meta/quota.go
@@ -889,12 +889,12 @@ func (m *baseMeta) handleQuotaList(ctx Context, qtype uint32, key uint64, quotas
 	}
 	for uid, quota := range userQuotas {
 		if match(UserQuotaType, uid, quota) {
-			quotas[fmt.Sprintf("UID:%d", uid)] = quota
+			quotas[fmt.Sprintf("uid:%d", uid)] = quota
 		}
 	}
 	for gid, quota := range groupQuotas {
 		if match(GroupQuotaType, gid, quota) {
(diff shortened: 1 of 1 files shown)


pkg/meta/quota.go
```diff
@@ -889,12 +889,12 @@ func (m *baseMeta) handleQuotaList(ctx Context, qtype uint32, key uint64, quotas
 	}
 	for uid, quota := range userQuotas {
 		if match(UserQuotaType, uid, quota) {
-			quotas[fmt.Sprintf("%d", uid)] = quota
+			quotas[fmt.Sprintf("UID:%d", uid)] = quota
```
jiefenghuang: Use lowercase to keep it consistent with inode.

PR #6786, opened 2026-03-26
cmd/config: improve remove tier

zhijian-pro:


cmd/config.go
```diff
@@ -316,13 +316,27 @@ func config(ctx *cli.Context) error {
 			format.MinClientVersion = "1.4.0-A"
 			clientVer = true
 		case "tier-id":
+			if !ctx.IsSet("tier-sc") {
+				logger.Errorf("missing required flag: --tier-sc")
```
jiefenghuang: logger.Fatal

Issue #6811, opened 2026-03-31
quota: After loading, the data written exceeded the quota limit.


YunhuiChen:
**What happened**:

CI： https://github.com/juicedata/juicefs/actions/runs/23774876646/job/69277404521?pr=6808 

test:
```
run_dump_load_uid_gid_case(){
    prepare_ug_quota_test
    resolve_test_users || return 0

    mkdir -p /jfs/dumpq
    chmod 777 /jfs/dumpq

    ./juicefs quota set $META_URL --uid "$TEST_UID_1" --capacity 1 --inodes 100
    ./juicefs quota set $META_URL --gid "$TEST_GID_2" --capacity 1 --inodes 100
    sleep $((HEARTBEAT_INTERVAL+HEARTBEAT_SLEEP))

    run_as_user_cmd "$TEST_USER_1" "dd if=/dev/zero of=/jfs/dumpq/uid_pre bs=512M count=1"
    run_as_user_cmd "$TEST_USER_2" "dd if=/dev/zero of=/jfs/dumpq/gid_pre bs=512M count=1"
    sleep $DIR_QUOTA_FLUSH_INTERVAL

    ./juicefs dump --log-level error $META_URL --fast > dump.json
    umount_jfs /jfs $META_URL
    python3 .github/scripts/flush_meta.py $META_URL
    ./juicefs load $META_URL dump.json
    ./juicefs mount $META_URL /jfs -d --heartbeat $HEARTBEAT_INTERVAL
    sleep $((HEARTBEAT_INTERVAL+HEARTBEAT_SLEEP))
    chmod 777 /jfs/dumpq

    run_as_user_cmd "$TEST_USER_1" "dd if=/dev/zero of=/jfs/dumpq/uid_post bs=1G count=1" 2>error.log \
        && echo "uid capacity quota should be enforced after dump/load" && exit 1 || true
    grep -i "Disk quota exceeded" error.log || (echo "uid quota not preserved by dump/load" && exit 1)
```

```
runner@runnervm19da2:~/work/juicefs/juicefs$ ./juicefs  quota list redis://127.0.0.1:6379/1?
warning: GOCOVERDIR not set, no coverage data emitted
2026/03/31 02:06:24.165580 juicefs[21482] <INFO>: Meta address: redis://127.0.0.1:6379/1? [NewClient@interface.go:641]
[...]

PR #6813, opened 2026-03-31
cmd/load: ensure user/group quota usage is scanned after files …

Xuyuchao-juice:
close https://github.com/juicedata/juicefs/issues/6811

1. Ensure quota usage scanning is performed **after** files are loaded.  
2. When exporting to Redis, filter out **user/group entries that have usage only but no quota configured**.  
3. In SQL `doSetQuota`, initialize the **max usage value** for user/group entries **without a configured quota**.

### DIFF
--- a/pkg/meta/sql.go
+++ b/pkg/meta/sql.go
@@ -4021,11 +4021,11 @@ func (m *dbMeta) doGetQuota(ctx Context, qtype uint32, key uint64) (*Quota, erro
 
 func updateQuotaFields(quota *Quota, exist bool, maxSpace, maxInodes *int64, usedSpace, usedInodes *int64) []string {
 	updateColumns := make([]string, 0, 4)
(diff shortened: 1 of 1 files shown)


pkg/meta/sql.go
jiefenghuang, line 4078: move here

pkg/meta/redis.go
```diff
@@ -4970,6 +4970,9 @@ func (m *redisMeta) loadQuotasForDump(ctx Context, quotaKey string) map[uint64]*
 		}
 		var quota DumpedQuota
 		quota.MaxSpace, quota.MaxInodes = m.parseQuota([]byte(v))
+		if quota.MaxSpace == -1 && quota.MaxInodes == -1 {
```
jiefenghuang, line 4973: sql and tkv

PR #6849, opened 2026-04-07
Add an option to redis metadata engine for a generic key prefix

gtratnack-alteryx:
As a JuiceFS user utilizing Redis Enterprise Cloud or Redis-compatible managed services, I want to configure an explicit metadata key prefix in the Redis META-URL, so that I can safely isolate multiple JuiceFS filesystems within the same Redis deployment even when non-zero Redis DB numbers are unsupported.

Background

Today JuiceFS derives Redis metadata namespacing from the /db portion of the META-URL.
For Redis Cluster, JuiceFS converts that DB number into a hash-tag prefix like {1} so all filesystem metadata stays in one slot.
For Redis Enterprise Cloud, it rejects any DB other than 0, which makes the current DB-based prefixing approach unusable for multi-filesystem isolation.

Implementation Notes

- Parse prefix from Redis metadata URL query params.
- If prefix is not provided, JuiceFS preserves the existing cluster behavior and uses {db} as the metadata key prefix.
- If prefix is provided, JuiceFS uses {prefix} as the metadata key prefix instead of {db}.
- If the supplied prefix already contains a Redis hash tag, JuiceFS preserves it as-is. 
- Avoid Redis glob metacharacters in because some internal scans use SCAN patterns.

Example

`redis://enterprise.example:6379/0?prefix=tenantA`

Standalone keys: `tenantAi123`, `tenantAsetting`

Cluster keys: `{tenantA}i123`, `{tenantA}setting`

Feature Request in this Issue: https://github.com/juicedata/juicefs/issues/6850


pkg/meta/redis.go
```diff
@@ -830,7 +852,7 @@ func (m *redisMeta) doSyncVolumeStat(ctx Context) error {
 	}
 
 	var inoKeys []string
-	if err := m.scan(ctx, m.prefix+"session*", func(keys []string) error {
+	if err := m.scan(ctx, "session*", func(keys []string) error {
```
gtratnack-alteryx, line 869: This seemed to be an existing bug, but would like for someone from the team to confirm?

docs/en/administration/metadata/redis_best_practices.md
```diff
@@ -72,14 +72,14 @@ It should be noted that since the data of the Redis master node is asynchronousl
 This feature requires JuiceFS v1.0.0 or higher
 :::
 
-JuiceFS also supports Redis Cluster as a metadata engine, the `META-URL` format is `redis[s]://[[USER]:PASSWORD@]ADDR:PORT,[ADDR:PORT],[ADDR:PORT][/DB]`. For example:
+JuiceFS also supports Redis Cluster as a metadata engine, the `META-URL` format is `redis[s]://[[USER]:PASSWORD@]ADDR:PORT,[ADDR:PORT],[ADDR:PORT][/DB][?prefix=<prefix>]`. For example:
```
CaitinChen, line 75:
```suggestion
JuiceFS also supports Redis Cluster as a metadata engine. The `META-URL` format is `redis[s]://[[USER]:PASSWORD@]ADDR:PORT,[ADDR:PORT],[ADDR:PORT][/DB][?prefix=<prefix>]`. For example:
```

docs/en/administration/metadata/redis_best_practices.md
```diff
@@ -72,14 +72,14 @@ It should be noted that since the data of the Redis master node is asynchronousl
...
 ```
 
 :::tip
-Redis Cluster does not support multiple databases. However, it splits the key space into 16384 hash slots, and distributes the slots to several nodes. Based on Redis Cluster's [Hash Tag](https://redis.io/docs/reference/cluster-spec/#hash-tags) feature, JuiceFS adds `{DB}` before all file system keys to ensure they will be hashed to the same hash slot, assuring that transactions can still work. Besides, one Redis Cluster can serve for multiple JuiceFS file systems as long as they use different db numbers.
+Redis Cluster does not support multiple databases. However, it splits the key space into 16384 hash slots, and distributes the slots to several nodes. Based on Redis Cluster's [Hash Tag](https://redis.io/docs/reference/cluster-spec/#hash-tags) feature, JuiceFS adds `{DB}` before all file system keys to ensure they will be hashed to the same hash slot, assuring that transactions can still work. If you set `?prefix=<prefix>`, JuiceFS will use that value as the hash tag instead. Besides, one Redis Cluster can serve for multiple JuiceFS file systems as long as they use different db numbers or explicit prefixes.
```
CaitinChen, line 82:
```suggestion
Redis Cluster does not support multiple databases. However, it splits the key space into 16,384 hash slots and distributes them across nodes. Based on Redis Cluster's [hash tag](https://redis.io/docs/reference/cluster-spec/#hash-tags) feature, JuiceFS adds `{DB}` to the beginning of each key to ensure that all keys of a JuiceFS file system are placed in the same hash slot, assuring that transactions can still work. If you set `?prefix=<prefix>`, JuiceFS uses that value as the hash tag instead. Moreover, a single Redis Cluster can serve multiple JuiceFS file systems as long as they use different database numbers or explicit prefixes.
```

docs/en/reference/how_to_set_up_metadata_engine.md
```diff
@@ -32,12 +32,13 @@ When you need to migrate between two types of metadata engines, you can use this
 JuiceFS requires Redis version 4.0 and above. Redis Cluster is also supported, but in order to avoid transactions across different Redis instances, JuiceFS puts all metadata for one file system on a single Redis instance.
 
 :::tip Redis Cluster Key Prefix
-When using Redis Cluster, the database number in the URL is used as a **key prefix** rather than for actual database selection (since Redis Cluster only supports database 0). The prefix format is `{N}` (e.g., `{1}`, `{2}`), which uses Redis hash tags to ensure all keys for one volume are routed to the same slot. This allows multiple JuiceFS file systems to share a single Redis Cluster:
+When using Redis Cluster, the database number in the URL is used as a **key prefix** rather than for actual database selection (since Redis Cluster only supports database 0). The prefix format is `{N}` (e.g., `{1}`, `{2}`), which uses Redis hash tags to ensure all keys for one volume are routed to the same slot. You can also set an explicit prefix using `?prefix=<value>`; JuiceFS will hash-tag it automatically in cluster mode. This allows multiple JuiceFS file systems to share a single Redis Cluster:
```
CaitinChen, line 35:
```suggestion
When using Redis Cluster, the database number specified in the connection URL is used as a **key prefix** rather than for actual database selection (since Redis Cluster only supports database 0). The prefix format is `{N}` (e.g., `{1}`, `{2}`), which uses Redis hash tags to ensure that all keys belonging to the same volume are routed to the same hash slot. You can also specify an explicit prefix with `?prefix=<value>`; JuiceFS will automatically wrap it in hash tags when running in cluster mode. This allows multiple JuiceFS file systems to share a single Redis Cluster:
```

docs/en/reference/how_to_set_up_metadata_engine.md
```diff
@@ -77,6 +78,7 @@ Where `[]` enclosed are optional and the rest are mandatory.
 - `<username>` is introduced after Redis 6.0 and can be ignored if there is no username, but the `:` colon in front of the password needs to be kept, e.g. `redis://:<password>@<host>:6379/1`.
 - The default port number on which Redis listens is `6379`, which can be left blank if the default port number is not changed, e.g. `redis://:<password>@<host>/1`.
 - Redis supports multiple [logical databases](https://redis.io/commands/select), please replace `<db>` with the actual database number used.
+- To prefix all JuiceFS metadata keys within a Redis database, add `?prefix=<value>` to the metadata URL. In Redis Cluster, the prefix is automatically converted into a Redis hash tag so all keys stay in one slot. Avoid Redis glob metacharacters such as `*`, `?`, and `[` in custom prefixes.
```
CaitinChen, line 81:
```suggestion
- To add a prefix to all JuiceFS metadata keys stored in a Redis database, include `?prefix=<value>` in the metadata URL. In Redis Cluster, the prefix is automatically converted into a Redis hash tag so all keys stay in the same slot. Avoid using Redis glob metacharacters such as `*`, `?`, and `[` in custom prefixes.
```

gtratnack-alteryx:
After discussion on the issue, closing PR. 

https://github.com/juicedata/juicefs/issues/6850

PR #6869, opened 2026-04-09
deps: bump badger and sql drivers

solracsf:
Also run `govulncheck` and fixed affected dependencies


zhijian-pro:
For each third-party library, we suggest upgrading with a separate pr

PR #6875, opened 2026-04-10
pkg/sync: ignore temp juicefs key

zxh326:


pkg/sync/sync.go
```diff
@@ -1434,8 +1434,20 @@ func parseIncludeRules(args []string) (rules []rule) {
 	return
 }
 
+func isJuiceFSTempKey(key string) bool {
```
zhijian-pro: Move next to the `TmpFilePath` function in file.go?

jiefenghuang: @copilot move this

jiefenghuang (approved):

PR #6884, opened 2026-04-10
pkg/chunk: use non-blocking send in scanDelayedStaging

solracsf:
The blocking send on `pendingCh` would stall `scanDelayedStaging` indefinitely when all upload workers are occupied (e.g. object store temporarily unreachable). During the stall the item's uploading flag stays true, preventing any other path from re-queuing it, and future scans are blocked behind the goroutine.

`addDelayedStaging` already guards this with select/default; apply the same pattern here: reset uploading to false on a full channel so the next scan iteration retries.


jiefenghuang:
This seems unnecessary. If all workers are blocked, then there is no need for scanDelayedStaging to continue scanning. Unlike addDelayedStaging, we dont' want to block other process, and scanDelayedStaging is an independent background task.

solracsf:
Good point, closing.

PR #6885, opened 2026-04-10
pkg/fuse: invalidate kernel inode cache on write-close

solracsf:
JuiceFS correctly invalidated its own internal attribute state after a write-close (`invalidateAttr`), but never sent `FUSE_NOTIFY_INVAL_INODE` to the kernel. Within the `--attr-cache`/`--entry-cache` timeout window, the kernel could serve stale page-cache data to any reader making the write invisible.

This adds an `InvalidateInode` callback to VFS, symmetric to the existing `InvalidateEntry`, wired in fuse.go's `Serve()` to `fssrv.InodeNotify`. It is called from `Release()` when the handle was opened for writing, forcing the kernel to drop its cached pages and attributes for that inode immediately on close.

The result is that write visibility on FUSE mounts is now consistent: after a writable file handle is closed, any subsequent reader on the same mount is guaranteed to go through JuiceFS and see current data, regardless of cache timeout settings.

### DIFF
--- a/pkg/fuse/fuse.go
+++ b/pkg/fuse/fuse.go
@@ -534,6 +534,14 @@ func Serve(v *vfs.VFS, options string, xattrs, ioctl bool) error {
 		v.InvalidateEntry = func(parent Ino, name string) syscall.Errno {
 			return syscall.Errno(fssrv.EntryNotify(uint64(parent), name))
 		}
+		v.InvalidateInode = func(ino Ino) syscall.Errno {
+			return syscall.Errno(fssrv.InodeNotify(uint64(ino), -1, 0))
+		}
+	} else if runtime.GOOS == "darwin" {
+		v.InvalidateInode = func(ino Ino) syscall.Errno {
+			go fssrv.InodeNotify(uint64(ino), -1, 0)
+			return 0
+		}
 	}
 
 	fsserv = fssrv
--- a/pkg/vfs/vfs.go
+++ b/pkg/vfs/vfs.go
@@ -673,6 +673,9 @@ func (v *VFS) Release(ctx Context, ino Ino, fh uint64) {
 			if f.writer != nil {
 				_ = f.writer.Flush(ctx)
 				v.invalidateAttr(ino)
+				if v.InvalidateInode != nil {
+					_ = v.InvalidateInode(ino)
+				}
 			}
 			if locks&1 != 0 {
 				_ = v.Meta.Flock(ctx, ino, fowner^fh, F_UNLCK, false)
@@ -1226,6 +1229,7 @@ type VFS struct {
 	Meta            meta.Meta
 	Store           chunk.ChunkStore
 	InvalidateEntry func(parent meta.Ino, name string) syscall.Errno
+	InvalidateInode func(ino meta.Ino) syscall.Errno
 	UpdateFormat    func(*meta.Format)
 	reader          DataReader
 	writer          DataWriter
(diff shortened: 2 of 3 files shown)


jiefenghuang:
Hi @solracsf , this has already been addressed in #6326 — the invalidate is performed during open.
Have you encountered a specific issue, Could you provide a reproducible example?

solracsf:
#6326 invalidates the page cache inside `fileSystem.Open`. That only fires when the kernel actually sends an OPEN op to FUSE. It helps a reader that is about to open a file, but it does nothing for a reader whose FD is already open when someone else writes. The write-close path `VFS.Release` had no hook into `InodeNotify` at all before this patch.

Minimal reproducer against a clean main (1 mount, two handles against the same inode):

```
juicefs mount $META /mnt/jfs
echo v1 > /mnt/jfs/f

# reader: open once, read once to populate the kernel page cache, keep the FD
exec 9< /mnt/jfs/f
dd if=/dev/fd/9 bs=4 count=1 2>/dev/null    # -> v1

# writer (separate handle) overwrites and closes
python3 -c "open('/mnt/jfs/f','w').write('v2\n')"

# reader reads again from the SAME fd (no new open() -> #6326 never fires)
dd if=/dev/fd/9 bs=4 count=1 skip=0 2>/dev/null   # -> still v1 on main
```

Without `-o writeback_cache`, the offending pages come from the kernel inode's page cache; the writer's Release is the only place we can tell the kernel to drop them before the reader's next `read()`.

With this patch the second `dd` returns `v2`.

jiefenghuang:
> #6326 invalidates the page cache inside `fileSystem.Open`. That only fires when the kernel actually sends an OPEN op to FUSE. It helps a reader that is about to open a file, but it does nothing for a reader whose FD is already open when someone else writes. The write-close path `VFS.Release` had no hook into `InodeNotify` at all before this patch.
> 
> Minimal reproducer against a clean main (1 mount, two handles against the same inode):
> 
> ```
> juicefs mount $META /mnt/jfs
> echo v1 > /mnt/jfs/f
> 
> # reader: open once, read once to populate the kernel page cache, keep the FD
> exec 9< /mnt/jfs/f
> dd if=/dev/fd/9 bs=4 count=1 2>/dev/null    # -> v1
> 
> # writer (separate handle) overwrites and closes
> python3 -c "open('/mnt/jfs/f','w').write('v2\n')"
> 
> # reader reads again from the SAME fd (no new open() -> #6326 never fires)
> dd if=/dev/fd/9 bs=4 count=1 skip=0 2>/dev/null   # -> still v1 on main
> ```
> 
> Without `-o writeback_cache`, the offending pages come from the kernel inode's page cache; the writer's Release is the only place we can tell the kernel to drop them before the reader's next `read()`.
> 
> With this patch the second `dd` returns `v2`.

However, JuiceFS currently only aims to guarantee close-to-open (CTO) consistency, which is a trade-off between performance and consistency. If we were to take opened handlers into account, there would be many more scenarios to consider. For now, we think this is not necessary to merge. I can add a label and revisit it later.

solracsf:
Understood — that's a reasonable line to hold. My reproducer deliberately relies on an **already-open FD**, which is exactly the regime CTO doesn't cover, so, under that contract, the current behavior is correct.

jiefenghuang:
We will temporarily close this PR. If we decide to strengthen consistency guarantees in the future, we will re-evaluate it, with broader scenario coverage and additional testing required.

solracsf:
No problem!

PR #6887, opened 2026-04-11
gateway: implement conditional matches

### LINKED ISSUE #6460: s3 gateway does not support CONDITIONAL WRITE
It seems juicefs s3 gateway does not support CONDITIONAL WRITE (detected by zerofs over s3). Maybe the problem is minio used by juicefs is too old?

stevefan1999-personal:
Fixes https://github.com/juicedata/juicefs/issues/6460.

This is needed for ZeroFS to run on JuiceFS's S3 gateway.

### DIFF
--- a/cmd/gateway.go
+++ b/cmd/gateway.go
@@ -233,6 +233,7 @@ func gateway(c *cli.Context) error {
 			},
 		},
 	}
+	registerGatewayConditionalMiddleware()
 	return app.Run(args)
 }
 
--- a/cmd/gateway_minio_hook.go
+++ b/cmd/gateway_minio_hook.go
@@ -1,0 +1,32 @@
+//go:build !nogateway
+// +build !nogateway
+
+package cmd
+
+import (
+	"sync"
+
+	"github.com/gorilla/mux"
+	jfsgateway "github.com/juicedata/juicefs/pkg/gateway"
+
+	_ "unsafe"
+)
+
+//go:linkname minioGlobalHandlers github.com/minio/minio/cmd.globalHandlers
+var minioGlobalHandlers []mux.MiddlewareFunc
+
+var registerGatewayConditionalMiddlewareOnce sync.Once
+
+func registerGatewayConditionalMiddleware() {
+	registerGatewayConditionalMiddlewareOnce.Do(func() {
+		// The juicedata/minio fork currently wires request middleware through the
+		// unexported cmd.globalHandlers slice, which configureServerHandler()
+		// consumes before requests reach the object layer. There is no public hook
+		// for JFS to register a gateway-specific middleware at that point, so keep
+		// this linkname in sync with upstream cmd/routers.go when bumping MinIO.
+		minioGlobalHandlers = append(
+			[]mux.MiddlewareFunc{mux.MiddlewareFunc(jfsgateway.ConditionalRequestMiddleware)},
+			minioGlobalHandlers...,
+		)
+	})
+}
--- a/cmd/gateway_minio_hook_test.go
+++ b/cmd/gateway_minio_hook_test.go
@@ -1,0 +1,49 @@
+//go:build !nogateway
+// +build !nogateway
+
+package cmd
+
+import (
+	"net/http"
+	"reflect"
+	"sync"
+	"testing"
+
+	"github.com/gorilla/mux"
+	jfsgateway "github.com/juicedata/juicefs/pkg/gateway"
+)
+
+func TestRegisterGatewayConditionalMiddleware(t *testing.T) {
+	savedHandlers := append([]mux.MiddlewareFunc(nil), minioGlobalHandlers...)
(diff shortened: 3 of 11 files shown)


stevefan1999-personal:
@zhijian-pro any suggestions?

stevefan1999-personal:
I've tested it with Pulumi as well, which also needs conditional match (If-None-Match) and it also seems to work well too

zhijian-pro:
Why is there no implementation in MinIO? https://github.com/juicedata/minio/tree/new-base

zhijian-pro:
This is support for s3 syntax. I think it's more appropriate to implement it in the [minio](https://github.com/juicedata/minio/tree/new-base) repository
There is already a similar implementation in MinIO.
We are currently using our own forked version, the branch is new-base.

stevefan1999-personal:
@zhijian-pro Okay, let me see how to move it there. But this PR hooks into Minio as well so that it is effectively the same. A small problem is that we probably cannot use the still-existing Minio's code at the moment, or maybe looking at the source code before the license change albeit there would probably a lot of merge conflicts, or you need to prompt your way out such that it does not use any Minio code.

zhijian-pro:
> @zhijian-pro Okay, let me see how to move it there. But this PR hooks into Minio as well so that it is effectively the same. A small problem is that we probably cannot use the still-existing Minio's code at the moment, or maybe looking at the source code before the license change albeit there would probably a lot of merge conflicts, or you need to prompt your way out such that it does not use any Minio code.

Please check this branch, https://github.com/juicedata/minio/tree/new-base, this branch is based on the Apache version minio's development, so there won't be the problem on the open source licenses. The gateway of juicefs references the code of this branch. So please based on https://github.com/juicedata/minio/tree/new-base the branch for development.

zhijian-pro:
> @zhijian-pro Okay, let me see how to move it there. But this PR hooks into Minio as well so that it is effectively the same. A small problem is that we probably cannot use the still-existing Minio's code at the moment, or maybe looking at the source code before the license change albeit there would probably a lot of merge conflicts, or you need to prompt your way out such that it does not use any Minio code.

Do not copy the implementation code of minio, you can implement it yourself.

stevefan1999-personal:
> > @zhijian-pro Okay, let me see how to move it there. But this PR hooks into Minio as well so that it is effectively the same. A small problem is that we probably cannot use the still-existing Minio's code at the moment, or maybe looking at the source code before the license change albeit there would probably a lot of merge conflicts, or you need to prompt your way out such that it does not use any Minio code.
> 
> Do not copy the implementation code of minio, you can implement it yourself.

No, what I mean is the code that predates the change to AGPL license can still freely be used. That means any code before that AGPL license switch commit is still legally Apache 2.0. If they happen to have implemented conditional read/write there, then we have a happy ending. Maybe we need to fix some bugs if there are some spec and semantic issues.

zhijian-pro:
> > > @zhijian-pro Okay, let me see how to move it there. But this PR hooks into Minio as well so that it is effectively the same. A small problem is that we probably cannot use the still-existing Minio's code at the moment, or maybe looking at the source code before the license change albeit there would probably a lot of merge conflicts, or you need to prompt your way out such that it does not use any Minio code.
> > 
> > 
> > Do not copy the implementation code of minio, you can implement it yourself.
> 
> No, what I mean is the code that predates the change to AGPL license can still freely be used. That means any code before that AGPL license switch commit is still legally Apache 2.0. If they happen to have implemented conditional read/write there, then we have a happy ending. Maybe we need to fix some bugs if there are some spec and semantic issues.

https://github.com/juicedata/minio/tree/new-base is already based on the use of Apache license last commit

stevefan1999-personal:
> > > > @zhijian-pro Okay, let me see how to move it there. But this PR hooks into Minio as well so that it is effectively the same. A small problem is that we probably cannot use the still-existing Minio's code at the moment, or maybe looking at the source code before the license change albeit there would probably a lot of merge conflicts, or you need to prompt your way out such that it does not use any Minio code.
> > > 
> > > 
> > > Do not copy the implementation code of minio, you can implement it yourself.
> > 
> > 
> > No, what I mean is the code that predates the change to AGPL license can still freely be used. That means any code before that AGPL license switch commit is still legally Apache 2.0. If they happen to have implemented conditional read/write there, then we have a happy ending. Maybe we need to fix some bugs if there are some spec and semantic issues.
> 
> [juicedata/minio@`new-base`](https://github.com/juicedata/minio/tree/new-base?rgh-link-date=2026-04-16T06%3A02%3A11Z) is already based on the use of Apache license last commit

Ah so you mean that the conditional matches have had to be implemented post-Apache-license change. That means we need to find another way around it. It shouldn't be a big deal then, given that we already have the minio hook on the juicefs side, now we can just find hunks to apply the PR straight into the minio branch without having the use the indirection hook. Is that what you think would go for greenlit?

zhijian-pro:
Yes, you can refer to the code here; this feature should be completed here.
https://github.com/juicedata/minio/blob/7875d472bc68e61e47cf9e48491da898454a1204/cmd/object-handlers-common.go#L51

jiefenghuang:
close by #7479

PR #6893, opened 2026-04-13
meta: fix stale cache of the overwritten file in rename

Xuyuchao-juice:


pkg/meta/base_test.go
jiefenghuang: defer m.Unlink(ctx, dir1, "trash_file2")

jiefenghuang: m.Unlink(ctx, dir2, "cross_trash2")

jiefenghuang (approved):

PR #6896, opened 2026-04-14
object/s3: use default credential chain in autoS3Region when accessKe…

zhijian-pro:


zhijian-pro:
object/s3: use default credential chain in autoS3Region when accessKey is empty (#6754)
(cherry picked from commit 1e9aac32c737f6e479a73658b43a7306421fd5c7)

PR #6903, opened 2026-04-15
Revert "pkg/sync: ignore temp juicefs key"

zxh326:
Reverts juicedata/juicefs#6875


jiefenghuang (approved):
TODO: clean or skip tmp file

PR #6960, opened 2026-04-24
fix(java-sdk): move null check before dereference in getFileBlockLocations

wzhallright:
Current behavior: In JuiceFileSystemImpl.getFileBlockLocations(), the null check for the 'file' parameter is placed after file.getPath() is already called in the permission check branch (line 990). This means when 'file' is null, a NullPointerException is thrown at file.getPath() before the null check on line 1004 is ever reached, making the null check dead code.


wzhallright:
@tangyoupeng Hi, Sir. Could you help review this PR?

tangyoupeng (approved):

PR #6966, opened 2026-04-27
CI: refactor rename dir stat tests with assertInodes helper

Xuyuchao-juice:
Some getDirStat calls remain unprotected and may return stale values under race conditions. This commit adds assertions to all these getDirStat invocations to ensure correctness.

### DIFF
--- a/pkg/meta/base_test.go
+++ b/pkg/meta/base_test.go
@@ -92,6 +92,16 @@ func waitCheckResult(m Meta, expect dirStat, statFn func() (*dirStat, syscall.Er
 	}
 }
 
+func assertInodes(t *testing.T, m Meta, label string, dirIno Ino, expected int64) {
+	t.Helper()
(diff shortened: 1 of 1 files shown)


pkg/meta/base_test.go
```diff
@@ -3392,212 +3337,154 @@ func testRenameDirStatWithTrash(t *testing.T, m Meta) {
...
 		m.FlushSession()
-
-		// Get the trash inode (it should be marked as trash)
 		if !trashInode.IsTrash() {
-			// File in trash has a special inode
-			trashInode = file1Inode // Try using the same inode as it might be converted
+			trashInode = file1Inode
 		}
-
 		stat1Before, _ := m.GetDirStat(ctx, dir1)
```
jiefenghuang: 这里?

jiefenghuang (approved):

PR #6986, opened 2026-04-29
Fix for any level .control

### LINKED ISSUE #6985: FUSE: user-created ".control" files collide with internal controlInode in subdirectories
**What happened**:

When a non-root user creates a file named `.control` in any directory on a JuiceFS mount,
the file is incorrectly associated with JuiceFS's internal `controlInode` 
(`0x7FFFFFFF00000002`, inode `9223372032559808514`). This internal inode is owned by
`root:root` with mode `0666`, making it impossible for non-root users to `chmod` or
execute it.

Steps that fail for non-root users:
1. `chmod +x .control` → `Operation not permitted` (EPERM)
2. `./.control` → `Permission denied` (if mode bits don't allow exec)
3. `rm .control` → `No such file or directory` (ENOENT) — the dentry persists in the
   kernel cache but the metadata engine has already removed the entry.

Additionally, affected directories exhibit a readdir/LOOKUP inconsistency:
- `ls -la` (readdir) does NOT show `.control`
- `stat .control` (LOOKUP) succeeds and returns the internal inode

This effectively breaks any application that creates `.control` files as part of its
normal workflow on JuiceFS.

A user-created `.control` file should be a regular file owned by the calling user,
with no connection to JuiceFS's internal `controlInode`. All standard file operations
(`chmod`, execute, `rm`) should work identically to any other filename.

```bash
# run this in a JuiceFS mount point as a non-root user

touch .control
stat -c 'inode=%i owner=%U mode=%a' .control
chmod +x .control
rm .control
ls -la .control
stat .control
```

In my machine:

```
[testuser@server1 test]$ touch .control
[testuser@server1 test]$ stat -c 'inode=%i owner=%U mode=%a' .control
inode=9223372032559808514 owner=root mode=666
[testuser@server1 test]$ chmod +x .control
chmod: changing permissions of '.control': Operation not permitted
[testuser@server1 test]$ rm .control
rm: remove regular empty file '.control'? y
rm: cannot remove '.control': No such file or directory
[testuser@server1 test]$ ls -la .control
-rw-rw-rw- 1 root root 0 Apr 29 23:39 .control
[testuser@server1 test]$ stat .control
  File: .control
  Size: 0               Blocks: 0          IO Block: 65536  regular empty file
Device: 33h/51d Inode: 9223372032559808514  Links: 1
Access: (0666/-rw-rw-rw-)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2026-04-29 23:39:22.000000000 +0800
Modify: 2026-04-29 23:39:22.000000000 +0800
Change: 2026-04-29 23:39:22.000000000 +0800
 Birth: -
```

- The issue is **not** related to `--writeback`, `--trash-days`, or cache settings.
  It reproduces with all combinations of these flags.
- Reproduced on both v1.3.1 (2025-12-02) and v1.4.0-beta1.
- `drop_caches` and `systemctl restart juicefs-mount` do **not** clear the ghost
  dentries — only `juicefs rmr` on the parent directory works.
- The same script on a local filesystem (ext4/xfs) works correctly — the problem is
  specific to JuiceFS FUSE.
- The `find` command cannot locate these ghost `.control` files (they are invisible
  to readdir), but `stat` on the exact path succeeds.

As ai infered, the issue is in `pkg/vfs/internal.go` lines 40-45:

```go
const (
    minInternalNode = 0x7FFFFFFF00000000
    controlInode    = minInternalNode + 2   // 0x7FFFFFFF00000002
)

var internalNodes = []*internalNode{
    {controlInode, ".control", &Attr{Mode: 0666}},   // ← global reserved name
    {logInode,     ".accesslog", &Attr{Mode: 0400}},
    {StatsInode,   ".stats",     &Attr{Mode: 0444}},
    {ConfigInode,  ".config",    &Attr{Mode: 0400}},
}
```

The `IsSpecialName(".control")` function returns `true` for this filename in **any**
directory. When a user creates a `.control` file, the FUSE name lookup can route to
the global internal `controlInode` instead of creating a proper per-directory inode.
The internal inode:
- Has a fixed inode number (`0x7FFFFFFF00000002`)
- Is owned by the FUSE daemon's UID (typically `root`)
- Has mode `0666` (no execute permission)
- Cannot be chmod'd or unlinked by non-root users

**Environment**:
- JuiceFS version: `juicefs version 1.3.1+2025-12-02.e0032b2` (also reproduced on
  v1.4.0-beta1)
- OS: RHEL 8.10 (kernel 4.18.0-553.el8_10.x86_64)
- Object storage: Self-maintained SeaweedFS 4.18 (S3-compatible, EC 10+4 on 7 HDDs,
  ~101 TiB usable). JuiceFS connects via `weed s3` gateway at `127.0.0.1:8333`.
- Metadata engine: PostgreSQL 17.9 (primary + 3 async streaming replicas).
  JuiceFS connects to primary at ~10-14 ms latency (same switch, 1GbE).
- Network: Compute nodes on 10GbE, control plane (PostgreSQL primary, SeaweedFS
  master) on 1GbE.
- Mount options: `--writeback --cache-dir /storage/cache/jfscache --cache-size
  8388608 --buffer-size 1024 --prefetch 3 --max-uploads 20 --backup-meta 3600
  --no-bgjob -o allow_other`

Yue-bin:
This may fix [#6985](https://github.com/juicedata/juicefs/issues/6985)

### DIFF
--- a/pkg/vfs/vfs.go
+++ b/pkg/vfs/vfs.go
@@ -176,7 +176,7 @@ var (
 func (v *VFS) Lookup(ctx Context, parent Ino, name string) (entry *meta.Entry, err syscall.Errno) {
 	var inode Ino
 	var attr = &Attr{}
-	if parent == rootID || name == internalNodes[0].name { // 0 is the control file
+	if parent == rootID {
 		n := getInternalNodeByName(name)
 		if n != nil {
 			entry = &meta.Entry{Inode: n.inode, Attr: n.attr}


Yue-bin:
It works well in my cluster after this patch to branch release-1.3😊

jiefenghuang:
This change would break the current usage pattern. For example, juicefs info /tmp/jfs/d1/f1 would fail, where /tmp/jfs is the mount point. The original reason for adding this was to allow internal operations to be executed from any directory.

PR #6998, opened 2026-05-07
cmd/mount: improve handing when process hung

Xuyuchao-juice:
rel https://github.com/juicedata/juicefs/issues/6691

When the mount process becomes unresponsive and FUSE requests are blocked, the value of `/sys/fs/fuse/connections/%d/waiting` should be used to determine whether there are pending FUSE requests and whether it is necessary to send an abort signal. The previous approach of matching hardcoded strings in the stack trace could not cover all required scenarios.

### DIFF
--- a/cmd/mount_unix.go
+++ b/cmd/mount_unix.go
@@ -57,6 +57,8 @@ import (
 
 var mountPid int
 
+const fuseConnectionsPath = "/sys/fs/fuse/connections"
+
 func showThreadStack(agentAddr string) {
 	if agentAddr == "" {
 		return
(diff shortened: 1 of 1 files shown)


cmd/mount_unix.go
```diff
@@ -95,25 +97,31 @@ func killMountProcess(pid int, dev uint64, lastActive *int64) {
...
-				break
+	if runtime.GOOS != "linux" || dev == 0 {
+		return
+	}
+	conn := devMinor(dev)
+	data, err := os.ReadFile(filepath.Join(fuseConnectionsPath, strconv.FormatUint(uint64(conn), 10), "waiting"))
+	if err == nil {
+		waiting, err := strconv.Atoi(strings.TrimSpace(string(data)))
+		if err == nil && waiting > 0 {
+			fuseMu.Lock()
```
jiefenghuang, line 125: 日志还是打出来

jiefenghuang (approved):

PR #7003, opened 2026-05-08
feat: add p2p-warmup subcommand

kyungwan-nam:
# Peer-to-peer warmup 

### 1.1 Problem

The existing `juicefs warmup` command runs on a single node, scanning the meta
engine and pulling blocks from object storage into the local disk cache. When
the same dataset has to be warmed on N instances of a distributed deployment
(Kubernetes StatefulSet, Ray cluster, etc.), two inefficiencies compound.

#### Object storage request amplification on cold start

When N nodes warm the same dataset simultaneously, every block is downloaded N
times.

- 100 GiB model file × 12 nodes = **1,200 GiB** of object storage egress
- On metered storage (S3, GCS) the cost is N×
- Even on internal storage, peers risk hitting rate limits or per-tenant
  bandwidth caps
- Job *time-to-ready* (an AI training/serving setup time) is bottlenecked
  on the per-peer storage cap — e.g. with a 100 MB/s cap, downloading a single
  100 GiB file takes 17 min 53 s, regardless of how parallel the cluster is

#### Repeated fetches when instances are added or restarted

Every time an instance joins the cluster (scale-out, node failure recovery,
rolling restart), the same data is re-fetched from object storage from scratch
— even though sibling instances already have it.

- GPU nodes added to an inference cluster, workers added to a distributed
  training job: each round pays the full storage time + cost
- For "warm cache as a service" workloads (AI inference fleets), this happens
  continuously throughout the cluster's lifetime

### 1.2 Inspiration — Naver p2pcp

[p2pcp](https://github.com/naver/p2pcp), open-sourced by Naver, is a tool for
distributing the same file from HDFS / object storage to many nodes via
peer-to-peer exchange, collapsing storage egress to roughly one node's worth
across the cluster. The core ideas:

- Each node fetches a different chunk from storage
- Each node exposes HTTP `/available` (the chunk list it currently holds) and
  `/chunk/<id>` (the chunk bytes)
- Other nodes pull missing chunks over LAN P2P instead of going to storage
- Net: cluster-wide storage egress converges to a single node's download

We hypothesized that the same design pattern (shuffled queue, availability
polling, HTTP chunk serving) would solve both problems above when applied to
JuiceFS warmup.

### 1.3 Proposal — `juicefs p2p-warmup`

Add a new subcommand `p2p-warmup` to the JuiceFS binary. It runs independently
of mount, talks directly to the meta engine and object storage, and shares the
on-disk cache directory with the mount daemon so cached blocks become
immediately useful.

We borrowed p2pcp's design pattern and adapted it for JuiceFS:

- **Manifest sharing** — to avoid hammering the meta engine N times, only the
  elected leader scans meta; the resulting block list is uploaded to object
  storage as a manifest, and followers download it. This collapses meta load
  from O(N peers × M files) to O(M files).
- **Rendezvous-hashing approximation** — collapses the partition / shuffle
  schedulers of p2pcp into a single algorithm. Every peer carries every block
  in its queue (full cover — no orphaned blocks if a peer dies), but the front
  of each peer's queue is statistically distinct, spreading concurrent storage
  hits.
- **No-communication leader election** — the lex-smallest peer address wins,
  no consensus protocol needed.
- **Shared on-disk cache directory with mount** — mount uses cached blocks
  immediately; no separate cache layout.
- **Compression-aware (lz4 / zstd)** — fetches from storage are decompressed
  before caching, matching the layout `pkg/chunk/disk_cache.go:openCacheFile`
  expects.

### 1.4 Measured performance

100 GiB single file, 12 peers (K8s StatefulSet, one pod per node),
`--download-limit 800 Mbps` (= 100 MB/s per peer):

| Scenario | Wall time | Throughput | Storage load (per peer) | vs storage-only baseline (17m 53s) |
|---|---:|---:|---:|---:|
| Cold start (12 peers in parallel) | **4m 21s** | 393 MiB/s | 11% (~45 MiB/s) | **4.1× faster** |
| Scale-out (13th peer joins) | **1m 54s** | 902 MiB/s | <1% (~8 MiB/s) | **9.4× faster** |

Cluster-wide:
- Cold start storage egress: **~89% reduction** (1,200 GiB → 132 GiB)
- Scale-out storage egress: **~99% reduction** (100 GiB → 944 MiB)

---

### 2.1 Architecture

```
   ┌────────────────────┐         ┌────────────────────────┐
   │    Meta Engine     │         │    Object Storage      │
   │  (Redis/SQL/TiKV)  │         │   chunks/...   (data)  │
   │                    │         │   p2p_warmup/  (mfst)  │
   └─────────┬──────────┘         └───────────┬────────────┘
             │                                │
             │ leader scans once,             │ all peers fetch:
             │ uploads manifest               │  • manifest (small)
             │                                │  • missing blocks
             │                                │ rate-limited per peer
             ▼                                ▼

       ┌──────────┐    ┌──────────┐         ┌──────────┐
       │ peer 1   │◀──▶│ peer 2   │◀──...──▶│ peer N   │
       │  warm    │    │  warm    │         │  warm    │
       │  cache   │    │  cache   │         │  cache   │
       └────┬─────┘    └────┬─────┘         └────┬─────┘
            │               │                    │
            ▼               ▼                    ▼
       juicefs         juicefs              juicefs
        mount           mount                mount
       (reads          (reads               (reads
        cache)          cache)               cache)

   peer ⇄ peer:  LAN P2P, unthrottled
                 /available, /block/<key>, /uuid, /status
```

Three flow paths:

1. **Meta engine** — only the leader peer scans. The result is uploaded as a
   manifest object so other peers can skip the scan entirely.
2. **Object storage** — every peer fetches only the blocks it does not
   already have, throttled by `--download-limit`. Rendezvous-hash sort makes
   each peer's queue prefix distinct, so concurrent storage hits land on
   different blocks.
3. **Peer ⇄ Peer** — what one peer fetches becomes immediately available to
   the rest over LAN (no throttling). Net effect: cluster-wide storage egress
   converges to ≈1×.

Each peer process **shares the same on-disk cache directory with the mount
daemon**, so every block warmed is consumable by mount immediately — no
intermediate format conversion needed.

### 2.2 Run flow (12 steps)

`Warmup.Run()` executes sequentially:

| # | Step | Notes |
|---|---|---|
| 1 | HTTP server listen + `scanExistingCache` | Pre-existing cache files are announced via `/available`; size-validated to skip partial-write residue |
| 2 | Initial peer discovery + self-filter (`/uuid`) | Block until `--min-peers` reached (lower-bound semantics) |
| 3 | **Block list acquisition** (see §2.4) | 4-way branch |
| 4 | Background discovery loop (3s) | Tracks pod additions/removals |
| 5 | Background availability polling (2s) | Detects dead peers within 5s |
| 6 | Build fetch queue (Rendezvous-hash sort) | Each peer sees a different order |
| 7 | Start fetch worker pool (`--threads`, default 50) | Peer-fetch first, storage fallback |
| 8 | Progress monitor (5s) | Auto-cancels at completion |
| 9 | `waitForCompletion` (CAS-based) | Wait until every block reaches a terminal state (`Done` or `Failed`) |
| 10 | Final stats (elapsed / throughput) | Two clocks: total / fetch |
| 11 | Keep-alive (`off`/`peers`/`forever`, default `peers`) | `peers`: serve until every live peer completes, then exit; `forever`: until SIGTERM/timeout |
| 12 | Graceful shutdown | bgCancel + bgWg.Wait + HTTP 10s timeout |

### 2.3 Block state machine

```
            TryDownload (CAS Pending→Downloading)        MarkDone
  Pending ────────────────────────────────────► Downloading ──────► Done
       ▲                                            │
       │            ResetToPending                  │
       └────────────────────────────────────────────┘
                   (on fetch failure)
                                                    │
                              IncrAttempts ≥ 5      │
                              or NoSuchKey           ▼
                                                  Failed (terminal)
```

State is an `atomic.Int32`. **All transitions are CAS-only**, so 50 worker
goroutines racing for the same block result in exactly one winning the
download claim. `Failed` represents permanently unfetchable blocks (e.g.
deleted by JuiceFS slice compaction mid-warmup) or retry-budget exhaustion
(`MaxBlockAttempts = 5`). `waitForCompletion` counts both `Done` and `Failed`
as terminal so the warmup never hangs on an unfetchable block.

### 2.4 Block list acquisition — 4-way branch

```
Step 3 entry
   │
   ▼
manifest sharing enabled? (--min-peers > 1 && !--disable-manifest-sharing)
   │
   ├── No → solo / debug mode: this peer scans the meta engine directly
   │
   └── Yes
        │
        ▼
    Does a valid manifest already exist in object storage? (fast-path)
        │
        ├── Yes → use it as-is (skip meta scan — critical for scale-out
        │                       and crash recovery)
        │
        └── No
             │
             ▼
         Is this peer the leader? (lex-smallest IP)
             │
             ├── Yes → meta.Resolve + upload manifest
             │
             └── No → poll for manifest (within --leader-timeout)
```

#### Manifest reuse fast-path
Every peer attempts a one-shot manifest GET *before* the leader / follower
branch decision. If a valid manifest already exists, it is reused — so a peer
joining a warmed cluster (e.g. with the lex-smallest IP, which would
otherwise mis-elect itself leader) does not redundantly re-scan the meta
engine.

### 2.5 Leader election — agreement without communication

```go
func isLeader(peers []string, selfAddr string) bool {
    for _, p := range peers {
        if p < selfAddr { return false }
    }
    return true
}
```

When every peer runs this function with the same input (the full peer set),
every peer reaches the same conclusion — no consensus protocol required.
Precondition: `--min-peers` blocks until all peers are discovered, so the
input set is consistent.

### 2.6 Manifest collision protection (`--manifest-name`)

When operators name a manifest with `--manifest-name foo` for human
readability, two unrelated warmups that happen to share the name must not
silently corrupt each other's data. Two-layer defense:

- **Leader pre-check** — before PUT, the leader does a GET. If a manifest
  exists with matching paths/blocksize/hashprefix, the upload is skipped
  (idempotent crash recovery). Mismatches return a *refusing to overwrite*
  error so the operator notices.
- **Follower post-check** — after downloading the manifest, the follower
  validates that paths/blocksize/hashprefix match its own request before
  using the block list.

When `--manifest-name` is omitted, the key is derived as a content-addressable
SHA-256 of the warmup args, so peers running the same warmup converge on the
same key without any shared name.

### 2.7 Fetch order distribution — Rendezvous-hashing approximation

```go
sort.Slice(pending, func(i, j int) bool {
    return hashKeyForPeer(pending[i].Key, myAddr) < hashKeyForPeer(pending[j].Key, myAddr)
})
// hashKeyForPeer = FNV-1a(peerAddr || 0x00 || blockKey)
```

| Property | Guarantee |
|---|---|
| Every peer carries every block in its queue | full cover → no orphaned blocks if a peer dies |
| Each peer's sort order is distinct | queue prefixes diverge → concurrent storage hits land on different blocks |
| Deterministic per (block-set, peer) | restarts and debugging see the same queue |

`peerAddr` is fed into the hash *first* because FNV-1a's avalanche on small
trailing changes is weak: feeding it last would let near-identical peer
addresses (e.g. `10.0.0.10` vs `10.0.0.11`) produce correlated orderings,
weakening the spread.

### 2.8 Fetch priority

For each block a worker handles:

```
1. Local cache hit (size match) → skip
2. A peer holds it (availability tracker hit) → HTTP GET /block/<key>
   - success → cache + MarkLocal (FromPeers++)
   - failure → fall through to storage
3. Object storage download (throttled by --download-limit)
```

Peer-to-peer transfers are unthrottled — they should run at LAN speed.
`--download-limit` applies only to the storage path.

### 2.9 Compression support (`lz4` / `zstd`)

JuiceFS storage and cache layouts are asymmetric:
- **Object storage**: compressed bytes
- **Disk cache**: decompressed bytes (mount's `openCacheFile` validates the
  on-disk size against the *logical* size encoded in the key)

`FetchFromStorage` decompresses with `compress.Compressor` before caching, so
mount accepts every block p2p-warmup writes. For uncompressed volumes (the
common default), the noOp compressor short-circuits to a zero-copy path —
`io.ReadAll`'s buffer is returned directly without an extra alloc + memcpy.

### 2.10 Failure handling

| Situation | Handling |
|---|---|
| Storage `NoSuchKey` / `not found` | Abandon immediately — retrying won't bring back a deleted object (e.g. compacted slice) |
| Transient errors (timeout, connection refused, 5xx) | Retry up to 5× then abandon |
| Peer fetch failure | Immediate fall-through to storage |
| 3 consecutive availability-poll failures from a peer | Treat peer as dead, drop its entries from the availability tracker |
| Peer disappears from discovery (DNS / static list) | Permanent removal |
| `--cache-size` cap reached | Skip the block (not cached), count it as `skipped`, and mark it Done so warmup never hangs on the cap |
| `keep-alive=peers`: a peer crashes / leaves mid-wait | Treated as complete so one dead peer cannot block shutdown; `--keep-alive-timeout` is an upper bound |

### 2.11 HTTP API (each peer exposes on `:19090`)

| Endpoint | Method | Response |
|---|---|---|
| `/uuid` | GET | `{"uuid":"..."}` — used for self-identification |
| `/status` | GET | `{"completed":bool, "progress":{total,done}, "role", "peers", "sources":{from_peers,from_storage,from_cache,skipped,failed}, "bytes":{from_peers,from_storage}}` — readiness-probe + observability |
| `/available` | GET | `{"blocks":[...], "updated_at":<ms>}` |
| `/available?since=<ms>` | GET | only blocks with `updated_at > since` (incremental polling) |
| `/block/<key>` | GET | binary block data, 404 if not in cache |

Self-identification: `filterSelf` calls `/uuid` on every discovered peer; on
UUID match, the matching address is recorded as `externalAddr` (the address
this peer is reachable at from the cluster's perspective). That address seeds
both leader election and the queue ordering — using listenAddr alone (which
may be `0.0.0.0:port`) would make every peer compute the same identity and
break agreement.

### 2.12 Object storage layout

```
<bucket>/
├── chunks/...                      ← JuiceFS data blocks (existing)
└── p2p_warmup/                     ← p2p-warmup prefix (new)
    ├── manifest-<sha256>.json.gz   ← when --manifest-name is unset (content-addressable)
    └── <name>.json.gz              ← when --manifest-name=<name> is set
```

Operators can attach a lifecycle policy to the `p2p_warmup/` prefix (e.g.
expire after 7 days) for automatic cleanup.

### 2.13 Code layout

```
cmd/
  p2p_warmup.go          # CLI entry point (urfave/cli command)

pkg/
  p2p/
    types.go             # Block, BlockState, Peer, WarmupConfig, KeepAliveMode/ParseKeepAliveMode
    warmup.go            # Run() 12-step flow, leader election, availability polling, keep-alive modes
    server.go            # HTTP /uuid /status /available /block/<key>
    discovery.go         # DNS A/SRV/static dedup + RunLoop
    resolver.go          # MetaResolver: paths → blocks via meta engine
    manifest.go          # Manifest, ManifestKey, Validate
    manifest_store.go    # uploadManifest, downloadManifest, tryFetchValidManifest, DeleteManifest
    scheduler.go         # OrderForPeer (FNV-1a), FetchQueue
    fetcher.go           # FetchBlock (peer→storage fallback, cache-size cap), Workers, PollAvailability, PollCompleted
    availability.go      # AvailabilityTracker (local append-only, remote map)
```

### 2.14 Reused existing packages

| Package | Usage |
|---|---|
| `pkg/meta` | `Meta.Resolve` / `Meta.Read` / `Meta.Lookup` / `Meta.Readdir` |
| `pkg/chunk` | `BlockKey`, `SliceBlockKeys` (block key derivation) |
| `pkg/object` | `ObjectStorage` (storage backend abstraction) |
| `pkg/compress` | `Compressor` (decompression on storage fetch) |
| `pkg/utils` | `Duration`, `ParseMbps` |

The existing `juicefs warmup` command is unchanged. `cmd/main.go` gets one
line to register the new subcommand.

---

### 3.1 CLI signature

```
juicefs p2p-warmup [options] META_URL PATH [PATH ...]
```

| Flag | Default | Description |
|---|---|---|
| `--peers-dns-a` | — | DNS A record FQDN (intended for K8s headless services) |
| `--peers-dns-srv` | — | DNS SRV record |
| `--peers-list` | — | Static `host:port,...` list |
| `--listen` | `:19090` | HTTP server listen address |
| `--min-peers` | `0` | Minimum total peer count (incl. self) to wait for; more peers proceed immediately. ≥ 2 enables manifest sharing |
| `--manifest-name` | — | Manifest object basename. All peers must pass the same value |
| `--delete-manifest` | `false` | Delete the manifest matching the args (paths, `--manifest-name`) and exit without warming (maintenance) |
| `--leader-timeout` | `10m` | Follower timeout for leader's manifest |
| `--disable-manifest-sharing` | `false` | Debugging escape hatch — every peer scans meta independently |
| `--discovery-interval` | `3s` | DNS re-resolve interval |
| `--availability-interval` | `2s` | Per-peer `/available` polling interval |
| `--threads`, `-p` | `50` | Concurrent fetch workers |
| `--file`, `-f` | — | Path list file |
| `--keep-alive` | `peers` | Post-warmup behavior: `off` (exit when local fetch done), `peers` (serve until all peers complete, then exit), `forever` (serve until SIGTERM/SIGINT) |
| `--keep-alive-timeout` | `0` | Upper bound on the keep-alive phase (0 = no bound); applies to `peers`/`forever`, ignored when `off` |
| `--cache-dir` | from format | JuiceFS cache directory |
| `--cache-size` | `100G` | Hard cap on total cached bytes at cache-dir (e.g. 100G, 50000 = 50000 MiB; 0 = unlimited). Counts pre-existing files |
| `--download-limit` | `0` | Object-storage download cap (Mbps per peer; 0 = unlimited). P2P transfers unaffected. |

### 3.2 Single node (solo mode)

Functionally equivalent to the existing `juicefs warmup`:

```bash
juicefs p2p-warmup redis://localhost:6379/1 /models/llama-7b
```

Without `--min-peers`, the command runs in solo mode — manifest sharing is
disabled and the single node scans meta + downloads all blocks itself.

### 3.3 Static peer list

```bash
juicefs p2p-warmup \
  --peers-list 10.0.0.1:19090,10.0.0.2:19090,10.0.0.3:19090 \
  --min-peers 3 \
  --manifest-name llama-7b-warmup \
  redis://redis:6379/1 /models/llama-7b
```

### 3.4 Kubernetes deployment (StatefulSet + Service)

For reliable cold start, three layers must be configured together — Service,
StatefulSet, and container args.

#### Headless Service

```yaml
apiVersion: v1
kind: Service
metadata:
  name: p2p-warmup
  namespace: default
  labels: { app: p2p-warmup }
spec:
  clusterIP: None                 # Headless service → DNS A returns all pod IPs
  publishNotReadyAddresses: true  # Pod IPs visible in DNS before readinessProbe passes
                                  # → discovery succeeds at boot, virtually eliminating split-view race
  selector: { app: p2p-warmup }
  ports:
    - { name: p2p, port: 19090, targetPort: 19090 }
```

#### StatefulSet

```yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: p2p-warmup
spec:
  podManagementPolicy: Parallel   # Boot all pods at once (required for --min-peers synchronization)
  replicas: 12
  serviceName: p2p-warmup
  selector: { matchLabels: { app: p2p-warmup } }
  template:
    metadata: { labels: { app: p2p-warmup } }
    spec:
      affinity:
        podAntiAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            - labelSelector: { matchLabels: { app: p2p-warmup } }
              topologyKey: kubernetes.io/hostname  # one pod per node — isolate NIC bandwidth
      containers:
        - name: warmup
          image: juicedata/mount:<tag>
          args:
            - juicefs
            - p2p-warmup
            - --peers-dns-a
            - p2p-warmup.default.svc.cluster.local
            - --min-peers
            - "12"                  # match replicas
            - --manifest-name
            - llama-7b-v1           # cluster-unique to prevent collisions
            - --leader-timeout
            - "2m"
            - --threads
            - "50"
            - --keep-alive
            - peers                 # serve until all peers complete, then exit
            - --download-limit
            - "800"                 # 800 Mbps = 100 MB/s per peer
            - redis://redis:6379/1
            - /models/llama-7b
          readinessProbe:
            httpGet: { path: /status, port: 19090 }
          livenessProbe:
            httpGet: { path: /status, port: 19090 }
          volumeMounts:
            - { mountPath: /var/jfsCache, name: cache }
      volumes:
        - name: cache
          hostPath: { path: /var/jfsCache, type: DirectoryOrCreate }
```

#### Operational notes

1. Without `publishNotReadyAddresses: true`, pods are absent from DNS until
   their readinessProbe passes — and p2p-warmup pods cannot become ready
   without first discovering each other. The cluster deadlocks.
2. Without `podManagementPolicy: Parallel`, pods boot sequentially, breaking
   `--min-peers`-based leader election timing.
3. `podAntiAffinity` (one pod per node) isolates NIC bandwidth for P2P.
4. `--manifest-name` must be cluster-unique — accidentally sharing it across
   unrelated warmups would cause cross-contamination (caught by the
   collision-protection logic, but still aborts the affected warmups).
5. The default `--keep-alive peers` exits once the current cohort finishes;
   use `--keep-alive forever` to keep nodes serving so follow-up workloads
   (inference pods, etc.) can join later and fetch entirely over LAN.

### 3.5 Object storage lifecycle policy

Manifest objects accumulate under the `p2p_warmup/` prefix. AWS S3 / NCP /
MinIO recommended: a 7-day expiration rule.

```json
{
  "Rules": [{
    "ID": "juicefs-p2p-warmup-cleanup",
    "Status": "Enabled",
    "Filter": { "Prefix": "p2p_warmup/" },
    "Expiration": { "Days": 7 }
  }]
}
```

For long-running keep-alive deployments, set the lifecycle TTL longer than
the expected late-joiner interval (or skip lifecycle entirely).

### 3.6 Operational logs

```
INFO  P2P server listening on 10.0.0.10:8080 (uuid=a1b2...)
INFO  cache-size cap: 100 GiB (pass --cache-size 0 for unlimited)
INFO  found existing manifest: 6400 blocks (...); skipping resolve   ← scale-out fast-path
INFO  discovered 11 peers (excluding self)
INFO  peer 10.0.0.21:19090 joined discovery                         ← late joiner under a low --min-peers
INFO  progress: 1280/6400 blocks done, 0 failed, 5120 pending; transferred 5.0 GiB peers (180 MiB/s), 1.2 GiB storage (45 MiB/s)
...
INFO  warmup complete: 6400 blocks (5692 from peers, 708 from storage, 0 from cache, 0 skipped, 0 failed)
INFO  bytes transferred: 89 GiB from peers, 11 GiB from storage
INFO  elapsed: 4m16.145s total, 4m05.019s fetch; average throughput: 400 MiB/s total, 418 MiB/s fetch
INFO  keep-alive=peers: serving blocks until all peers complete
INFO  all peers complete; shutting down
```

`elapsed` reports two clocks:
- **total** — `Run()` entry through Step 10 (includes peer discovery wait
  and manifest wait — the operator-visible wall clock)
- **fetch** — block list acquired through Step 10 (the fetch engine itself —
  fair for peer-to-peer comparison)

Operator-signal warnings:

```
WARN  block "..." abandoned: not found in object storage (likely deleted by compaction): ...
       → data was modified during warmup, investigate

WARN  block "..." abandoned after 5 attempts: ... i/o timeout
       → transient storage / network failure
```

---

## 4. Compatibility and migration

- The existing `juicefs warmup` command is unchanged (this is a separate
  subcommand).
- Mount and format are unchanged — p2p-warmup follows the same
  `openCacheFile` validation contract.
- Apart from a one-line export of `BlockKey` / `SliceBlockKeys` from
  `pkg/chunk`, no existing code is modified.
- Default behavior is solo mode (manifest sharing disabled). Multi-peer
  features activate only when the operator passes `--min-peers >= 2`.

## 5. References

- [Naver p2pcp](https://github.com/naver/p2pcp) — design inspiration
- [Rendezvous Hashing (Wikipedia)](https://en.wikipedia.org/wiki/Rendezvous_hashing)
  — algorithmic background

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -80,6 +80,7 @@ func Main(args []string) error {
 			cmdObjbench(),
 			cmdMdtest(),
 			cmdWarmup(),
+			cmdP2PWarmup(),
 			cmdRmr(),
 			cmdSync(),
 			cmdDebug(),
--- a/cmd/p2p_warmup.go
+++ b/cmd/p2p_warmup.go
@@ -58,6 +58,7 @@ Examples:
 			&cli.BoolFlag{Name: "disable-manifest-sharing", Usage: "force every peer to resolve metadata independently (escape hatch for debugging)"},
 			&cli.StringFlag{Name: "leader-timeout", Value: "10m", Usage: "follower max wait for the leader's manifest before erroring out"},
 			&cli.StringFlag{Name: "manifest-name", Usage: "manifest object basename under p2p_warmup/ (default: content-addressable hash). All peers must pass the same value to agree on the key."},
+			&cli.BoolFlag{Name: "delete-manifest", Usage: "delete the manifest matching the other flags (paths, --manifest-name) from object storage and exit without warming"},
 			// General flags
 			&cli.IntFlag{Name: "threads", Aliases: []string{"p"}, Value: 50, Usage: "concurrent fetch workers"},
 			&cli.StringFlag{Name: "file", Aliases: []string{"f"}, Usage: "file containing a list of paths"},
@@ -113,6 +114,19 @@ func p2pWarmup(c *cli.Context) error {
 	blob, err := createStorage(*format)
 	if err != nil {
 		return err
+	}
+
+	// Maintenance mode: delete the manifest matching these args and exit.
+	// Done before any warmup wiring so a typo doesn't accidentally start
+	// the heavy machinery.
+	if c.Bool("delete-manifest") {
+		name := strings.TrimSpace(c.String("manifest-name"))
+		key, err := p2p.DeleteManifest(context.Background(), blob, paths, format.BlockSize*1024, format.HashPrefix, name)
+		if err != nil {
+			return err
+		}
+		logger.Infof("deleted manifest %q", key)
+		return nil
 	}
 
 	// Determine cache directory.
--- a/cmd/p2p_warmup.go
+++ b/cmd/p2p_warmup.go
@@ -64,7 +64,7 @@ Examples:
 			&cli.BoolFlag{Name: "keep-alive", Usage: "keep serving blocks to peers after warmup completes (until SIGTERM/SIGINT)"},
 			&cli.StringFlag{Name: "keep-alive-timeout", Value: "0", Usage: "auto-shutdown after this duration in keep-alive mode (0=wait for SIGTERM); ignored without --keep-alive"},
 			&cli.StringFlag{Name: "cache-dir", Usage: "cache directory (default: from volume format)"},
-			&cli.Int64Flag{Name: "download-limit", Usage: "object-storage download bandwidth cap in Mbps per peer (0=unlimited); peer-to-peer transfers are not throttled"},
+			&cli.StringFlag{Name: "download-limit", Usage: "bandwidth limit for download from object storage in Mbps per peer (e.g. 800 or 1G); peer-to-peer transfers are not throttled"},
 		},
 	}
 }
--- a/cmd/p2p_warmup.go
+++ b/cmd/p2p_warmup.go
@@ -1,0 +1,182 @@
+/*
+ * JuiceFS, Copyright 2024 Juicedata, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"bufio"
+	"context"
+	"os"
+	"os/signal"
+	"path"
+	"runtime"
+	"strings"
+	"syscall"
+
+	"github.com/juicedata/juicefs/pkg/meta"
+	"github.com/juicedata/juicefs/pkg/p2p"
+	"github.com/juicedata/juicefs/pkg/utils"
+	"github.com/urfave/cli/v2"
+)
+
+func cmdP2PWarmup() *cli.Command {
+	return &cli.Command{
+		Name:      "p2p-warmup",
+		Action:    p2pWarmup,
+		Category:  "TOOL",
+		Usage:     "Build cache using peer-to-peer block exchange",
+		ArgsUsage: "META-URL PATH ...",
+		Description: `Warm up cache across multiple nodes using P2P block exchange.
+Each node fetches different blocks from object storage and shares them with peers,
+reducing object storage load to 1/N across N peers.
+
+Examples:
+  juicefs p2p-warmup --peers-dns-a my-svc.ns.svc.cluster.local redis://redis:6379/1 /models/llama
+  juicefs p2p-warmup --peers-list 10.0.0.1:19090,10.0.0.2:19090 tikv://tikv:2379 /data/dataset`,
+		Flags: []cli.Flag{
+			// P2P flags
+			&cli.StringFlag{Name: "peers-dns-srv", Usage: "DNS SRV record for peer discovery"},
+			&cli.StringFlag{Name: "peers-dns-a", Usage: "DNS A record FQDN for peer discovery"},
+			&cli.StringFlag{Name: "peers-list", Usage: "static peer list, comma-separated"},
+			&cli.StringFlag{Name: "listen", Value: ":19090", Usage: "HTTP server listen address"},
+			&cli.StringFlag{Name: "discovery-interval", Value: "3s", Usage: "peer discovery polling interval"},
+			&cli.StringFlag{Name: "availability-interval", Value: "2s", Usage: "availability polling interval"},
+			&cli.IntFlag{Name: "min-peers", Usage: "minimum total peer count including self; wait until at least this many are discovered before starting. Discovery returning more proceeds immediately (lower bound, not fixed size). >=2 enables manifest sharing. 0 or 1 = solo mode."},
+			&cli.BoolFlag{Name: "disable-manifest-sharing", Usage: "force every peer to resolve metadata independently (escape hatch for debugging)"},
+			&cli.StringFlag{Name: "leader-timeout", Value: "10m", Usage: "follower max wait for the leader's manifest before erroring out"},
+			&cli.StringFlag{Name: "manifest-name", Usage: "manifest object basename under p2p_warmup/ (default: content-addressable hash). All peers must pass the same value to agree on the key."},
+			// General flags
+			&cli.IntFlag{Name: "threads", Aliases: []string{"p"}, Value: 50, Usage: "concurrent fetch workers"},
+			&cli.StringFlag{Name: "file", Aliases: []string{"f"}, Usage: "file containing a list of paths"},
+			&cli.BoolFlag{Name: "keep-alive", Usage: "keep serving blocks to peers after warmup completes (until SIGTERM/SIGINT)"},
+			&cli.StringFlag{Name: "keep-alive-timeout", Value: "0", Usage: "auto-shutdown after this duration in keep-alive mode (0=wait for SIGTERM); ignored without --keep-alive"},
+			&cli.StringFlag{Name: "cache-dir", Usage: "cache directory (default: from volume format)"},
+			&cli.Int64Flag{Name: "download-limit", Usage: "object-storage download bandwidth cap in Mbps per peer (0=unlimited); peer-to-peer transfers are not throttled"},
+		},
+	}
+}
+
+func p2pWarmup(c *cli.Context) error {
+	setup0(c, 2, 0) // at least META-URL + one PATH, unlimited max
+
+	addr := c.Args().Get(0)
+	removePassword(addr)
+
+	// Collect paths from positional args and --file flag.
+	paths := c.Args().Tail()
+	if fname := c.String("file"); fname != "" {
+		fd, err := os.Open(fname)
+		if err != nil {
+			logger.Fatalf("Failed to open file %s: %s", fname, err)
+		}
+		defer fd.Close()
+		scanner := bufio.NewScanner(fd)
+		for scanner.Scan() {
+			if p := strings.TrimSpace(scanner.Text()); p != "" {
+				paths = append(paths, p)
+			}
+		}
+		if err = scanner.Err(); err != nil {
+			logger.Fatalf("Reading file %s failed with error: %s", fname, err)
+		}
+	}
+	if len(paths) == 0 {
+		logger.Infof("no paths specified")
+		return nil
+	}
+
+	// Initialize meta client.
+	metaConf := meta.DefaultConf()
+	metaConf.ReadOnly = true
+	metaConf.NoBGJob = true
+	metaCli := meta.NewClient(addr, metaConf)
+	format, err := metaCli.Load(true)
+	if err != nil {
+		return err
+	}
+
+	// Create object storage from format.
+	blob, err := createStorage(*format)
+	if err != nil {
+		return err
+	}
+
+	// Determine cache directory.
+	cacheDir := c.String("cache-dir")
+	if cacheDir == "" {
+		cacheDir = defaultCacheDir()
+	}
+	// Append volume UUID to cache dir, same as chunk.Config.SelfCheck().
+	cacheDir = path.Join(cacheDir, format.UUID)
+
+	// Build WarmupConfig from CLI flags.
+	config := p2p.WarmupConfig{
+		ListenAddr:           c.String("listen"),
+		DiscoveryInterval:    utils.Duration(c.String("discovery-interval")),
+		AvailabilityInterval: utils.Duration(c.String("availability-interval")),
+		Threads:              c.Int("threads"),
+		KeepAlive:            c.Bool("keep-alive"),
+		KeepAliveTimeout:     utils.Duration(c.String("keep-alive-timeout")),
+		MinPeers:             c.Int("min-peers"),
+
+		DisableManifestSharing: c.Bool("disable-manifest-sharing"),
+		LeaderTimeout:          utils.Duration(c.String("leader-timeout")),
+		ManifestName:           strings.TrimSpace(c.String("manifest-name")),
+
+		PeersDNSSRV: c.String("peers-dns-srv"),
+		PeersDNSA:   c.String("peers-dns-a"),
+		PeersList:   c.String("peers-list"),
+
+		CacheDir:   cacheDir,
+		BlockSize:  format.BlockSize * 1024,
+		HashPrefix: format.HashPrefix,
+		Compress:   format.Compression,
+
+		DownloadLimit: utils.ParseMbps(c, "download-limit") * 1e6 / 8,
+	}
+
+	// Set up context with signal cancellation.
+	ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGTERM, syscall.SIGINT)
+	defer cancel()
+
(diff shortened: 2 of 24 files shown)


kyungwan-nam:
@jiefenghuang 
I've pushed 5 fix commits on top of the original
Thanks for taking the time to review

PR #7034, opened 2026-05-15
cmd/mount: improve mount process kill and diagnostics

Xuyuchao-juice:
- Fix lastActive initial value to avoid false positive at startup
- Add processExists() to verify process death before signal escalation
- Extend double-check wait from 10s to 30s for slow core dumps
- Add debug=1 goroutine aggregation before debug=2 full stacks
- Add printThreadsStack() with status/schedstat/syscall for kernel diagnosis
- Remove redundant post-kill /proc/PID/task stack reading
- Restructure watchdog to print stacks before kill for accurate diagnosis

### DIFF
--- a/cmd/mount_unix.go
+++ b/cmd/mount_unix.go
@@ -78,7 +78,9 @@ func showThreadStack(agentAddr string) {
 		return nil
 	}
 	if err := fn(1); err == nil {
-		fn(2)
+		if err := fn(2); err != nil {
+			logger.Warnf("list goroutines debug=2 from %s: %s", agentAddr, err)
+		}
 	}
 }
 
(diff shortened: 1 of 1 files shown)


cmd/mount_unix.go
```diff
@@ -202,16 +242,29 @@ func watchdog(ctx context.Context, mp string) {
 	for ctx.Err() == nil {
 		now := time.Now().Unix()
 		if atomic.LoadInt64(&lastActive)+30 < now {
+			logger.Infof("mount point %q is not active for %s, start another check thread", mp, time.Since(time.Unix(atomic.LoadInt64(&lastActive), 0)))
+			go printThreadsStack(pid)
 			showThreadStack(agentAddr)
```
jiefenghuang: go printGoroutinesStack(agentAddr)

Xuyuchao-juice:
部分日志截图
image

jiefenghuang (approved):

PR #7037, opened 2026-05-15
cmd/gateway: fix infinite loop in config handling

zhijian-pro:


zhijian-pro:
ref https://github.com/juicedata/minio/pull/83

davies (approved):

PR #7038, opened 2026-05-15
CI:add cleanup logic in testCheckAndRepair for idempotency

Xuyuchao-juice:
If the file is created but not cleaned up, it may cause the next test to fail.


pkg/meta/base_test.go
```diff
@@ -2900,6 +2900,16 @@ func setAttr(t *testing.T, m Meta, inode Ino, attr *Attr) {
 func testCheckAndRepair(t *testing.T, m Meta) {
 	var checkInode, d1Inode, d2Inode, d3Inode, d4Inode Ino
 	dirAttr := &Attr{Mode: 0644, Full: true, Typ: TypeDirectory, Nlink: 3}
+	if st := m.Lookup(Background(), RootInode, "check", &checkInode, dirAttr, false); st == 0 {
```
jiefenghuang, line 2903: 完整清理RootInode下的, 后面还有d1,d2,d3...

jiefenghuang (changes requested):

jiefenghuang (approved):

Issue #7045, opened 2026-05-18
support http conn max age time


jiefenghuang:
**Why is this needed**:

Go 1.26 (https://github.com/golang/go/issues/75772) provides more customizable interfaces, making it easily to gradually evict old HTTP connections and prevent DNS changes from being ineffective because long-lived old connections never close.

qingketsing:
I’m interested in working on this issue.

Before starting the implementation,I want to confirm the expected design:

1. Should this be exposed as a new user-facing option, for example --http-conn-max-age / config field?
2. Since this issue mentions Go 1.26, should the implementation rely on the new net/http APIs, or should it keep compatibility with the current minimum Go version used by JuiceFS?
3. Is the expected behavior to gradually retire old idle HTTP connections after the configured max age or close all the connections immediately?

If this direction sounds reasonable, I’d be happy to investigate the current HTTP client/transport usage first and then submit a small PR.

jiefenghuang:
A user has already proposed an implementation in #7025  to address the issue where old connections in the connection pool may remain active for a long time without becoming idle and therefore never get evicted. Setting a maximum lifetime for connections is a reasonable approach.

Since Go 1.26 exposes interfaces that allow customization, this implementation can be simplified based on those new interfaces.

qingketsing:
Thanks for the clarification. I’ll take a look at #7025 first to understand the previous approach and the edge cases it tried to cover. Then I’ll check how the new Go 1.26 net/http interfaces can simplify the implementation, especially around retiring old active connections without interrupting in-flight requests.

qingketsing:
Thanks, I’ve looked into #7025 and the current HTTP client/transport usage in the object package.

My understanding is that #7025 is request-count-based: it wraps `net.Conn`, detects request boundaries through Read/Write, and retires a connection after it has served a configured number of requests. For this issue, I think the expected behavior should be time-based instead. It means retiring connections after they exceed a configured max lifetime.

I also noticed that some code paths and SDK integrations rely on `object.GetHttpClient().Transport.(*http.Transport)`, for example TLS-related configuration and some object storage SDK integrations. So I’ll try to avoid replacing `httpClient.Transport` with a custom `RoundTripper` unless that is considered acceptable.

### My plan is:

1. Add an opt-in HTTP connection max-age setting, disabled by default.
2. Keep the default behavior unchanged when the setting is not configured.
3. Apply it to the global object HTTP transport path first.
[...]

PR #7052, opened 2026-05-20
object: support HTTP connection max age with Go 1.26 ClientConn

qingketsing:
## Summary

This PR adds opt-in HTTP connection max-age support for the shared object HTTP client.

When `JFS_HTTP_CONN_MAX_AGE` is set and JuiceFS is built with Go 1.26+, the shared object HTTP client uses `net/http` `ClientConn` to retire expired connections gracefully:

* stop assigning new requests to expired connections
* allow in-flight requests to finish
* close expired connections after they become idle

When the variable is unset, set to `0`, or JuiceFS is built with Go < 1.26, the default behavior remains unchanged.

Related to #7045.

## Motivation

#7045 is about old pooled HTTP connections remaining alive for a long time, which may prevent DNS changes from taking effect promptly.

Go 1.26 exposes `net/http` `ClientConn`, which makes it possible to manage connection lifetime at the HTTP connection-management layer instead of relying on `net.Conn` read/write boundary detection.

Compared with a `net.Conn` wrapper approach, this avoids closing connections from the raw Read/Write path and allows in-flight requests to finish before expired connections are closed.

### 1. Keep the base transport as `*http.Transport`

The shared object HTTP stack is split into:

* a shared base `*http.Transport`
* a shared `*http.Client`

When max age is disabled, the client uses the base transport directly.

When max age is enabled on Go 1.26+, the client uses a custom `RoundTripper` backed by `ClientConn`. The base `*http.Transport` is still kept and exposed through `GetHttpTransport()` for integrations that require the concrete transport type.

This avoids breaking existing code paths that previously relied on:

```go
object.GetHttpClient().Transport.(*http.Transport)
```

### 2. Use `ClientConn` to retire expired connections gracefully

For the shared object HTTP client path, the Go 1.26+ implementation maintains a small per-host `ClientConn` pool.

For each host:

* connections are tracked with `createdAt`, `maxAge`, and retiring state
* once a connection exceeds max age, it is marked retiring
* retiring connections do not accept new requests
* if a retiring connection still has in-flight requests, it stays alive
* once `InFlight() == 0`, it is closed

The implementation uses the common `ClientConn` interface, so the retirement logic is shared by HTTP/1 and HTTP/2 connections.

### 3. Bound connection creation under concurrency

The custom transport also bounds connection creation under concurrency:

* at most one in-progress dial per host
* waiters are woken when pool state changes
* per-host connection limits are derived from the base transport
* expired/dead connections are swept safely

### 4. Keep pre-Go-1.26 behavior unchanged

For Go versions before 1.26:

* the shared client keeps using the base transport directly
* `JFS_HTTP_CONN_MAX_AGE` is ignored with a warning

This preserves compatibility with the current minimum Go version in `go.mod`.

## Scope and coverage

This PR intentionally scopes graceful retirement to the shared object HTTP client path.

It does not force all SDK integrations onto the custom `RoundTripper`.

Some integrations still require direct access to `*http.Transport`, for example:

* TLS configuration paths in `cmd/format.go`
* SDK integrations such as OBS / Swift / TOS

For those paths, this PR keeps using the shared base transport via `GetHttpTransport()` to avoid a broader refactor.

So the boundary is:

* shared object HTTP client path: `ClientConn` max-age retirement
* direct `*http.Transport` integrations: continue to use the shared base `*http.Transport`

## User-facing behavior

This is opt-in and disabled by default.

For example:

```bash
JFS_HTTP_CONN_MAX_AGE=30m
```

The value is parsed with `time.ParseDuration`, so values such as `10m`, `30m`, and `1h` are supported.

If unset or set to `0`, behavior remains unchanged.

On Go < 1.26, the setting is ignored with a warning.

## Tests

Added Go 1.26+ tests covering:

* max age disabled behavior
* retiring expired idle connections
* keeping in-flight requests running before closing expired connections
* HTTP/1 and HTTP/2 paths
* concurrent request scheduling around the per-host connection pool

## Verification

Verified locally with:

```bash
go test -count=1 ./pkg/object -run 'TestHTTPConnMaxAgeTransport' -timeout 60s -v
go test -count=1 ./pkg/object
go test -race -count=1 ./pkg/object -run 'TestHTTPConnMaxAgeTransport' -timeout 60s -v
go test -count=1 -run '^$' ./cmd
```

Also verified the legacy path with Go 1.25:

```bash
go test -count=1 -run '^$' ./pkg/object ./cmd
```

`go test -run '^$' ./cmd` was used as a compile check because the full `./cmd` test suite requires a local Redis service.

`git diff --check` also passes.

## Notes

This PR uses an environment variable to keep the change scoped and low-risk. A CLI/config option can be added later without changing the core connection-retirement logic.

### DIFF
--- a/cmd/format.go
+++ b/cmd/format.go
@@ -24,7 +24,6 @@ import (
 	"fmt"
 	"io"
 	"math/rand"
-	"net/http"
 	_ "net/http/pprof"
 	"net/url"
 	"os"
@@ -244,7 +243,7 @@ func createStorage(format meta.Format) (object.ObjectStorage, error) {
 			if tlsSkipVerify, err = strconv.ParseBool(values.Get("tls-insecure-skip-verify")); err != nil {
 				return nil, err
 			}
-			object.GetHttpClient().Transport.(*http.Transport).TLSClientConfig.InsecureSkipVerify = tlsSkipVerify
+			object.GetHttpTransport().TLSClientConfig.InsecureSkipVerify = tlsSkipVerify
 			values.Del("tls-insecure-skip-verify")
 			u.RawQuery = values.Encode()
 			format.Bucket = u.String()
@@ -268,8 +267,8 @@ func createStorage(format meta.Format) (object.ObjectStorage, error) {
 				return nil, fmt.Errorf("error appending CA cert to pool")
 			}
 
-			object.GetHttpClient().Transport.(*http.Transport).TLSClientConfig.RootCAs = certPool
-			object.GetHttpClient().Transport.(*http.Transport).TLSClientConfig.Certificates = []tls.Certificate{clientTLSCert}
+			object.GetHttpTransport().TLSClientConfig.RootCAs = certPool
+			object.GetHttpTransport().TLSClientConfig.Certificates = []tls.Certificate{clientTLSCert}
 		}
 	}
 
--- a/pkg/object/http_conn_max_age_go126.go
+++ b/pkg/object/http_conn_max_age_go126.go
@@ -20,6 +20,7 @@ package object
 
 import (
 	"fmt"
+	"math/rand"
 	"net"
 	"net/http"
 	"net/url"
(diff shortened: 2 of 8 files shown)


CaitinChen:
@jiefenghuang PTAL

qingketsing:
Hi @jiefenghuang, gentle ping on this PR.

This PR implements the HTTP connection max-age support discussed in #7045, using Go 1.26 `net/http` `ClientConn` for graceful connection retirement while keeping the behavior disabled by default.

All checks are passing from my side. Could you please take a look when you have time? I’m happy to adjust the design or implementation if needed.

jiefenghuang:
Hi @qingketsing sorry for the delayed reply. I’ll take a look this week, but this will probably be planned for version 1.5. Recently, we’ll prioritize issues related to the 1.4 release.

qingketsing:
any process now qwq

Looking forward team's reply~ @jiefenghuang

jiefenghuang:
Hi @qingketsing, because some users already depend on the implementation in #7025. Also, we may not update to version 1.26 very soon. Going forward, the plan is to track go-stable minus one or two versions. So we should probably still adopt the implementation from #7025.

If we upgrade the Go version later, we can reconsider this PR.

PR #7053, opened 2026-05-21
meta/trash: add trash_cleanup_overdue_seconds metric to monitor trash backlog

polyrabbit:
This PR adds a `trash_cleanup_overdue_seconds` metric to expose how far trash cleanup is lagging behind the configured retention window.

When trash cleanup cannot keep up or gets stuck on old trash entries, we have no visibility unless inspecting logs or the trash dir manually. This metric helps detect delayed trash cleanup early, makes alerting on cleanup health much easier.

Since each JuiceFS instance exports this metric independently, users should aggregate it using `max()` to get a volume-level backlog.


jiefenghuang:
This metric is not really necessary; you can check .trash to get this information?

PR #7054, opened 2026-05-21
pkg/chunk: bug fixes and Go 1.25 polish

solracsf:
## Summary

Findings from an audit of `pkg/chunk`, split across 15 small, self-contained commits. **Every commit stands alone**. Each touches a single concern, has its own message explaining the why, and can be cherry-picked, reordered, reverted, or dropped without affecting the others.

Maintainers should feel free to merge a subset and/or request changes on the rest.

PR #7056, opened 2026-05-22
cache/stage: fix free stage blocks when disk full

jiefenghuang:
ref: https://github.com/juicedata/juicefs/pull/7054/changes/3b54fb96c062253023f17d503e7f816ef856a8a6

### DIFF
--- a/pkg/chunk/disk_cache.go
+++ b/pkg/chunk/disk_cache.go
@@ -23,7 +23,6 @@ import (
 	"hash/fnv"
 	"io"
 	"io/fs"
-	"math"
 	"os"
 	"path/filepath"
 	"reflect"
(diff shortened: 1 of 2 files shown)


pkg/chunk/disk_cache.go
```diff
@@ -796,6 +802,20 @@ func (cache *cacheStore) uploaded(key string, size int) {
...
 
+func (cache *cacheStore) spaceToFree(usage DiskFreeRatio) int64 {
+	if usage.br < cache.freeRatio {
+		return int64(float64(usage.spaceCap) * float64(cache.freeRatio-usage.br))
+	}
+	return 0
+}
+
+func (cache *cacheStore) inodesToFree(usage DiskFreeRatio) int64 {
+	if usage.fr < cache.freeRatio {
```
solracsf, line 813:
```suggestion
	if usage.inodeCap > 0 && usage.fr < cache.freeRatio {
```
`isFull` guards against zero-inode filesystems, but `inodesToFree` does not

jiefenghuang: Since usage.inodeCap is 0, inodeToFree is also 0.

Xuyuchao-juice (approved):

zxh326 (approved):

PR #7081, opened 2026-05-27
deps: bump `storj.io/uplink` to v1.14.2

solracsf:


jiefenghuang:
I think we don’t need to update the referenced SDK version so frequently unless there is a serious bug.

solracsf:
I understand. This version introduces important timeout handlings, but your call here 👍 
https://github.com/storj/uplink/commit/787d21613e5b426eabdd7abd5661056ecfb44909

jiefenghuang (approved):

PR #7088, opened 2026-05-28
Report host disk stats on Windows mounts

anasalzuvix:
Add option to report real host volume disk statistics when mounting on Windows. Introduces mountLocal and hostVolRoot fields and imports golang.org/x/sys/windows. Statfs now calls GetDiskFreeSpaceEx for the host volume (using a 4KB block size) when mountLocal is enabled, falling back to existing metadata-based stats on error. Serve reads the CLI flag as-local-volume, derives the host volume root from the current working directory, and logs the chosen host volume or any error.

### DIFF
--- a/pkg/winfsp/winfs.go
+++ b/pkg/winfsp/winfs.go
@@ -145,7 +145,6 @@ func (j *juice) Statfs(path string, stat *fuse.Statfs_t) int {
 				return 0
 			}
 		}
-		logger.Warnf("Failed to get host disk stats for %s, falling back to metadata stats", j.hostVolRoot)
 	}
 
 	ctx := j.newContext()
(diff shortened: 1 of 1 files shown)


HJiefeng (dismissed):
~

jiefenghuang:
statfs shows the capacity information of the entire distributed file system. Adding two parameters to switch the view to local disk capacity information does not seem necessary. Is there any practical use case for this?

anasalzuvix:
I know statfs is traditionally supposed to represent the total capacity of the remote backend. But when we're mounting virtual filesystems on desktop environments (especially Windows), we really need an option to emulate the local host's drive stats instead as cryptomator have an option to treat the drive as a local drive (this project have also introduce 
`as-local-volume` paramater recently) so with this changes we will be able to achieve a local disk style partition where it will show the total capacity left to prevent a large writes as an S3 bucket is practically infinite, but the user's local SSD isn't. If someone tries to drop a 200 GB file into the mount, but their local drive only has 50 GB free, the local write-cache will fill up and crash the transfer midway through. If we report the local host's disk space, the OS knows the actual physical write limit before things go sideways. (if i am not wrong)

also, It's honestly just confusing for end-users. When someone opens Windows Explorer and sees a 1.00 PB drive (when use `as-local-volume`) it looks like a glitch and immediately breaks the illusion of a native drive. Showing a realistic capacity bar (e.g., 120 GB free out of 500 GB) makes the mount feel familiar and trustworthy.


Screenshot 2026-06-10 at 1 54 10 PM

the upper image is a feature that can be achieve with JuiceFS too if we show real data usages volume to treat the volume as a part of local disk.

jiefenghuang:
Do you mean the backend storage is using local filesystem? You can set --capacity when creating JuiceFS volume with cmd/format.

jiefenghuang:
Close this PR for now.

PR #7142, opened 2026-06-05
meta: avoid creating massive tkv tombstones for quota keys that were never set

polyrabbit:
Currently, `rmdir` and `rename` always issue a delete for `QD<inode>`, even if the directory never had quota configured. With an MVCC backend such as TiKV/etcd, deleting always leaves a tombstone record regardless of whether the key existed. Those tombstone records make subsequent scans over the `QD` prefix very expensive. In one of our clusters, when users deleted massive directories, the TiKV cluster became overloaded and slowed down significantly. The root cause was that the periodic background `refresh` calls `loadQuotas`, which scans the entire `QD` prefix, eventually exhausting all CPU.

### DIFF
--- a/pkg/meta/tkv.go
+++ b/pkg/meta/tkv.go
@@ -3534,34 +3534,28 @@ func (m *kvMeta) doLoadQuotas(ctx Context) (map[uint64]*Quota, map[uint64]*Quota
 	userQuotas := make(map[uint64]*Quota)
 	groupQuotas := make(map[uint64]*Quota)
 
-	if format.DirStats {
-		pairs, err := m.scanValues(ctx, m.fmtKey("QD"), -1, nil)
-		if err != nil {
-			return nil, nil, nil, fmt.Errorf("failed to load dir quotas: %w", err)
-		}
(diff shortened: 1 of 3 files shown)


pkg/meta/tkv.go
```diff
@@ -3525,40 +3529,43 @@ func (m *kvMeta) doDelQuota(ctx Context, qtype uint32, key uint64) error {
...
 
 func (m *kvMeta) doLoadQuotas(ctx Context) (map[uint64]*Quota, map[uint64]*Quota, map[uint64]*Quota, error) {
-	quotaTypes := []struct {
-		prefix string
-		name   string
-	}{
-		{"QD", "dir"},
-		{"QU", "user"},
-		{"QG", "group"},
+	format := m.getFormat()
```
jiefenghuang, line 3532: Similar to the Redis change, adding a configuration filter to `quotaTypes` should be enough.

polyrabbit: OK, updated

jiefenghuang (approved):

PR #7143, opened 2026-06-05
pkg/chunk: add `--max-staging-size` to bound staged writeback data

### LINKED ISSUE #4051: Add `--upload-treshold` to better control not uploaded cached data

A command like `--upload-treshold MiB` to be used when `--writeback` is used.
Actually, only `--upload-delay` can be used to delay the uploads. In this mode, the amount of data to be upload can't be controlled. By adding `--upload-treshold xxx` one could control that:

1. when `--writeback` is set
2. and either or both `--upload-delay` and `--upload-treshold` are set, the first param being reached will trigger the upload to the object store. 

Example:

`juicefs mount --writeback --upload-delay 90 --upload-treshold 300`

Here, once 90 seconds have passed _OR_ data to be upload on the cache is `>= 300 MiB`, data would be uploaded.

**Why is this needed**:

Better control of local data.

solracsf:
In writeback mode only `--upload-delay` (time-based) and the free-space ratio could trigger uploads, so there was no way to cap the absolute amount of not-yet-uploaded data staged on local disk.

Add `--max-staging-size`: once the total staging size reaches the configured value, staged blocks are uploaded immediately regardless of `--upload-delay`, while still respecting `--upload-hours`. Total staged bytes are tracked via a new atomic counter mirroring the existing staging_block_bytes accounting and exposed through `CacheManager.usedStaging()`.

Fixes #4051


jiefenghuang:
The existing free-space-ratio and upload-delay should already be sufficient. The background task periodically reclaims space and keeps disk usage under control, so users do not need to worry about this. Therefore, I don’t think we need to add this configuration for now.

solracsf:
No problem. This has been implemented in our own internal fork.

But please, don't keep issues opened for years if you don't intended to implement them... 🙃

PR #7153, opened 2026-06-09
feat: opt readahead in some situation

201341:


201341:
@SandyXSD @jiefenghuang PTAL

PR #7171, opened 2026-06-12
meta/tkv: batch get quota key to reduce one rpc

polyrabbit:
Supplement to #7142: reduce one extra RPC caused by `tx.get(quotaKey)`.

### DIFF
--- a/pkg/meta/tkv.go
+++ b/pkg/meta/tkv.go
@@ -2220,8 +2220,7 @@ func (m *kvMeta) doRename(ctx Context, parentSrc Ino, nameSrc string, parentDst 
 				return syscall.EEXIST
 			}
 			dtyp, dino = m.parseEntry(dbuf)
-			dgets := tx.gets(m.inodeKey(dino), m.dirQuotaKey(dino))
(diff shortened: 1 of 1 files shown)


pkg/meta/tkv.go
```diff
@@ -2220,7 +2220,8 @@ func (m *kvMeta) doRename(ctx Context, parentSrc Ino, nameSrc string, parentDst
 				return syscall.EEXIST
 			}
 			dtyp, dino = m.parseEntry(dbuf)
-			a := tx.get(m.inodeKey(dino))
+			dgets := tx.gets(m.inodeKey(dino), m.dirQuotaKey(dino))
```
jiefenghuang: We can keep the original code logic here.

Quotas only need to be deleted for directories; regular files do not need this, and they are a much more common case.

jiefenghuang (approved):

pkg/meta/tkv.go
polyrabbit: OK, it will be much simpler to keep as is.

PR #7172, opened 2026-06-12
object: enhance S3 configuration with default checksum options

zhijian-pro:


solracsf:
Other AWS-SDK providers are skipped in purpose?

May I suggest a shared helper (e.g. `loadS3Config(ctx, extraOpts ...func(*config.LoadOptions) error)`) that prepends the defaults, used by all call sites?

After this PR, a user on genuine AWS S3 who wants the stronger when_supported integrity back has no way to opt in. Maybe only append each default when the corresponding env var is unset, e.g.

```
if os.Getenv("AWS_REQUEST_CHECKSUM_CALCULATION") == "" {
    loadOpts = append(loadOpts, config.WithRequestChecksumCalculation(aws.RequestChecksumCalculationWhenRequired))
}
```
(or wire it to a URI query param, matching the existing disable-checksum convention).

jiefenghuang:
By default, disable it by setting AWS SDK environment variables. Users who need checksum calculation can enable it through environment variables.
image

jiefenghuang (approved):

PR #7202, opened 2026-06-24
fuse: experimental write-path FUSE passthrough (WIP)

EngHabu:
> **Status: experimental / WIP, opt-in and off by default.** Nothing changes for normal mounts.

## What this is

Write-path **FUSE passthrough** (kernel ≥ 6.9) for JuiceFS. When a file is opened for write, JuiceFS hands the kernel a local **staging file** as a passthrough backing file (`FOPEN_PASSTHROUGH` + `OpenOut.BackingID`). The kernel then serves that file's `read()`/`write()` **directly against the staging file**, with **no per-op upcall to the JuiceFS daemon**. On release, the staging file is reconciled into normal JuiceFS slices via the existing writer path, and the kernel's attr/data cache is invalidated so readers see the committed file.

This removes the dominant cost of small random writes over FUSE — the kernel↔daemon round trip per operation.

## How it works

- **Open/Create (write):** allocate `‹staging-dir›/‹ino›-‹fh›.tmp`, register it with `Server.RegisterBackingFd`, return `FOPEN_PASSTHROUGH` + `BackingID`. Reads/writes now bypass the daemon.
- **Release:** read the staging file back, replay it through `vfs.Write` → `vfs.Flush` (normal slice upload + metadata commit), `UnregisterBackingFd`, delete the staging file, then `InodeNotify` to refresh the kernel cache.
- **Durability model:** writes are durable to object storage at **release/commit**, not per-op — the same commit-style semantics passthrough implies (writes land in the local staging file first). For overwrite-heavy workloads this also *reduces* object-store write amplification, since reconcile uploads each block's final state once.

Opt-in (prototype): `JUICEFS_PASSTHROUGH=1` and `JUICEFS_PASSTHROUGH_DIR=‹dir›`. The staging dir **must be on a non-stacked filesystem** (tmpfs/ext4/xfs) — an overlayfs/fuse backing returns `ELOOP` from the kernel's `max_stack_depth` check. On a real node the JuiceFS cache dir (node disk) satisfies this.

## Dependency

Requires FUSE passthrough support in the go-fuse binding, which `release-2.5` doesn't have yet — see **juicedata/go-fuse#53**. This PR currently wires it via a `replace` to that branch; it would point at the merged go-fuse commit once #53 lands.

## Benchmarks (4 KiB random write, local `file://` backend, kernel 6.12)

| | vanilla | **passthrough** | factor |
|---|---|---|---|
| buffered | ~780 IOPS | **~1.88M IOPS** | ~2400× |
| durable (`fdatasync`) | ~270 IOPS | **~152k IOPS** | ~560× |

The per-op FUSE daemon tax is gone — writes hit the local staging file at native speed. (Benchmark staged on tmpfs, so the durable figure is RAM-fsync-bound; on node NVMe expect lower but still ~10–50k, i.e. well above the ~270 vanilla floor.) Reads and metadata ops are unchanged — this is purely a write-path optimization.

## Correctness

- Data + metadata are always committed correctly (verified: post-flush metadata size matches bytes written, and remount reads match the original).
- **Read-your-writes verified** for 1 / 17 / 64 / 200 / 333 MB (multi-chunk) after the `InodeNotify` cache-invalidation fix.
- The cache-invalidation fix adds **no throughput cost** (two notify calls at close; measured IOPS unchanged).

## What's changed

- `pkg/fuse/passthrough.go` (new) — staging-file lifecycle + reconcile.
- `pkg/fuse/fuse.go` — Open/Create inject passthrough; Release reconciles; Serve negotiates the capability.
- `pkg/vfs/vfs.go` — mirror the two new `MountOptions` fields into `FuseOptions`.
- `go.mod` — `replace` go-fuse with the passthrough-enabled fork.

## Remaining WIP (before this is merge-ready)

- `getattr` returns a stale size **while** a passthrough file is open and being written (daemon never sees the writes); report the backing file's size.
- Sub-millisecond race if a file is `stat`'d microseconds after `close()` (async FUSE notify); closeable by disabling attr-cache for passthrough files.
- Concurrent handles to the same inode; promote the env-var opt-in to a real `--passthrough` mount flag; add a regression test per `AGENTS.md`.


EngHabu:
Depends on **juicedata/go-fuse#53**, which adds the FUSE passthrough primitives (`RegisterBackingFd` / `FOPEN_PASSTHROUGH` / INIT negotiation) to `release-2.5`. This PR's `go.mod` currently `replace`s go-fuse with that branch; once #53 merges it will point at the merged commit. #53 should land first.

EngHabu:
@jiefenghuang please take a look.

jiefenghuang:
> @jiefenghuang please take a look.

Thank you very much for your contribution. We’ll be handling the 1.4 release this week, I’ll review this a little later.

PR #7208, opened 2026-06-28
meta: fix openfiles cleanup goroutine leak on Shutdown

### LINKED ISSUE #7207: Shutdown never stops the openfiles cleanup goroutine
**What happened**:

`newOpenFiles` starts a background `cleanup` goroutine, but `Shutdown` never stops it. `Shutdown` only closes the backend connection. So every `meta.NewClient` leaks one goroutine.

With the in-memory `memkv` backend, repeat `NewClient` + `Shutdown` and read `runtime.NumGoroutine()`. It grows by one per cycle (100 cycles → +100), all stuck at `(*openfiles).cleanup`.

**Environment**:
- JuiceFS version: current `main`

Looka149:
Fixes #7207

`newOpenFiles` starts a `cleanup` goroutine that `Shutdown` never stops, leaking one goroutine per `meta.NewClient`. The fix adds a stop channel to `openfiles` and closes it in `Shutdown`, matching the other `baseMeta` goroutines that are stopped via `CloseSession`.

Verified with `memkv`: 100× `NewClient`+`Shutdown` leaks +100 goroutines before, +0 after. Regression test included; `go test -race` clean.

### DIFF
--- a/pkg/meta/openfile.go
+++ b/pkg/meta/openfile.go
@@ -43,9 +43,11 @@ func (o *openFile) release() {
 
 type openfiles struct {
 	sync.Mutex
-	expire time.Duration
-	limit  uint64
-	files  map[Ino]*openFile
+	expire    time.Duration
+	limit     uint64
+	files     map[Ino]*openFile
+	stop      chan struct{}
+	closeOnce sync.Once
 }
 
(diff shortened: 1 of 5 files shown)


jiefenghuang (approved):

Looka149:
@jiefenghuang Thank you for the review and merge.

PR #7229, opened 2026-07-06
sync: don't skip destination listing with --force-update and --delete-dst

### LINKED ISSUE #7216: sync: --force-update disables --delete-dst behavior
## Describe the bug

When using `juicefs sync` with both `--force-update` and `--delete-dst`, the `--delete-dst` flag has no effect — extraneous objects in the destination are not deleted.

## Root Cause

In `startSingleProducer` and `startProducer` (`pkg/sync/sync.go`), when `--force-update` is set, the destination listing is skipped entirely by replacing `dstkeys` with an immediately-closed empty channel:

```go
// sync.go:1373-1376
if config.ForceUpdate {
    t := make(chan object.Object)
    close(t)
    dstkeys = t
}
```

This is a reasonable optimization since `--force-update` always overwrites existing files and doesn't need to compare src vs dst. However, it means the `produce` function never sees any destination objects, so it can never detect "extraneous objects from destination" that `--delete-dst` is supposed to identify and delete.

In the `produce` function (`sync.go:1508-1521`), the `--delete-dst` cleanup code iterates over `dstkeys` to find and delete extra dst objects:

```go
if config.DeleteDst {
    // ...
    for dstobj = range dstkeys {  // empty channel, no iteration
        // ...
        if handleExtraObject(tasks, dstobj, config, checkpointMgr, prefix) {
            return nil
        }
    }
}
```

Since `dstkeys` is an empty closed channel, this loop produces no iterations, and no destination objects are ever identified as extraneous.

The same pattern exists in both the single-producer path (`startSingleProducer`, line 1372) and the multi-threaded listing path (`startProducer`, line 2032).

## Expected behavior

When both `--force-update` and `--delete-dst` are specified, the sync should:
1. Copy all source files to destination (overwriting existing ones) — this works
2. Delete any files in destination that don't exist in source — **this does not work**

## Suggested fix

When `--delete-dst` is true, the destination should still be listed even if `--force-update` is true. The destination listing is still needed to identify extraneous objects to delete, even though it's not needed for the copy decision.

For example, change both occurrences from:
```go
if config.ForceUpdate {
```
to:
```go
if config.ForceUpdate && !config.DeleteDst {
```

zhijian-pro:


pkg/sync/sync.go
```diff
@@ -1937,7 +1937,7 @@ func startProducer(tasks chan<- object.Object, src, dst object.ObjectStorage, pr
 	defer func() {
 		<-config.concurrentList
 	}()
-	if config.Limit == 1 && len(config.rules) == 0 {
+	if config.Limit == 1 && len(config.rules) == 0 && !config.DeleteDst {
```
jiefenghuang: remove

jiefenghuang (approved):

PR #7232, opened 2026-07-07
meta: fix setgid strip ignoring CheckPermission and supplementary groups

### LINKED ISSUE #6940: Does not allow sticky bit on group when group is not primary group
**What happened**:
Setting sticky bit on directory chgrp'd by a secondary group does not allow the sticky bit to be set.

Sticky bit gets set.å

- create two groups
- create a user with one primary and one secondary group
- create a directory
- chown the directory to the user and its secondary group
- chmod g+s directory --> silently fails.  sticky bit does not get set, and chmod exits with success

**Environment**:
  - juicefs version 1.3.1+2025-12-02.e0032b2
  - AWS, redis/valkey + s3
  - Red Hat Enterprise Linux 9.7 (Plow)
- Kernel (e.g. `uname -a`):
  - Linux ip-1-2-3-4.ec2.internal 5.14.0-611.45.1.el9_7.x86_64 #1 SMP PREEMPT_DYNAMIC Sat Mar 21 06:08:21 EDT 2026 x86_64 x86_64 x86_64 GNU/Linux
  - AWS, us-east-1
  - Amazon MemoryDB cluster, valkey 7.3
  - AWS instance to AWS MemoryDB to AWS s3 buckets (4 shards), same VPC
- Others:

Xuyuchao-juice:
close https://github.com/juicedata/juicefs/issues/6940


pkg/meta/base.go
```diff
@@ -1575,7 +1575,7 @@ func (m *baseMeta) inheritMode(ctx Context, _type uint8, parentGid uint32, paren
 		if _type == TypeDirectory {
 			childMode |= 02000
 		} else if childMode&02010 == 02010 && ctx.Uid() != 0 {
```
jiefenghuang: else if ctx.CheckPermission() && childMode&02010 == 02010 && ctx.Uid() != 0

jiefenghuang (approved):

PR #7247, opened 2026-07-10
CI: Retry tier storage class assertion after forced set

zhijian-pro:


zhijian-pro:
## Rationale

`juicefs info` may trigger asynchronous compaction when a file contains multiple slices. In this test, compaction can start while the file is still on tier 2.

The following race may then occur:

1. Background compaction starts with tier 2.
2. `juicefs tier set --tier 3` rewrites the objects visible at that moment and updates the inode metadata to tier 3.
3. The command exits successfully.
4. The earlier compaction finishes and publishes a new tier 2 object.
5. The inode reports tier 3, while the actual object remains in the tier 2 storage class.

The existing `tier_set_no_err` retry does not handle this case because it only retries when the command exits with a non-zero status or emits `<ERROR>`/`<FATAL>`. Here, the command itself succeeds and the mismatch appears afterward.

This change adds a postcondition-aware retry for the affected roundtrip test:

1. Run the initial `tier set` normally.
2. Check the actual object storage class through `juicefs info`.
3. If it does not match the expected class, wait briefly and rerun `tier set` with `--force`.
4. Require two consecutive successful checks, followed by the existing direct S3 `HeadObject` verification.

`--force` is required for the retry because the inode metadata may already contain the target tier ID. Without it, a repeated same-tier command would return early without rescanning and rewriting the current objects.

The test intentionally does not compact the file before changing tiers. This keeps the test aligned with the production recovery procedure: retry `tier set` after observing a compaction race. Only one postcondition retry is allowed; a second mismatch still fails the test and exposes a persistent regression.

zhijian-pro:
ref #https://github.com/juicedata/juicefs/actions/runs/28974755340/job/85979057242

YunhuiChen (approved):

YunhuiChen:
合并出的对象去获取当前文件的sc 然后去设置，可能这个问题概率可以小很多

PR #7257, opened 2026-07-13
cmd/fsck: fix unexpected object storage head errors

### LINKED ISSUE #7246: S3 client silently signs us-east-1 for custom-endpoint buckets when AWS_REGION is unset, then fsck reports every object missing on custom-region stores
**What happened**:
Running `juicefs fsck` against a healthy volume on a custom-region S3 store (MinIO with `MINIO_SITE_REGION=eu-lab-1`), from an environment without `AWS_REGION` set, reported every block of every file as missing:
```
<ERROR>: can't find block 0/0/1_0_6 for file /f: operation error S3: HeadObject, https response error StatusCode: 400 ...
<FATAL>: 1 objects are lost (6 bytes), 1 broken files:
```
The store's response shows why: every request was signed for the wrong region:
```xml
<Error><Code>AuthorizationHeaderMalformed</Code><Message>The authorization header is
malformed; the region is wrong; expecting 'eu-lab-1'.</Message>...</Error>
```
Nothing is wrong with the data. The same fsck with `AWS_REGION=eu-lab-1` exported reports `Found 1 blocks (6 bytes)` and exits 0. In a pipeline that trusts fsck as a verification gate, this is a guaranteed false data-loss report; an operator could plausibly "restore from backup" over a healthy volume.

The client signs with the region the volume was created against (or probes the store), and fsck distinguishes "object absent" from "request rejected" instead of reporting `can't find block` for a 400.

```
docker run -d --name minio -p 127.0.0.1:9000:9000 \
    -e MINIO_ROOT_USER=testkey -e MINIO_ROOT_PASSWORD=testsecret12 \
    -e MINIO_SITE_REGION=eu-lab-1 quay.io/minio/minio server /data
# create the bucket (mc, or any client signed with the store's region):
docker run --rm --network host --entrypoint sh quay.io/minio/mc -c \
    "mc alias set m http://127.0.0.1:9000 testkey testsecret12 && mc mb m/jfs"

AWS_REGION=eu-lab-1 juicefs format --storage s3 \
    --bucket http://127.0.0.1:9000/jfs --access-key testkey --secret-key testsecret12 \
    sqlite3:///tmp/repro-meta.db vol
AWS_REGION=eu-lab-1 juicefs mount -d sqlite3:///tmp/repro-meta.db /tmp/mnt
echo hello > /tmp/mnt/f && umount /tmp/mnt

env -u AWS_REGION -u AWS_DEFAULT_REGION juicefs fsck sqlite3:///tmp/repro-meta.db --path / --recursive
# -> <ERROR>: can't find block 0/0/1_0_6 for file /f: ... StatusCode: 400
#    <FATAL>: 1 objects are lost (6 bytes), 1 broken files
#    (store responds AuthorizationHeaderMalformed: "the region is wrong; expecting 'eu-lab-1'")

AWS_REGION=eu-lab-1 juicefs fsck sqlite3:///tmp/repro-meta.db --path / --recursive
# -> Found 1 blocks (6 bytes); exit 0
```

_Root cause (pkg/object/s3.go, v1.4.0)_
For `[ENDPOINT]/[BUCKET]`-style URLs on non-AWS hosts, `newS3` takes the "compatible s3" branch, which never calls `autoS3Region` and never consults the store:
```go
// uri.Path != "" && !strings.Contains(uri.Host, ".amazonaws.com")
ep = uri.Host            // region stays ""
...
if region == "" { region = os.Getenv("AWS_REGION") }
if region == "" { region = os.Getenv("AWS_DEFAULT_REGION") }
if region == "" { region = awsDefaultRegion }   // "us-east-1"
```
The volume format carries the bucket URL but no region, so every process re-derives the region from its own environment.

_Why it looks intermittent when it is strictly deterministic_
Mount/format typically run in environments where `AWS_REGION` happens to be exported (containers, shell profiles), so the failure surfaces only in contexts with a clean environment: cron jobs, systemd units, orchestration tools spawning fsck. Same binary, same volume, different env, opposite verdicts. Note that shell-profile exports do NOT reach systemd/cron/non-interactive-ssh contexts, so "just set it in bashrc" does not close the hole.

_Suggested fixes_
1. Persist the region in the volume format at `juicefs format` time (it is known there) and use it for every subsequent client, instead of re-deriving from the process environment.
2. Failing that: for compatible-s3 endpoints, probe the store (GetBucketLocation) before falling back to us-east-1, and log the chosen region at INFO on client creation.
3. Defense in depth for fsck: distinguish 404 ("object absent") from other 4xx ("request rejected"): today both surface as `can't find block`, which converts a client-side configuration defect into a false data-loss report.

_Workaround_
Export `AWS_REGION=<store region>` in the environment of every juicefs invocation. For services that means `Environment=` lines in the systemd units, since shell profiles don't apply there.

**Environment**:
`juicefs version 1.4.0+2026-07-06.62bedf3` (the region-resolution code is unchanged on main)
self-hosted x86_64; the repro is hardware-independent
Debian GNU/Linux 13 (trixie)
- Kernel (e.g. `uname -a`):
6.12.95+deb13-amd64 x86_64
MinIO (`quay.io/minio/minio`), self-maintained, custom region `eu-lab-1`, path-style bucket URL
any (the bug is in the object client); the minimal repro uses SQLite
same host (Docker), sub-millisecond; irrelevant to the bug
- Others:
any store that validates the SigV4 credential scope region (MinIO with a site region does) rejects the mis-signed requests with 400; stores that ignore scope region mask the bug.

zhijian-pro:


cmd/fsck.go
```diff
@@ -194,14 +196,20 @@ func fsck(ctx *cli.Context) error {
...
-								brokens[inode] = ps[0]
+								filePath = ps[0]
 							} else {
-								brokens[inode] = fmt.Sprintf("inode:%d", inode)
+								filePath = fmt.Sprintf("inode:%d", inode)
 							}
 						}
-						logger.Errorf("can't find block %s for file %s: %s", objKey, brokens[inode], err)
+						if !errors.Is(err, os.ErrNotExist) {
+							progress.Done()
```
jiefenghuang: 不直接退出

jiefenghuang (approved):

PR #7324, opened 2026-07-29
meta: add sync option for badger metadata engine

### LINKED ISSUE #7249: Support SyncWrites on BadgerDB

On a badger-backed volume, `fsync()` returns success once the metadata transaction commits, but the commit isn't on stable storage: a power loss/kernel panic loses acknowledged, fsync'd operations (creates, renames, slice commits). Every other JuiceFS engine is durable by default: SQLite (go-sqlite3 defaults to synchronous=FULL), MySQL/Postgres (redo/WAL fsync at commit), TiKV/etcd/FDB (consensus + fsync). Redis is the only other soft spot by default, and the docs explicitly tell users to configure AOF.

Badger's gap is silent and undocumented.

Happy to work on it if accepted. 👍

The only thing needed is: should the default be kept to false, or switch to true by default, and offer the possibility to disable it with `badger:///path/to/badger?sync=false` (or `sync=true` if false remains the default).

**Why is this needed**:

Metadata-engine durability parity.

dutor:
## Summary
- add `sync` query option for Badger metadata URL (`badger://...?...`)
- map `sync=true|false` to Badger `SyncWrites` (default: false)
- treat invalid values (including `sync=1|0`) as false and log a warning
- add Badger tests for default/true/false/invalid values
- update Badger metadata docs in English and Chinese

## Testing
- `go test ./pkg/meta -run 'TestBadger(SyncOption|Client|KV|ScanKeysOnlyNilValues|SimpleTxnReadOnly|DeleteTxnTooBig)$'`

## Notes
- `make test.meta.core` requires Redis in this environment and fails at `TestRedisClient` when `127.0.0.1:6379` is unavailable.

Closes #7249

### DIFF
--- a/pkg/meta/tkv_badger.go
+++ b/pkg/meta/tkv_badger.go
@@ -265,7 +265,7 @@ func newBadgerClient(addr string) (tkvClient, error) {
 		}
 		syncValue := strings.ToLower(query.Get("sync"))
 		switch syncValue {
-		case "", "false":
+		case "false":
 			opt.SyncWrites = false
 		case "true":
 			opt.SyncWrites = true
--- a/pkg/meta/tkv_badger.go
+++ b/pkg/meta/tkv_badger.go
@@ -22,8 +22,10 @@ package meta
 import (
 	"bytes"
 	"context"
+	"net/url"
 	"os"
 	"strconv"
+	"strings"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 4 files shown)


pkg/meta/tkv_badger.go
```diff
@@ -254,7 +256,24 @@ func (c *badgerClient) close() error {
...
+	dataPath, queryStr, _ := strings.Cut(addr, "?")
+	opt := badger.DefaultOptions(dataPath)
+	if queryStr != "" {
+		query, err := url.ParseQuery(queryStr)
+		if err != nil {
+			return nil, err
+		}
+		syncValue := strings.ToLower(query.Get("sync"))
+		switch syncValue {
+		case "", "false":
```
jiefenghuang: Remove ""
```go
case "false":
```

jiefenghuang (approved):

PR #7340, opened 2026-08-01
meta/redis: fix stale F_RDLCK not released by session cleanup

prOOrc:
# Fix: register BSD read-locks (`F_RDLCK`) in the `locked$` index so stale-session cleanup can release them

## Summary

`redisMeta.Flock` writes the lock field into `lockf$<inode>` for **both** read and write locks, but adds the reverse-index entry `SAdd(locked$<sid>, lockf$<inode>)` **only** in the `F_WRLCK` branch — the `F_RDLCK` branch is missing it.

Since `doCleanStaleSession` frees a dead session's locks by iterating **only** `locked$<sid>` (there is no sid-prefix scan of `lockf*` anywhere in the cleanup path), a read-lock whose holder dies without a clean `F_UNLCK` is **never released**: the field stays in `lockf$<inode>` forever while the session itself is removed from `sessionInfos`/`allSessions` as if nothing were wrong. Any later `F_WRLCK` on the same inode then returns `EAGAIN` indefinitely.

This PR adds the missing `SAdd` to the read branch, making it symmetric with the write branch and consistent with the index contract documented at the top of `redis.go` (`locked$sid -> { lockf$inode or lockp$inode }`, with no write-only qualifier). The cleanup code is correct and untouched; the index was simply never populated for reads.

> **TL;DR** — a one-line omission in the read-lock path makes dead sessions' shared locks uncleanable, permanently blocking exclusive locks on the same inode.

## How it was discovered

The bug was triggered in production by **MinIO gateway** on JuiceFS. MinIO's `NewNSLock(MinioMetaBucket, MinioMetaLockFile)` takes a BSD read-lock (`F_RDLCK`) on the `.sys.minio` metadata file through the JuiceFS FUSE layer. On every unclean shutdown of the gateway (e.g., `SIGKILL` in Kubernetes), the read-lock field remained in `lockf$<inode>` forever while the session itself was removed from `allSessions` by stale-session cleanup. Subsequent MinIO gateway restarts then failed to acquire the exclusive lock on `.sys.minio`, spinning on `EAGAIN` indefinitely. The only recovery was a manual `HDEL lockf$<inode> <sid>_<owner>`.

Every orphaned field observed in production was of type `R`, never `W` — exactly what this asymmetry predicts.

## Why it stayed hidden

`meta.Flock(..., F_RDLCK, ...)` is rarely exercised upstream (FUSE mostly sees `LOCK_EX`; record locks go through `Setlk`/`lockp$`), so a "dead shared-lock gets released" case was likely never covered. It becomes fatal for any real `F_RDLCK` consumer — we hit it through the JuiceFS-backed MinIO gateway that maps shared namespace locks to `F_RDLCK` on the `.sys.minio` metadata file, where the omission reproduces on essentially every unclean shutdown.

## Symptoms (after an unclean shutdown, e.g. `SIGKILL` in k8s)

- `<sid>_<owner>: R` remains in `lockf$<inode>` (standard `ownerKey` format).
- `locked$<sid>` is **empty**; `sessionInfos`/`allSessions` no longer contain the sid (session "cleaned").
- A new process needing the exclusive lock spins forever: `get write lock timed out ino:<inode>` / `trying to acquire lock`, because `F_WRLCK` sees the orphaned `R` via `HGetAll` and refuses (`len(owners) > 0 → EAGAIN`), and nothing will ever remove it.
- Only recovery is a manual `HDEL lockf$<inode> <sid>_<owner>`.

## Root cause & fix

```diff
 if ltype == F_RDLCK {
     for _, v := range owners {
         if v == "W" {
             return syscall.EAGAIN
         }
     }
     _, err = tx.TxPipelined(ctx, func(pipe redis.Pipeliner) error {
         pipe.HSet(ctx, ikey, lkey, "R")
+        pipe.SAdd(ctx, r.lockedKey(r.sid), ikey)
         return nil
     })
     return err
 }
```

The `SAdd` sits in the same `TxPipelined` / under the same `WATCH(ikey)` as the `HSet`, so the field write and the index registration stay atomic — exactly as in the write branch.

We confirmed the cleanup side is innocent: had `doCleanStaleSession(<sid>)` ever run while `locked$<sid>` contained the inode, then with `fail == false` (required for the session to have been removed) it would have `HDel`'d the matched field. The field's survival proves the inode was never indexed for that read-lock — i.e. the defect is on the acquisition side.

## Safety

- **Idempotent** — repeated `F_RDLCK` on the same `(sid, inode)` adds no duplicate members; no unbounded growth of `locked$`.
- **Atomic** with the field write (same pipeline/`WATCH`); a concurrent mutation aborts and retries, as today.
- **Cleanup now frees R** — `doCleanStaleSession` sees the inode, the sid-prefix filter matches the field, `HDel` releases it; `juicefs status` also starts reporting live read-locks correctly as a side benefit.
- **Readers of `locked$` are type-agnostic** — the known consumers either delete fields by sid prefix regardless of type (`doCleanStaleSession`, correct for `R`) or only display them (`getSession`); a grep of `lockedKey` / `F_*LCK` in `pkg/meta` found no consumer assuming write-only entries (a maintainer's confirmation for any non-OSS paths would be appreciated).
- **One benign side effect:** in a shared scenario (A and B both hold `R`), A's normal `RUnlock` sees `len(lkeys) > 1` and skips `SRem`, leaving a transient stale `locked$A → ikey` pointer. It is harmless — it self-clears when A dies (`doCleanStaleSession` runs the trailing `SRem` even with an empty field list), and stale pointers of this kind already occur on the write path today. `F_UNLCK` is left unchanged to keep the fix minimal.

## Not retroactive (operational note)

The fix only registers read-locks taken **after** the patched binary runs; fields orphaned by previous versions are in no `locked$` set and won't be reclaimed by cleanup even after upgrade. Affected clusters need a one-time reconciliation that scans `lockf*`/`lockp*` directly and `HDEL`s fields whose sid is absent from `sessionInfos`/`allSessions` (safe — live holders are never touched). This is intentionally out of scope here; happy to contribute it as a follow-up.

## Testing

Two regression tests, both fail without the fix and pass with it:

### Test 1 — R-lock registration in `testLocks` (`pkg/meta/base_test.go`)

Inserted into the existing flock sequence in `testLocks`: take `F_RDLCK` on a clean inode → verify `SMembers(locked$<sid>)` returns exactly 1 entry → unlock → verify `locked$` is empty again. This catches the root cause (missing `SAdd`) and also serves as an indicator for `db`/`kv` backends if they share the same asymmetry.

```go
// Regression: a BSD read-lock must be registered in the per-session index
// (locked$<sid>) exactly like a write-lock. Otherwise stale-session cleanup
// and GetSession(detail) cannot see it, so a read-lock left behind by a
// crashed holder is never released and blocks every future write-lock on
// the same inode with EAGAIN forever.
if st := m.Flock(ctx, inode, o1, syscall.F_RDLCK, false); st != 0 {
    t.Fatalf("flock rlock: %s", st)
}
if r, ok := m.(*redisMeta); ok {
    ms, err := r.rdb.SMembers(context.Background(), r.lockedKey(r.sid)).Result()
    if err != nil {
        t.Fatalf("SMembers %s: %s", r.lockedKey(r.sid), err)
    }
    if len(ms) != 1 {
        t.Fatalf("read-lock not registered in session index: got %d entries, want 1", len(ms))
    }
}
if st := m.Flock(ctx, inode, o1, syscall.F_UNLCK, false); st != 0 {
    t.Fatalf("flock unlock: %s", st)
}
```

### Test 2 — `TestRedisStaleReadLockCleanup` (`pkg/meta/base_test.go`)

End-to-end reproduction of the operational scenario: session A takes `F_RDLCK` and "crashes" without `F_UNLCK`; `doCleanStaleSession` runs for its sid; assert (a) orphan field is gone from `lockf$<inode>` and (b) a fresh `F_WRLCK` on the same inode succeeds. Without the fix, (a) fails with `orphan read-lock survived stale-session cleanup: map[1_F000000000000001:R]`.

```go
func TestRedisStaleReadLockCleanup(t *testing.T) {
    // ... setup redis meta, create file ...
    if st := m.Flock(ctx, inode, owner, syscall.F_RDLCK, false); st != 0 {
        t.Fatalf("flock rlock: %s", st)
    }
    // Simulate crash — cleanup without F_UNLCK
    sid := r.sid
    if err := r.doCleanStaleSession(sid); err != nil {
        t.Fatalf("doCleanStaleSession: %s", err)
    }
    // (a) orphan field must be gone
    if fields, err := r.rdb.HGetAll(ctx, r.flockKey(inode)).Result(); err != nil {
        t.Fatalf("HGetAll %s: %s", r.flockKey(inode), err)
    } else if len(fields) != 0 {
        t.Fatalf("orphan read-lock survived stale-session cleanup: %v", fields)
    }
    // (b) fresh write-lock must succeed
    if st := m.Flock(ctx, inode, owner+1, syscall.F_WRLCK, false); st != 0 {
        t.Fatalf("write-lock after cleanup should succeed, got %s", st)
    }
}
```

## Out of scope

Built-in reconciler for pre-existing orphans; tightening `F_UNLCK` to drop the index entry by "no fields left for this sid"; gateway-side self-healing on lock-acquisition timeout. Listed for context only.

---

We believe this is a long-dormant omission in a rarely exercised branch, not a deliberate design choice. Happy to adjust the test or fold in any follow-up — just let us know.

### DIFF
--- a/pkg/meta/base_test.go
+++ b/pkg/meta/base_test.go
@@ -1370,6 +1370,28 @@ func testLocks(t *testing.T, m Meta) {
 	if st := m.Flock(ctx, inode, o1, syscall.F_UNLCK, false); st != 0 {
 		t.Fatalf("flock unlock: %s", st)
 	}
+
+	// Regression: a BSD read-lock must be registered in the per-session index
+	// (locked$<sid>) exactly like a write-lock. Otherwise stale-session cleanup
+	// and GetSession(detail) cannot see it, so a read-lock left behind by a
+	// crashed holder is never released and blocks every future write-lock on
+	// the same inode with EAGAIN forever.
+	if st := m.Flock(ctx, inode, o1, syscall.F_RDLCK, false); st != 0 {
+		t.Fatalf("flock rlock: %s", st)
+	}
+	if r, ok := m.(*redisMeta); ok {
+		ms, err := r.rdb.SMembers(context.Background(), r.lockedKey(r.sid)).Result()
+		if err != nil {
+			t.Fatalf("SMembers %s: %s", r.lockedKey(r.sid), err)
+		}
+		if len(ms) != 1 {
+			t.Fatalf("read-lock not registered in session index: got %d entries, want 1", len(ms))
+		}
+	}
+	if st := m.Flock(ctx, inode, o1, syscall.F_UNLCK, false); st != 0 {
+		t.Fatalf("flock unlock: %s", st)
+	}
+
 	if r, ok := m.(*redisMeta); ok {
 		ms, err := r.rdb.SMembers(context.Background(), r.lockedKey(r.sid)).Result()
 		if err != nil {
@@ -6167,4 +6189,69 @@ func newFuseDefaultCtx(uid, primaryGid uint32) *fuseDefaultCtx {
 		uid:     uid,
 		gid:     primaryGid,
 	}
+}
+
+// TestRedisStaleReadLockCleanup is a regression test for the gateway outage:
+// a read-lock whose holder crashed (no F_UNLCK) used to survive stale-session
+// cleanup, because F_RDLCK never added the inode to locked$<sid>, which is the
+// only thing doCleanStaleSession scans. The orphan R then made every F_WRLCK
+// on the same inode return EAGAIN forever. This test reproduces the crash +
+// cleanup + a fresh write-lock.
+func TestRedisStaleReadLockCleanup(t *testing.T) {
+	if os.Getenv("SKIP_NON_CORE") == "true" {
+		t.Skipf("skip non-core test")
+	}
+	m, err := newRedisMeta("redis", "127.0.0.1:6379/10", testConfig())
+	if err != nil {
+		t.Fatalf("create meta: %s", err)
+	}
+	if err := m.Reset(); err != nil {
+		t.Fatalf("reset meta: %s", err)
+	}
+	if err := m.Init(testFormat(), false); err != nil {
+		t.Fatalf("init meta: %s", err)
+	}
+	if err := m.NewSession(true); err != nil {
+		t.Fatalf("new session: %s", err)
+	}
+	defer m.CloseSession() // err ignored: cleanup below already removed the session
+
+	r, ok := m.(*redisMeta)
+	if !ok {
+		t.Fatal("not a redisMeta")
+	}
+
+	ctx := Background()
+	var inode Ino
+	var attr = &Attr{}
+	if st := m.Create(ctx, 1, "f_stale_rlock", 0644, 0, 0, &inode, attr); st != 0 {
+		t.Fatalf("create f: %s", st)
+	}
+	defer m.Unlink(ctx, 1, "f_stale_rlock")
+
+	owner := uint64(0xF000000000000001)
+	if st := m.Flock(ctx, inode, owner, syscall.F_RDLCK, false); st != 0 {
+		t.Fatalf("flock rlock: %s", st)
+	}
+
+	// Simulate a crash: the holder dies without F_UNLCK. A surviving cleaner
+	// (any live client) eventually runs cleanup for this sid.
+	sid := r.sid
+	if err := r.doCleanStaleSession(sid); err != nil {
+		t.Fatalf("doCleanStaleSession: %s", err)
+	}
+
+	// (a) the orphan read-lock field must be gone from lockf$<inode>
+	if fields, err := r.rdb.HGetAll(ctx, r.flockKey(inode)).Result(); err != nil {
+		t.Fatalf("HGetAll %s: %s", r.flockKey(inode), err)
+	} else if len(fields) != 0 {
+		t.Fatalf("orphan read-lock survived stale-session cleanup: %v", fields)
+	}
+
+	// (b) the operational symptom: a fresh write-lock on the same inode must
+	// succeed now (previously it looped on EAGAIN against the dead R).
+	if st := m.Flock(ctx, inode, owner+1, syscall.F_WRLCK, false); st != 0 {
+		t.Fatalf("write-lock after cleanup should succeed, got %s", st)
+	}
+	_ = m.Flock(ctx, inode, owner+1, syscall.F_UNLCK, false)
 }
--- a/pkg/meta/base_test.go
+++ b/pkg/meta/base_test.go
@@ -6255,3 +6255,167 @@ func TestRedisStaleReadLockCleanup(t *testing.T) {
 	}
 	_ = m.Flock(ctx, inode, owner+1, syscall.F_UNLCK, false)
 }
+
+// TestRedisLockIndexRelease covers the release side of the locked$<sid> index:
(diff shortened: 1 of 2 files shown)


pkg/meta/redis_lock.go
jiefenghuang: After this change, the release logic also needs to be updated; otherwise, it will cause a leak. When none of the owners in a session holds a lock on a given inode, the corresponding lockedKey should be deleted.

The setlk logic also needs to be updated, and unit tests should be added to cover these cases.

prOOrc: Good catch — the old `len(lkeys) == 1 && lkeys[0] == lkey` condition only worked
while an inode could never be shared between sessions (true for W-only flock
registration, false once R-locks are indexed, and already false for shared
POSIX read-locks in `Setlk`). Replaced it with a per-session check
(`lastLockOfSession`): the `locked$<sid>` entry is dropped exactly when no
other owner of the same session still holds a field in `lockf/lockp$<inode>`,
and fields of other sessions neither keep our entry alive nor are removed by
us. Applied to both `Flock` and `Setlk` release paths (both read `lkeys`
inside the same WATCHed txn, so the check is atomic). Added
`TestRedisLockIndexRelease` covering: release while another session still
holds the inode (index entry must be dropped, foreign lock untouched), and
multi-owner release within one session (entry must survive until the last
owner), for both flock and setlk.

pkg/meta/base_test.go
```diff
@@ -6168,3 +6190,232 @@ func newFuseDefaultCtx(uid, primaryGid uint32) *fuseDefaultCtx {
...
+// TestRedisStaleReadLockCleanup is a regression test for the gateway outage:
+// a read-lock whose holder crashed (no F_UNLCK) used to survive stale-session
+// cleanup, because F_RDLCK never added the inode to locked$<sid>, which is the
+// only thing doCleanStaleSession scans. The orphan R then made every F_WRLCK
+// on the same inode return EAGAIN forever. This test reproduces the crash +
+// cleanup + a fresh write-lock.
+func TestRedisStaleReadLockCleanup(t *testing.T) {
+	if os.Getenv("SKIP_NON_CORE") == "true" {
+		t.Skipf("skip non-core test")
+	}
```
zxh326: It looks like the current CI does not run these tests. Could we remove the SKIP_NON_CORE?

pkg/meta/redis_lock.go
```diff
@@ -29,6 +29,20 @@ import (
 	"github.com/redis/go-redis/v9"
 )
 
+// lastLockOfSession reports whether no field in the lock hash other than lkey
```
jiefenghuang: The function name and the PR history are already clear enough, so we can remove this comment.

pkg/meta/base_test.go
prOOrc: Removed

pkg/meta/redis_lock.go
prOOrc: Removed

zxh326 (approved):
LGTM

jiefenghuang (approved):

Issue #7344, opened 2026-08-03
cache: avoid selecting store twice in exist


Light-City:
**What happened**:

`cacheManager.exist` calls `getStore(key)` twice:

```go
store := m.getStore(key)
if store == nil {
    return "", false
}
loc := store.dir
existed, err := m.getStore(key).exist(key)
```
A cache store may become unavailable and be removed between these two calls.
In that case, the second getStore(key) may:
- Return nil, causing a nil pointer panic when calling .exist(key).
- Return a different store, making loc inconsistent with the store used for the existence check.
This is a check-then-use race caused by selecting the store twice.

**What you expected to happen**:
cacheManager.exist should use one consistent store instance for the primary existence check and should not panic when cache directories become unavailable.

**How to reproduce it (as minimally and precisely as possible)**:
no

**Anything else we need to know?**
no

**Environment**:
- JuiceFS version (use `juicefs --version`) or Hadoop Java SDK version:
- Cloud provider or hardware configuration running JuiceFS:
- OS (e.g `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Object storage (cloud provider and region, or self maintained):
- Metadata engine info (version, cloud provider managed or self maintained):
- Network connectivity (JuiceFS to metadata engine, JuiceFS to object storage):
- Others:

PR #7345, opened 2026-08-03
cache: avoid selecting store twice in exist

Light-City:
This PR fixes a potential nil pointer panic and inconsistent result in `cacheManager.exist`.

The function previously selected the primary cache store twice:

```go
store := m.getStore(key)
if store == nil {
    return "", false
}
loc := store.dir
existed, err := m.getStore(key).exist(key)
```
This ensures that:
- The nil check applies to the store that is actually used.
- The returned location matches the queried store.
- No additional lock or synchronization is added.
The existing legacy store fallback behavior remains unchanged.

### DIFF
--- a/pkg/chunk/disk_cache.go
+++ b/pkg/chunk/disk_cache.go
@@ -1310,7 +1310,7 @@ func (m *cacheManager) exist(key string) (string, bool) {
 		return "", false
 	}
 	loc := store.dir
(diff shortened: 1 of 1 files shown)


Light-City:
related issue: https://github.com/juicedata/juicefs/issues/7344

jiefenghuang (approved):

jiefenghuang:
fix #7344

Issue #7343, opened 2026-08-03
Chunk/Cache: Data races in cached store upload and disk cache full checking


Light-City:
**What happened**:

Running `TestStoreDefault` with Go's race detector reports two data races in the chunk cache implementation.

**What you expected to happen**:
TestStoreDefault should pass with the race detector enabled and without data race warnings.

**How to reproduce it (as minimally and precisely as possible)**:
```bash
MOCKEY_CHECK_GCFLAGS=false \
go test -race -run '^TestStoreDefault$' -count=1 ./pkg/chunk
```

outputs:
```
~/work/juicefs (main) » MOCKEY_CHECK_GCFLAGS=false \                       
go test -race -run '^TestStoreDefault$' -count=1 ./pkg/chunk
# github.com/juicedata/juicefs/pkg/chunk.test
ld: warning: '/private/var/folders/sw/_0l2l2hd0cnd74cy415t2r980000gn/T/go-link-1795917873/000078.o' has malformed LC_DYSYMTAB, expected 98 undefined symbols to start at index 1626, found 95 undefined symbols starting at index 1626

Mockey check failed, please add -gcflags="all=-N -l".
(Set env MOCKEY_CHECK_GCFLAGS=false to disable gcflags check) 

2026/08/03 23:17:23.512285 juicefs[79782] <WARNING>: unknown cache eviction policy: "",fallback to 2-random [newCacheStore@disk_cache.go:119]
2026/08/03 23:17:23.512562 juicefs[79782] <WARNING>: not enough space (4%) or inodes (97%) for caching in /var/folders/sw/_0l2l2hd0cnd74cy415t2r980000gn/T/diskCache-79782/: free ratio should be >= 10% [newCacheStore@disk_cache.go:152]
2026/08/03 23:17:23.512605 juicefs[79782] <INFO>: Disk cache (/var/folders/sw/_0l2l2hd0cnd74cy415t2r980000gn/T/diskCache-79782/): used ratio - [space 95%, inode 2.3%] [newCacheStore@disk_cache.go:154]
2026/08/03 23:17:23.512628 juicefs[79782] <INFO>: Adjusted max items based on freeratio: from 0 to 221032952 items [setLimitByFreeRatio@disk_cache.go:191]
==================
WARNING: DATA RACE
Write at 0x00c0006d4908 by goroutine 42:
  github.com/juicedata/juicefs/pkg/chunk.(*cacheStore).scanCached()
[...]

PR #7346, opened 2026-08-03
cache: fix data races in cached store.

Light-City:
## What does this PR do?

This PR fixes two data races in the chunk cache implementation.

### Race 1: concurrent access to `Page.Data`

In `cachedStore.upload()`, the block was published to the asynchronous disk cache before compression completed:

```go
store.bcache.cache(key, block, false, false)
n, err := store.compressor.Compress(buf.Data, block.Data)
buf.Data = buf.Data[:n]
```

When using the no-op compressor, `buf` and `block` reference the same `Page`. The disk cache goroutine could therefore read `Page.Data` while the upload goroutine was modifying its contents or slice header.

`Page.Acquire()` only extends the lifetime of the page and does not synchronize concurrent access to its data.

#### Fix

Complete compression and update `buf.Data` before publishing the block to the disk cache:

```go
n, err := store.compressor.Compress(buf.Data, block.Data)
if err != nil {
    block.Release()
    return fmt.Errorf("Compress block key %s: %s", key, err)
}
buf.Data = buf.Data[:n]

if sync && (blen < store.conf.BlockSize || store.conf.CacheLargeWrite) {
    store.bcache.cache(key, block, false, false)
}
block.Release()
```

This ensures the page becomes immutable before another goroutine can access it.

### Race 2: concurrent access to `rawFull`

`cacheStore.checkFreeSpace()` periodically updated `rawFull`, while `cacheStore.cache()` read it concurrently without synchronization:

```text
cacheStore.checkFreeSpace -> rawFull write
cacheStore.cache          -> rawFull read
```

#### Fix

Change `rawFull` to `atomic.Bool` and use `Store()` and `Load()` for concurrent access.

A local variable is used during each free-space checking iteration to avoid unnecessary atomic reads:

```go
rawFull := cache.isFull(usage, false)
cache.rawFull.Store(rawFull)
```

The cache path reads the published state with:

```go
if cache.rawFull.Load() && cache.keys.name() == EvictionNone {
    // Drop the cache request.
}
```

This fixes the race without adding mutex contention to the frequently executed cache path.

## Test

```bash
MOCKEY_CHECK_GCFLAGS=false \
go test -race -run '^TestStoreDefault$' -count=1 ./pkg/chunk
```

Result:

```text
ok github.com/juicedata/juicefs/pkg/chunk
```

### DIFF
--- a/pkg/chunk/cached_store.go
+++ b/pkg/chunk/cached_store.go
@@ -365,16 +365,17 @@ func (store *cachedStore) upload(ctx context.Context, key string, block *Page, s
 		buf.Acquire()
 	}
 	defer buf.Release()
+	n, err := store.compressor.Compress(buf.Data, block.Data)
+	if err != nil {
+		block.Release()
+		return fmt.Errorf("Compress block key %s: %s", key, err)
+	}
+	buf.Data = buf.Data[:n]
 	if sync && (blen < store.conf.BlockSize || store.conf.CacheLargeWrite) {
 		// block will be freed after written into disk
 		store.bcache.cache(key, block, false, false)
 	}
-	n, err := store.compressor.Compress(buf.Data, block.Data)
 	block.Release()
-	if err != nil {
-		return fmt.Errorf("Compress block key %s: %s", key, err)
-	}
-	buf.Data = buf.Data[:n]
 
 	try, max := 0, 3
 	if sync {
--- a/pkg/chunk/disk_cache.go
+++ b/pkg/chunk/disk_cache.go
@@ -92,7 +92,7 @@ type cacheStore struct {
 	keys      KeyIndex
 	scanned   bool
 	stageFull bool
-	rawFull   bool
+	rawFull   atomic.Bool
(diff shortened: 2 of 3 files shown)


Light-City:
related issue: https://github.com/juicedata/juicefs/issues/7343

jiefenghuang:
Race1: "while the upload goroutine was modifying its contents or slice header."  There is nowhere else that modifies page.Data.
Race2: It is acceptable not to use atomic operations here, since rawFull (whether the disk is full) is itself updated through periodic checks and does not need to reflect changes in real time.

Light-City:
> while the upload goroutine was modifying its contents or slice header." There is nowhere else that modifies page.Data.

Yeah, it's fine if we don't fix it. It only shows up when the race detector is on. I think we can fix the first race since it's just moving some code around, which can avoid the noise.

jiefenghuang:
This is not a bug, and no actual data race occurs, so I’m closing this PR for now.

PR #7351, opened 2026-08-04
meta/fdb: use snapshot read for simpleTxn

201341:


jiefenghuang:
Hi @201341 this LGTM, and do you have any benchmark results demonstrating the performance benefits of this change?

201341:
> Hi @201341 this LGTM, and do you have any benchmark results demonstrating the performance benefits of this change?

I think this change doesn't improve metadata performance a lot. It improve the fdb resolvers' pressure and benefits the scanLogRange

jiefenghuang:
If this reduces the load on the FDB resolvers, it should increase the maximum workload the system can handle. Put another way, under the same workload, how much does this change reduce FDB’s resource usage? Has this been validated with any benchmark tests? 
I’m mainly asking so that we have some benchmark comparison results on record for future reference.

jiefenghuang (approved):

PR #7356, opened 2026-08-05
meta: speedup trash cleanup by parallel directory deletion

polyrabbit:
`BatchUnlink` has greatly improved trash cleanup. However, it only applies to file entries. Directories are still deleted sequentially, so trash buckets with many directories are still slow.

This PR optimizes directory cleanup by:

- deleting trash directories concurrently with 10 workers, as many other cleanup routines do;
- removing transaction slot lock for trash parents, because deleting sibling entries in the trash does not update parent attributes and can safely run in parallel;
- skipping unnecessary quota updates for trash parents;

For one of our internal volume used to stores many GitHub repositories, the trash contains a very large number of files and directories:

- With parallel directory deletion: about 600 entries/s and no txn conflicts.
image

- With the previous serial directory deletion: about 130 entries/s.
image

As a longer-term optimization, it's better to support a dedicated `BatchRmdir` api.

### DIFF
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -3184,7 +3184,7 @@ func (m *baseMeta) CleanupTrashBefore(ctx Context, edge time.Time, increProgress
 		}()
 	}
 
-	concurrent := make(chan int, 10)
+	concurrent := make(chan int, backgroundDeleteThreads)
 	for len(entries) > 0 {
 		if ctx.Canceled() {
 			return errno(ctx.Err())
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -1838,7 +1838,9 @@ func (m *baseMeta) Rmdir(ctx Context, parent Ino, name string, skipCheckTrash ..
 			m.parentMu.Unlock()
 		}
 		m.updateDirStat(ctx, parent, 0, -align4K(0), -1)
-		m.updateDirQuota(ctx, parent, -align4K(0), -1)
+		if !parent.IsTrash() {
+			m.updateDirQuota(ctx, parent, -align4K(0), -1)
+		}
 	}
 	return st
 }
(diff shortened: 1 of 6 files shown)


pkg/meta/base.go
```diff
@@ -3182,7 +3184,7 @@ func (m *baseMeta) CleanupTrashBefore(ctx Context, edge time.Time, increProgress
 		}()
 	}
 
-	concurrent := make(chan int, 1) // no effect for flatterned trash dirs
+	concurrent := make(chan int, 10)
```
jiefenghuang: It can be set to RmrDefaultThreads.

pkg/meta/tkv.go
```diff
@@ -1117,6 +1117,9 @@ func (m *kvMeta) txn(ctx Context, f func(tx *kvTxn) error, inodes ...Ino) error
 			return syscall.EINTR
 		}
 		err := m.client.txn(ctx, f, i)
+		if err != nil && ctx.Canceled() {
```
jiefenghuang: This is already handled by func errno(err error) syscall.Errno, so this can be removed.

pkg/meta/base.go
polyrabbit: Setting background deletion concurrency to 50 is too high. Existing code already uses a concurrency of 10 for background deletions — this should stay consistent. You can search for `rmConcurrent := make(chan int, 10)`.

polyrabbit: I’ve also extracted this 10 into a shared constant so background-deletion concurrency is unified.

pkg/meta/tkv.go
polyrabbit: When context is cancelled, TiKV wraps the error in a long string. `errno(err)` doesn't recognize it and treats it as EIO, which then dumps thousands of lines via `debug.Stack()`. You can easily reproduce to see those noisy stack traces.

pkg/meta/utils.go
```diff
@@ -139,7 +139,7 @@ func errno(err error) syscall.Errno {
 	if err == nil {
 		return 0
 	}
-	if err == context.Canceled {
+	if errors.Is(err, context.Canceled) {
```
jiefenghuang: if err == context.Canceled || strings.Contains(err.Error(), "context canceled") { ?

polyrabbit: OK, updated

jiefenghuang (approved):

PR #7368, opened 2026-08-07
feat: add Lance dataset warmup support

hiltonhe:
## Summary

Add Lance dataset warmup support to the `warmup` command. This allows JuiceFS to parse Lance manifest files and automatically discover/warm up all related data files, including column-level metadata warmup for V2 format.

## Changes

- **Dataset-level warmup**: Parse Lance manifest → auto-discover all data/deletion/overlay/transaction files
- **Column-level metadata warmup** (V2): Parse V2 footers and CMO tables → warm up column metadata byte ranges
- New CLI options: `--lance`, `--lance-version`, `--lance-columns`, `--lance-column-data`, `--lance-manifest-only`, `--lance-include-indices`
- 13 tests (9 unit + 4 E2E)
- Protobuf definitions from Lance project (Apache 2.0 licensed)

## Usage

```bash
# Dataset-level warmup
juicefs warmup --lance /mnt/jfs/data/dataset.lance

# Column-level metadata warmup
juicefs warmup --lance --lance-columns "col_a,col_b" /mnt/jfs/data/dataset.lance

### CODE REVIEW

**Comment by CaitinChen** [comment]
@jiefenghuang PTAL

**Comment by Suave** [comment]
Thanks for the great work on adding Lance dataset warmup support. Overall, I think the direction is valuable.

From an ecosystem perspective, format-aware warmup is a meaningful feature. Supporting formats like Lance, Parquet, Iceberg, etc. can make JuiceFS more attractive for users building AI/data workloads, because the system can understand higher-level dataset semantics instead of only operating on individual files.

Also, this PR has a good isolation boundary. The existing JuiceFS workflow is not affected unless users explicitly enable the Lance-related options, so there is no impact on the normal warmup path or the main filesystem workflow. 

That said, I have some concerns about the long-term design:

*1. CLI scalability*
Currently, supporting Lance introduces 6 format-specific flags. This is still manageable for one format, but if we add more dataset formats in the future (for example Parquet dataset, Iceberg, Delta Lake, etc.), the warmup command may become difficult to use and maintain:

```
juicefs warmup --lance-xxx ...
juicefs warmup --iceberg-xxx ...
juicefs warmup --delta-xxx ...
```

Maybe we should consider introducing a more generic dataset-aware abstraction, such as:

```
juicefs warmup --format lance ...
```

or a configuration file/plugin mechanism, so format-specific options do not continuously expand the top-level CLI.

*2. Dataset intelligence vs filesystem responsibility*

The current implementation starts to make JuiceFS understand dataset formats and decide which files need to be warmed up.

This is useful, but long term it may make the client binary more coupled with external data ecosystems. Supporting more formats means adding more parsers, metadata definitions, and compatibility logic into the JuiceFS client.
Another possible architecture is separating dataset intelligence from JuiceFS warmup:

```
Dataset resolver
    |
    |  (Lance/Iceberg/Parquet specific logic)
    |
    v
List of files (or future byte ranges)
    |
    v
juicefs warmup
```

In this model, JuiceFS remains focused on cache population, while external tools handle dataset discovery and format evolution.

Overall, I think supporting Lance is a good direction and this PR is a useful first step. My suggestion is to keep the current implementation isolated, but consider a more extensible abstraction before adding more dataset formats in the future.

hiltonhe:
> Thanks for the great work on adding Lance dataset warmup support. Overall, I think the direction is valuable.
> ...

@Suave  Thanks for the thorough review — really appreciate the architectural
perspective.

**On CLI scalability (point 1):** Agreed. The right fix is a generic
`--format` flag backed by a `DatasetResolver` interface in JuiceFS.

**On filesystem responsibility (point 2):** The external resolver pattern
(a standalone Lance resolver → `juicefs warmup --from-file -`) is
architecturally cleaner — JuiceFS stays completely format-agnostic. One
practical advantage: when Lance adds a V3 format, only the standalone tool
updates, JuiceFS doesn't need a release.

**On lancedb-go:** I checked it but it's a LanceDB database SDK (CGO + native
libs), it doesn't expose the low-level format APIs we need (manifest parsing,
footer reading, CMO table).

**Tradeoff summary:**

| | Internal Interface | External Resolver |
|---|---|---|
| UX | Single command | Two commands via pipe |
| Column-level warmup | Straightforward | Needs byte-range support in `--from-file` |
| Format evolution | JuiceFS releases needed | Independent tool updates |
| New formats | Implement interface in JuiceFS | Write a standalone tool |

Both approaches have merit. I'd lean toward the external resolver pattern
long-term for the decoupling benefit, but I'm happy to go with whatever
direction the maintainers prefer. The current Lance logic is already
self-contained in `lance_warmup.go`, so either refactor path is clean.

Would love to hear your thoughts on which direction you'd like to see.

jiefenghuang:
I think it would be better to implement this using an External Resolver approach:

1. The External Resolver should be a standalone tool responsible for parsing datasets in specific formats and generating an input list that can be consumed by the warmup tool. For now, it can live under the tools/ directory. The generated warmup format can be extended to support ranges, while remaining backward-compatible with the existing warmup format.
2. The warmup tool should focus only on filesystem-semantic operations, with additional support for warming up specified ranges of a file. This feature can be merged first as a separate PR.

hiltonhe:
> I think it would be better to implement this using an External Resolver approach:
> ...

@jiefenghuang  Thanks for the clear direction! That makes sense.

split this into two PRs:

1. **Warmup range support** (first): Add `--from-file` to the warmup command,
   supporting both file paths and byte ranges (`path [start,end)`). Pure
   JuiceFS infrastructure, no format dependency.

2. **Lance resolver** (second): Move the Lance parsing logic to
   `tools/lance-resolver/` as a standalone tool that outputs the format
   PR 1 consumes. Usage:
   ```bash
   lance-resolver --columns "id" /mnt/jfs/data/ds.lance | juicefs warmup --from-file -

Does that sound right?

jiefenghuang:
> > I think it would be better to implement this using an External Resolver approach:
> > ...
> 
> @jiefenghuang Thanks for the clear direction! That makes sense.
> 
> split this into two PRs:
> 
> 1. **Warmup range support** (first): Add `--from-file` to the warmup command,
>    supporting both file paths and byte ranges (`path [start,end)`). Pure
>    JuiceFS infrastructure, no format dependency.
> 2. **Lance resolver** (second): Move the Lance parsing logic to
>    `tools/lance-resolver/` as a standalone tool that outputs the format
>    PR 1 consumes. Usage:
>    ```shell
>    lance-resolver --columns "id" /mnt/jfs/data/ds.lance | juicefs warmup --from-file -
>    ```
> 
> Does that sound right?

Yes, and we already have the `file` flag, so we just need to define a compatible format.
For example, each line in the specified file could look something like:
file A-B;C-D;...

hiltonhe:
@jiefenghuang  Thanks for the guidance on splitting the work. I've now
submitted it as two PRs:

1. #7398 — warmup: support byte range warmup via --file flag
   (the generic range infrastructure, independent of Lance)

2. #7399 — tools/lance-resolver: standalone Lance dataset resolver
   (external tool, outputs warmup-compatible format)

PR 2 depends on PR 1. Looking forward to your review!

PR #7378, opened 2026-08-10
meta: fix races on the format

solracsf:
`baseMeta.fmt` is replaced whenever the format is reloaded, while other goroutines read it concurrently. `Load()` took the `baseMeta` mutex around the write, but `doInit()` did not take it at all and none of the readers did either, so the locking never actually excluded anything. The race detector reports it on every engine, e.g. `doInit()` writing the pointer against a background compaction goroutine reading `fmt.ChangeLog` through `deleteSlice()`.

Store it as an `atomic.Pointer` and route every access through `getFormat`/`setFormat`. An atomic load rather than the mutex because `genLog` reads the format inside every mutating transaction, and `baseMeta`'s mutex also covers session handling and quotas. Changing the field type makes the compiler point at each of the remaining accesses.

`cleanupChangelog` read `ChangeLogMaxAge` and `ChangeLogMaxLines` as two separate loads; take one snapshot so the pair stays consistent across a reload.

### Split out

The delete-slice channel fix this PR used to carry (`startDeleteSliceTasks` guarding `dslices` with the `baseMeta` mutex while `stopDeleteSliceTasks` and `deleteSlice` guard it with `dSliceMu`) moved to its own PR, as asked in review. It does not depend on this one, and the two can land in either order. Until both are in, a `-race` run still reports the other pair on each branch.

### Not covered here

The atomic pointer protects the pointer, not the pointee. Two call sites still mutate a `Format` after it has been published, so they remain racy against readers of those fields. Both predate this change and sit on separate paths, so they are left for a follow-up:

- `handleQuotaSet` (`pkg/meta/quota.go`) takes the published format from `getFormat()` and writes `DirStats` / `UserGroupQuota` into it before handing it to `doInit`.
- The reload loop invokes the `OnReload` callbacks with the same pointer `Load` has just stored, and the callback registered in `cmd/mount.go` writes `Bucket`, `Storage` and `Tiers` into it on every reload.

Both fix the same way, by patching a copy and then publishing it, which builds on the `setFormat` introduced here.

### DIFF
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -273,7 +273,9 @@ type baseMeta struct {
 	sync.Mutex
 	addr string
 	conf *Config
-	fmt  *Format
+	// fmt is reloaded in the background while other goroutines read it, so it is
+	// only ever accessed through getFormat/setFormat
+	fmt atomic.Pointer[Format]
 
 	root         Ino
 	txlocks      [nlocks]sync.Mutex // Pessimistic locks to reduce conflict
@@ -728,9 +730,7 @@ func (m *baseMeta) Load(checkVersion bool) (*Format, error) {
 	if format.Tiers == nil {
 		format.Tiers = object.NewTiers(format.StorageClass)
 	}
-	m.Lock()
-	m.fmt = format
-	m.Unlock()
+	m.setFormat(format)
 	return format, nil
 }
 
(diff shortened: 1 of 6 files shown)


pkg/meta/base.go
```diff
@@ -834,8 +834,10 @@ const (
 )
 
 func (m *baseMeta) startDeleteSliceTasks() {
-	m.Lock()
-	defer m.Unlock()
+	// dslices is guarded by dSliceMu, the same lock stopDeleteSliceTasks and
```
jiefenghuang: Fix this in a separate PR.

jiefenghuang:
There isn’t really a race condition at the moment, but it does make sense to unify the way fmt is accessed.

jiefenghuang (approved):

Issue #7374, opened 2026-08-10
warmup: skip fully-invisible blocks


jiefenghuang:
optimizations for `juicefs warmup` (and the related `check` / `evict`): **Skip fully-invisible blocks.**

PR #7385, opened 2026-08-11
cache: skip invisible blocks during warmup

Light-City:
## What does this PR do?

  Skip fully invisible blocks when running cache warmup operations, including:

  - `juicefs warmup`
  - `juicefs warmup --check`
  - `juicefs warmup --evict`

  Metadata returns the visible part of each slice through `Slice.Off` and `Slice.Len`. Previously, the cache operations only
  received the slice ID and full size, causing all blocks of a partially visible slice to be processed.

  This PR passes the visible slice range to the chunk store and only processes cache blocks overlapping `[offset, offset +
  length)`.

  A partially visible block is still processed in full because cache blocks are the minimum cache unit.

related issue: https://github.com/juicedata/juicefs/issues/7374

### DIFF
--- a/pkg/chunk/cached_store.go
+++ b/pkg/chunk/cached_store.go
@@ -82,16 +82,17 @@ func (s *rSlice) index(off int) int {
 	return off / s.store.conf.BlockSize
 }
 
-func (s *rSlice) keys() []string {
-	if s.length <= 0 {
-		return nil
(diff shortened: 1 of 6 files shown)


jiefenghuang:
Thanks for your contribution. However, this conflicts with #7398, so the implementation has been adjusted there instead. Closing this PR.

PR #7389, opened 2026-08-13
meta/redis: batch load quotas to reduce round trips

### LINKED ISSUE #7388: perf: redis meta loads quotas with one round trip per entry

In the Redis metadata engine, `doLoadQuotas` reads quota usage and limits with two serialized HGET round trips per quota entry:

- `usedInodesKey` is scanned via HSCAN (10k pairs per call), then for **each** entry the current code issues `HGET usedSpaceKey <key>` followed by `HGET quotaKey <key>` — 2N round trips for N quotas.

This path runs at mount time (`loadQuotas`), on every `juicefs quota list`, and during user/group quota check/repair. With a large number of dir/user/group quotas (tens of thousands of entries), mount takes minutes and `quota list` becomes effectively unusable.

The TKV engine already has batching for quota key fetches (#7171); Redis has no equivalent.

**Why is this needed**:

- Mount latency grows linearly with the number of quota entries (2 RTTs per entry).
- `juicefs quota list` on volumes with many quotas blocks on the same N+1 pattern.
- Nothing changes in quota semantics; this is purely reducing round trips.

**Proposed solution**:

Pipeline the HGETs in batches (`Pipelined`, 1000 entries ≈ 2000 commands per round trip) while preserving the exact per-entry semantics of the current loop:

- missing `usedSpace` / `quota` keys (redis.Nil) → `UsedSpace=0`, `MaxSpace=-1`, `MaxInodes=-1`, same as today
- invalid quota values (not 16 bytes) and unparseable keys/`usedInodes` are skipped, same as today
- additionally guards a latent out-of-range panic when HSCAN ever returns an odd number of fields (current loop indexes `keys[i+1]`)

Implementation sketch:

- `loadQuotaBatch(ctx, config, name, keys, quotas)` — parse HSCAN pairs into entries
- `loadQuotaEntries(ctx, config, name, batch, quotas)` — one `Pipelined` round trip per batch of 1000 entries, then per-entry handling identical to the current loop

Tests cover: batch boundary (>1000 entries), missing keys, invalid quota value / garbage fields, user/group quota types, and the odd-length HSCAN error path. A PR with the implementation will follow.

leeezep:
Closes #7388

## Problem

`redisMeta.doLoadQuotas` issues two serialized HGET round trips per quota entry (`HGET usedSpaceKey` + `HGET quotaKey`), driven by the HSCAN over `usedInodesKey`. With N dir/user/group quotas this is 2N round trips, executed at mount time (`loadQuotas`), on every `juicefs quota list`, and during user/group quota check/repair. Large volumes (tens of thousands of quota entries) mount slowly and `quota list` blocks on the same N+1 pattern. The TKV engine already batches quota key fetches (#7171).

## Change

- `loadQuotaBatch`: parses HSCAN pairs into entries (skipping unparseable keys/`usedInodes` as before).
- `loadQuotaEntries`: fetches `usedSpace`/`quota` for 1000 entries in one `Pipelined` round trip, then applies per-entry handling identical to the previous loop:
  - missing keys (`redis.Nil`) → `UsedSpace=0`, `MaxSpace=-1`, `MaxInodes=-1`
  - invalid quota values (not 16 bytes) are skipped with a log, as before
- Guards a latent out-of-range panic: the previous loop indexed `keys[i+1]` without checking for an odd-length HSCAN result; the new code returns an error instead.

## Tests

`pkg/meta/redis_quota_test.go` (new):

- batch boundary: 1001 entries across two pipelines, values verified
- missing `usedSpace`/`quota` keys (redis.Nil path)
- invalid quota value and garbage fields skipped
- user/group quota types loaded through the same path
- odd-length HSCAN result → error, empty result

Verified against a real Redis (`databases 32`): `go test ./pkg/meta -run TestRedisLoad -count=1 -v` passes.

### DIFF
--- a/pkg/meta/redis.go
+++ b/pkg/meta/redis.go
@@ -4478,6 +4478,95 @@ func (m *redisMeta) doDelQuota(ctx Context, qtype uint32, key uint64) error {
 	return err
 }
 
+const quotaLoadBatchSize = 1000
+
+type quotaLoadEntry struct {
+	key        string
+	id         uint64
+	usedInodes int64
+	usedSpace  *redis.StringCmd
+	quota      *redis.StringCmd
+}
+
+func (m *redisMeta) loadQuotaBatch(ctx Context, config *quotaKeys, name string, keys []string, quotas map[uint64]*Quota) error {
+	if len(keys)%2 != 0 {
+		return fmt.Errorf("invalid HSCAN result for %s quotas: odd number of fields", name)
+	}
+
+	entries := make([]quotaLoadEntry, 0, len(keys)/2)
+	for i := 0; i < len(keys); i += 2 {
+		key := keys[i]
+		id, err := strconv.ParseUint(key, 10, 64)
+		if err != nil {
+			logger.Errorf("invalid key in %s: %s", name, key)
+			continue
+		}
+		usedInodes, err := strconv.ParseInt(keys[i+1], 10, 64)
+		if err != nil {
+			logger.Errorf("invalid usedInodes for %s %s: %s", name, key, keys[i+1])
+			continue
+		}
+		entries = append(entries, quotaLoadEntry{
+			key:        key,
+			id:         id,
+			usedInodes: usedInodes,
+		})
+	}
+
+	for start := 0; start < len(entries); start += quotaLoadBatchSize {
+		end := start + quotaLoadBatchSize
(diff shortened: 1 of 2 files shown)


leeezep:
Closing; will reopen from a clean branch if needed.

PR #7398, opened 2026-08-14
warmup: support byte range warmup via --file flag

### LINKED ISSUE #7374: warmup: skip fully-invisible blocks

hiltonhe:
### Motivation

Currently `juicefs warmup -f` only supports file paths, one per line.
For columnar formats like Lance, users often need to warm up specific
byte ranges within a file (e.g. only certain column pages), not the
entire file.

### File format

Each line of the `-f` file is either a path, or a path followed by a
bracketed list of byte ranges:

```
/jfs/f1
/jfs/f2 [0-4096]
/jfs/f3 [0-4096;1048576-1052672]
```

Ranges are `start-end` pairs separated by `;`, where `start` is inclusive and
`end` is exclusive, so `[0-4096]` covers the first 4096 bytes. Because the
brackets delimit the list, paths containing spaces need no quoting. A line
whose bracketed list is malformed is reported and skipped, rather than being
silently treated as a plain path.

### Changes

**`cmd/warmup.go`**

- Parse `path [start-end;...]` lines, and keep that format when passing
  targets to the mounted client. An `inode:N` target carries its ranges the
  same way.

**`pkg/vfs/fill.go`**

- `SplitTarget` / `JoinTarget` define the target format in a single place and
  are reused by `cmd`, so the two ends cannot disagree on it.
- `parseRanges` validates strictly: any invalid part rejects the whole line.
  Ranges are then sorted and merged, so the parts derived from them never
  overlap.
- `sliceIterator` skips chunks that no range touches, and `next` returns, for
  each slice, the parts of its object that must be processed.

**`pkg/chunk`**

- `FillCache` / `EvictCache` / `CheckCache` take the parts of the object to
  operate on. Blocks shared by several parts are collected once, so a block is
  never loaded, evicted or counted twice.

### Two problems this also fixes

**Cache operations were slice-wide.** The requested ranges only decided
whether a slice was worth visiting; the store then loaded, evicted or checked
*every* block of the object. Asking to evict a hundred bytes could discard a
whole 64 MiB slice, and `--check` reported whole slices as well.

**Overwritten data was warmed up.** The size of a slice object is not the
amount of data still readable through it: a slice that has been partially
overwritten keeps its original `Size` while `Off`/`Len` shrink to the visible
part. Warming up a 64 MiB object with only 1 MiB left downloaded 63 MiB that
can never be read. Parts are now always clamped to `[Off, Off+Len)`.

The second fix applies to plain warmup as well, so warming up data that has
been overwritten now transfers less than before. Progress is also reported
from the bytes actually processed instead of the full slice size.
fix https://github.com/juicedata/juicefs/issues/7374 take over https://github.com/juicedata/juicefs/pull/7385

### Granularity

A block is the smallest unit of the cache, so a range always covers at least
one whole block (`--block-size`, 4 MiB by default). This matters most for
`--evict`, which discards every block the ranges touch, and for `--check`,
which reports whole blocks. Documented in `command_reference.mdx` (en and
zh_cn), along with the file format.

### Compatibility

- `-f` files without ranges keep working unchanged.
- No change to the CLI flags, the control protocol framing, or the on-disk
  format.
- `ChunkStore.FillCache/EvictCache/CheckCache` changed signature; all callers
  in this repository are updated. Passing no parts keeps the previous
  "whole object" behaviour.
- Mixed versions: a line *with* ranges sent to an older mounted client is
  resolved as a literal path, so it fails to resolve and is skipped with a
  warning on the mount side. Nothing is corrupted, but the warmup does not
  take effect until the mount is upgraded. Lines without ranges are
  unaffected.

### Tests

- `pkg/vfs`: `TestSplitTarget`, `TestParseRanges`, `TestParseRangesMerge`,
  `TestOverlapsRange`, `TestFillWithRanges`, `TestSliceIteratorRanges`,
  `TestSliceIteratorParts`, `TestSliceIteratorSkipsOverwrittenData`
- `pkg/chunk`: `TestBlockIndexes`, `TestFillCache`
- `cmd`: `TestWarmupWithRangeFile`

`TestSliceIteratorSkipsOverwrittenData` and the "slices returned by Read are
not modified" assertion in `TestSliceIteratorRanges` were both checked to fail
against the previous behaviour.


cmd/warmup.go
jiefenghuang: 这里带range, 路径有误

cmd/warmup.go
```diff
@@ -221,12 +221,24 @@ func warmup(ctx *cli.Context) error {
...
-			if p := strings.TrimSpace(scanner.Text()); p != "" {
-				if abs, e := filepath.Abs(p); e == nil {
-					paths = append(paths, abs)
-				} else {
-					logger.Warnf("Skipped path %q because it fails to get absolute path: %s", p, e)
+			line := strings.TrimSpace(scanner.Text())
+			if line == "" {
+				continue
+			}
+			// Parse: "path" or "path  start-end;start-end;..."
```
jiefenghuang: 文件名有可能包含 空格或'\t', 可以换一个特殊的组织方式, 直接正则解析

pkg/vfs/fill.go
```diff
@@ -316,7 +373,18 @@ func (iter *sliceIterator) hasNext() bool {
 		iter.nextChunkIndex++
 	}
 
-	return true
+	// Skip slices that don't overlap with any range
+	for iter.nextSliceIndex < uint64(len(iter.slices)) {
```
jiefenghuang: 这里可以简化, 还有这里提前中断了

hiltonhe:
@jiefenghuang  你好, 修改已重新提交, 非常感谢~

jiefenghuang (approved):

PR #7399, opened 2026-08-14
tools/lance-resolver: add standalone Lance dataset resolver for warmup

hiltonhe:
Depends on: #7398

### Motivation

Lance datasets span multiple files (manifests, data, transactions,
deletions). Users need to warm up the exact set of files for a specific
version, not blindly traverse the directory.

### Changes

Add `tools/lance-resolver/`, a standalone CLI tool that:
- Reads a Lance dataset and resolves all data files for a version
- Outputs paths in `juicefs warmup -f` compatible format
- Supports `--version`, `--manifest-only`, `--columns` flags

### Usage

```bash
lance-resolver /mnt/jfs/dataset.lance > /tmp/list.txt
juicefs warmup -f /tmp/list.txt
```

### Architecture

- No changes to core JuiceFS code (pure `tools/` addition)
- Zero external dependencies (protobuf code is vendored)
- Proto files from Lance project (Apache 2.0)

### Test Coverage

- `resolver_test.go`: 53 sub-cases covering manifest parsing, dataset
  resolution, error handling, and idempotency


jiefenghuang:
The warmup PR has already been merged. Please remove the changes that are no longer needed and resolve the conflicts. Also, the format supported by warmup has been adjusted.

hiltonhe:
@jiefenghuang Thanks for the review.

Done:
- Rebased onto the latest main.
- Removed the changes that were already merged in #7398.
- Updated the lance-resolver output format to match the new bracketed warmup syntax (`path [start-end;...]`).

Could you PTAL again? Thanks.

tools/lance-resolver/resolver.go
```diff
@@ -0,0 +1,485 @@
+/*
+ * JuiceFS, Copyright 2024 Juicedata, Inc.
```
jiefenghuang: 2026

tools/lance-resolver/resolver_test.go
```diff
@@ -0,0 +1,365 @@
+/*
+ * JuiceFS, Copyright 2024 Juicedata, Inc.
```
jiefenghuang: 2026

tools/lance-resolver/resolver.go
```diff
@@ -0,0 +1,485 @@
...
+			manifestLen := binary.LittleEndian.Uint32(manifestData[:4])
+			if uint32(len(manifestData)) >= 4+manifestLen {
+				manifest := &lancepb.Manifest{}
+				if err := proto.Unmarshal(manifestData[4:4+manifestLen], manifest); err != nil {
+					return nil, fmt.Errorf("unmarshal V2 manifest: %w", err)
+				}
+				return manifest, nil
+			}
+		}
+	}
```
jiefenghuang: **Blocking: this misparses every real Lance manifest and silently returns an empty `Manifest`.**

Lance always appends a 16-byte footer (`write_magics`): `[manifest_pos:i64 LE][major:i16][minor:i16]["LANC"]`, and the protobuf is located via `manifest_pos` — it is not guaranteed to start at offset 0.

Because of those trailing 16 bytes, the `>= 4+txnLen+4` heuristic matches *every* manifest. For the common `manifest_pos=0` layout it reads the footer as the "manifest section", gets `manifestLen=0`, and `proto.Unmarshal([]byte{}, ...)` **succeeds**, yielding a zero-value Manifest with `err == nil`.

Measured on identical manifests differing only by the real footer:
```
no footer   => err=nil version=7 fragments=2 fields=1   (ok)
with footer => err=nil version=0 fragments=0 fields=0   (all data files silently dropped)
```
End result: the tool prints only the manifest path and exits 0, so `warmup` gets an empty list with no warning.

Suggest dropping the heuristic entirely: read the trailing 16 bytes, validate `"LANC"`, take `manifest_pos` from `buf[len-16:len-8]`, read the u32 length there, and unmarshal `[manifest_pos+4 : size-16]`.

tools/lance-resolver/resolver.go
```diff
@@ -0,0 +1,485 @@
...
+	data, err := os.ReadFile(manifestPath)
+	if err != nil {
+		return nil, fmt.Errorf("read manifest %s: %w", manifestPath, err)
+	}
+
+	// V2 format: [txn_len:4][Transaction][manifest_len:4][Manifest][footer:16]
+	// Detect V2 by checking if there's a second section after the first protobuf
+	if len(data) >= 8 {
+		txnLen := binary.LittleEndian.Uint32(data[:4])
+		if uint32(len(data)) >= 4+txnLen+4 {
```
jiefenghuang: **Blocking: `uint32` overflow defeats the bounds check and panics.**

`4+txnLen+4` wraps in `uint32`. With the first 4 bytes = `F8 FF FF FF`, the guard passes and `data[4+txnLen:]` indexes at 4294967292:
```
panic: runtime error: slice bounds out of range [4294967292:128]
  main.readAndParseLanceManifest ... resolver.go:227
```
Reproduced against a 128-byte file. A truncated/corrupt manifest crashes with a stack trace instead of a diagnostic. Please widen to `uint64`/`int64` before comparing and bounds-check in 64-bit arithmetic.

tools/lance-resolver/resolver.go
```diff
@@ -0,0 +1,485 @@
...
+func parseLanceManifestBytes(data []byte) (*lancepb.Manifest, error) {
+	if len(data) < 4 {
+		return nil, fmt.Errorf("manifest too short: %d bytes", len(data))
+	}
+	pbLen := binary.LittleEndian.Uint32(data[:4])
+	if uint32(len(data)) < 4+pbLen {
+		return nil, fmt.Errorf("manifest truncated: need %d bytes, got %d", 4+pbLen, len(data))
+	}
+	manifest := &lancepb.Manifest{}
+	if err := proto.Unmarshal(data[4:4+pbLen], manifest); err != nil {
```
jiefenghuang: Same overflow class as above: `pbLen = 0xFFFFFFFF` makes `4+pbLen == 3`, so the `len(data) < 3` guard passes and `data[4:3]` panics:
```
panic: runtime error: slice bounds out of range [4:3]
```
Also note this hard-codes offset 0, so any manifest with a preceding index/transaction section unmarshals the wrong message — protobuf is permissive and will rarely error. Both should be fixed by seeking via the footer's `manifest_pos`.

tools/lance-resolver/resolver.go
```diff
@@ -0,0 +1,485 @@
...
+				f.Close()
+				continue
+			}
+
+			// Column metadata position and length are relative to decoded data,
+			// not raw file offsets. For V2 encoded files, we need to decode the
+			// encoding wrapper first. This is complex and format-specific.
+			// TODO: update for Lance V2 encoding format
+			f.Close()
+			continue
```
jiefenghuang: **Blocking: `--columns` makes the tool warm up *strictly less* than not passing it, silently.**

This loop always hits the TODO and `continue`s, so `results` is never appended to and is always empty. Combined with `colOnlyMode` at line 88 (which suppresses data-file/overlay paths), asking for specific columns drops the data files entirely:
```
$ lance-resolver ds              # 3 paths: manifest + 2 data files
$ lance-resolver --columns id ds # 1 path: manifest only, no ranges, exit 0
```
Until column resolution is implemented, please either fail loudly on `--columns`, or fall back to emitting the full data-file paths so the data is still warmed.

tools/lance-resolver/resolver.go
```diff
@@ -0,0 +1,485 @@
...
+				latestPath = path.Join(versionsDir, name)
+			}
+		} else {
+			// V2 naming: {inverted}.manifest
+			inverted := ^uint64(0) - latestVersion
+			invStr := fmt.Sprintf("%020d", inverted)
+			if base == invStr {
+				latestPath = path.Join(versionsDir, name)
+			}
+		}
```
jiefenghuang: **Blocking: this branch is dead code, and V2 datasets resolve to the *oldest* version.**

V2 names are `{u64::MAX-version:020}.manifest`, e.g. `18446744073709551610.manifest`. That string is `<= 2^64-1`, so `strconv.ParseUint` on line 193 **succeeds** and it takes the V1 branch, comparing the inverted value as a raw version. Larger inverted value means *smaller* real version, so `v > latestVersion` picks the oldest manifest.

Verified with a dir holding V2 manifests for versions 1..5:
```
SELECTED 18446744073709551614.manifest => version 1 (expected 5)
```
No error is returned — it silently resolves the wrong dataset version. Reached whenever `latest_version_hint.json` is missing or stale (older Lance writers, copied/rsynced datasets, `LANCE_USE_VERSION_HINT=0`).

Fix: in the V2 branch parse `base` as the inverted value, recover `v = ^uint64(0) - parsed`, and track the max `v`.

tools/lance-resolver/resolver.go
```diff
@@ -0,0 +1,485 @@
...
+			// TODO: update for Lance V2 encoding format
+			f.Close()
+			continue
+		}
+	}
+
+	return results, nil
+}
+
+func columnByteRanges(cm *file2pb.ColumnMetadata, includeDataPages bool) []byteRange {
```
jiefenghuang, line 519: CI lint will fail here. `.golangci.yml` sets `run.tests: false`, so references from `_test.go` don't count as usage. With the CI version (v2.6):
```
resolver.go:57:6:  type byteRange is unused (unused)
resolver.go:348:6: func columnByteRanges is unused (unused)
resolver.go:365:6: func mergeByteRanges is unused (unused)
```
`golangci-lint-action@v7` scans `./...`, so `tools/` is included. Note `file2pb` is referenced *only* here — if the `--columns` stub is removed, `proto/lance/file2/` (723 lines) can be dropped entirely. `lanceCmoEntrySize` and the `includeDataPages` param are unused too, which makes `--include-data-pages` a no-op flag.

tools/lance-resolver/resolver.go
```diff
@@ -0,0 +1,485 @@
...
+			printUsage()
+			return
+		default:
+			if !strings.HasPrefix(args[i], "-") {
+				// Dataset path
+				datasetPath := args[i]
+				if !isLanceDataset(datasetPath) {
+					fmt.Fprintf(os.Stderr, "Error: %s is not a Lance dataset (no _versions/ directory found)\n", datasetPath)
+					os.Exit(1)
+				}
```
jiefenghuang: Hand-rolled parsing returns on the first non-`-` arg, so anything after the path is silently ignored, and unknown `-` flags fall into `default` and are discarded:
```
lance-resolver ds --columns id     # --columns ignored, full listing, exit 0
lance-resolver ds ds2              # ds2 ignored, exit 0
lance-resolver ds --version        # missing value ignored, exit 0
lance-resolver --colums id,name ds # typo -> "id,name" treated as dataset path
```
Given the documented `lance-resolver <path> ... > list.txt` usage, this produces a wrong warmup list with exit code 0. Suggest using the stdlib `flag` package and erroring on unknown args.

jiefenghuang (commented):
Verified the tool against synthetic Lance datasets. The byte-range plumbing in #7398 looks good, but the resolver itself has a few blocking correctness issues — most importantly it fails on *real* Lance manifests, which the current tests don't catch because they write a layout Lance never produces. Details inline.

tools/lance-resolver/resolver_test.go
```diff
@@ -0,0 +1,365 @@
...
+		},
+	}
+
+	// Write manifest
+	pbData, _ := proto.Marshal(manifest)
+	buf := make([]byte, 4+len(pbData))
+	binary.LittleEndian.PutUint32(buf[:4], uint32(len(pbData)))
+	copy(buf[4:], pbData)
+	manifestPath := filepath.Join(versionsDir, "1.manifest")
+	os.WriteFile(manifestPath, buf, 0644)
```
jiefenghuang: This writes the manifest as exactly `4+len(pb)` bytes with **no 16-byte footer** — a layout Lance cannot produce, and the one input shape where the V2 heuristic happens not to misfire. Appending a real footer flips this test (and `TestResolveLanceDataset_ManifestOnly`) to an empty manifest, which is why the suite is green while the tool is broken on every real dataset.

Also worth covering: `readAndParseLanceManifest` (the function actually used at runtime) has no test at all — only the `parseLanceManifestBytes` fallback does; `resolveColumnByteRanges` is untested; `TestFindLatestManifestByListing` has no V2 case; and nothing asserts emitted output against `vfs.SplitTarget`.

Separately: `go test ./tools/...` isn't covered by any `Makefile` target, so these 365 lines never run in CI.

jiefenghuang:
The proto files can reference a specific version of the Lance project directly, so there is no need to introduce an additional copy.

The upstream go_package configuration currently has some issues. I think we can add a shell script that downloads the proto files from an explicitly pinned Lance version and then uses a specific version of protoc to generate the Go code.

Makefile
```diff
@@ -118,6 +118,12 @@ test.meta.non-core:
 test.pkg:
 	go test -tags gluster -v -cover -count=1  -failfast -timeout=12m $$(go list ./pkg/... | grep -v /meta) -args -test.gocoverdir="$(shell realpath cover/)"
 
+test.tools:
```
jiefenghuang: remove this, or move to tools/

jiefenghuang (approved):

PR #7400, opened 2026-08-15
ci: add a crash consistency test

solracsf:
Nothing currently checks that a mount which dies without unmounting keeps the data it said was durable. The chaos workflow runs vdbench under injected faults, but the configs set no data validation and the error check on its log is commented out, so a fault never fails the run. The kill -9 calls under `.github/scripts/command` kill a CLI command, not the mount.

The writer records a file only after `fsync()` on the file and `fsync()` on its parent directory both returned, into a manifest kept outside the mount and fsync'd per entry. Every entry is therefore a promise POSIX requires to survive a crash, and files that were written but not yet fsync'd are ignored, since losing those is allowed. The mount is killed with SIGKILL mid-write, remounted, and every recorded file is checked by size and sha256, followed by fsck for metadata pointing at objects that were never uploaded.

Each test then re-reads everything through an empty cache directory, so the blocks come from object storage rather than the local cache that survived the kill.

The `--writeback` test mounts with `--upload-delay` so that blocks acked by fsync are still sitting in the staging directory when the kill lands, which is what exercises the staging replay on the next mount. Without the delay the uploads keep up with the writer and the kill finds staging empty, so the run passes without covering the replay at all; the test fails instead of passing quietly in that case.

The mount process is matched by its mount point: CI runs a second juicefs mount for the coverage directory, and killing that one would leave the filesystem under test healthy.


.github/workflows/crash-consistency.yml
```diff
@@ -0,0 +1,106 @@
+name: "crash-consistency"
```
jiefenghuang: We can add this test to commands.yml as one of the steps.

Xuyuchao-juice (approved):

.github/workflows/crash-consistency.yml
solracsf: That's what I did in first time, but there were CI failures. 😞 
Feel free to check this.

jiefenghuang: https://github.com/juicedata/juicefs/actions/runs/31882545667/job/95007043421 this one? 
It looks like an issue with the CI environment and is unrelated to your changes. We can add it to command and see how it goes. Since this test is relatively time-consuming, running it once a day in CI should be sufficient.

solracsf: Yes it was it, and now the same one. I'll let you debug this 😉

PR #7402, opened 2026-08-15
object/file: reject object keys that escape the filestore root

### LINKED ISSUE #7401: Local filestore object backend writes outside its configured root via a crafted object key
Affected versions: confirmed on the tip of `main`, commit a617e1b (2026-08-14)

## Summary

`pkg/object/file.go`'s `filestore` type (the local-filesystem object storage backend, used directly via `file://` and as a common destination for `juicefs sync`) builds each operation's target path by joining an object key directly onto the configured root, with no validation. Since `juicefs sync <source> file:///local/destination/` is a documented, ordinary usage pattern, and the object keys enumerated during a sync come from the source backend's own listing (which, unlike a real local filesystem walk, has no character restrictions), a key containing `../` segments causes data to be written outside the intended local destination directory.

## Details

`pkg/object/file.go`, before the fix:
```go
func (d *filestore) path(key string) string {
	if strings.HasSuffix(d.root, dirSuffix) {
		return filepath.Join(d.root, key)
	}
	return filepath.Clean(d.root + key)
}

func (d *filestore) Put(ctx context.Context, key string, in io.Reader, getters ...AttrGetter) (err error) {
	p := d.path(key)
	...
}
```
`Get`, `Head`, `Delete`, `Chmod`, `Chown`, `Symlink`, and `Readlink` all use `d.path(key)` the same way, with no containment check.

## POC

(available upon request)

## Impact

Confirmed directly against `filestore.Put`: given a store rooted at a directory, calling `Put` with a key containing `../` segments writes a file completely outside the configured root, with no error returned. Since object keys enumerated from a source backend during `juicefs sync` (or any code path constructing a key from another object storage's own listing) have no character restrictions, a source the operator doesn't fully control the contents of, a shared or public bucket, or any bucket an attacker has write access to, can contain a key that writes attacker-chosen content to an attacker-influenced location outside the intended local destination directory.

I have a fix ready and will follow up with a PR.

carfeii:
Fixes #7401.

filestore builds each operation's target path by joining an object key directly onto the configured root, with no validation. Since `juicefs sync <source> file:///local/destination/` is a documented, ordinary usage pattern, and the object keys enumerated during a sync come from the source backend's own listing (which, unlike a real local filesystem walk, has no character restrictions), a key containing `../` segments causes data to be written outside the intended local destination directory.

Adds `safePath`, wrapping the existing `path` helper with a containment check (using `filepath.Rel` against the storage root, or the root's parent directory when root is used as a flat-namespace prefix rather than a directory, matching both of `path`'s existing modes). Updates all 8 call sites (`Put`, `Get`, `Head`, `Delete`, `Chmod`, `Chown`, `Symlink`, `Readlink`) to use it.

Adds a regression test (`TestFilestoreRejectsKeyPathTraversal`).

Verified against a scratch store: a key of `../outside/pwned.txt` written via `Put` escaped the configured root before this fix, confirmed via direct filesystem inspection; after the fix, the same call is rejected with a clear error and no file escapes. `go test ./pkg/object/... ./pkg/sync/...` all pass.

### DIFF
--- a/pkg/object/file.go
+++ b/pkg/object/file.go
@@ -45,7 +45,7 @@ type filestore struct {
 }
 
 func (d *filestore) Symlink(oldName, newName string) error {
-	p, err := d.safePath(newName)
+	p, err := d.path(newName)
 	if err != nil {
 		return err
 	}
@@ -60,7 +60,7 @@ func (d *filestore) Symlink(oldName, newName string) error {
 }
 
 func (d *filestore) Readlink(name string) (string, error) {
-	p, err := d.safePath(name)
+	p, err := d.path(name)
 	if err != nil {
 		return "", err
 	}
@@ -74,21 +74,13 @@ func (d *filestore) String() string {
 	return "file://" + d.root
 }
 
-func (d *filestore) path(key string) string {
+func (d *filestore) path(key string) (string, error) {
+	var p string
 	if strings.HasSuffix(d.root, dirSuffix) {
-		return filepath.Join(d.root, key)
+		p = filepath.Join(d.root, key)
+	} else {
+		p = filepath.Clean(d.root + key)
 	}
-	return filepath.Clean(d.root + key)
-}
-
-// safePath is like path, but rejects a key that would resolve outside the
-// storage root (or, when root is used as a flat-namespace prefix rather
-// than a directory, outside root's parent directory). An object key
-// commonly originates from another, possibly remote or shared, object
-// storage backend's own listing, which unlike a real local filesystem
-// walk has no character restrictions on what a key can contain.
-func (d *filestore) safePath(key string) (string, error) {
-	p := d.path(key)
 
 	boundary := d.root
 	if !strings.HasSuffix(boundary, dirSuffix) {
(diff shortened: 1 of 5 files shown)


pkg/object/file_traversal_test.go
```diff
@@ -0,0 +1,60 @@
+/*
+ * JuiceFS, Copyright 2018 Juicedata, Inc.
```
jiefenghuang: 2026, and put it in file_test.go?

pkg/object/file.go
```diff
@@ -74,8 +81,33 @@ func (d *filestore) path(key string) string {
 	return filepath.Clean(d.root + key)
 }
 
-func (d *filestore) Head(ctx context.Context, key string) (Object, error) {
+// safePath is like path, but rejects a key that would resolve outside the
```
jiefenghuang: Remove the comment, or just keep a simple one-line explanation. This can be fixed directly in the path function. Also, there is another chtime case that hasn’t been fixed yet.

jiefenghuang:
Hi @carfeii please sign the cla.

PR #7407, opened 2026-08-17
meta: reap detached trees that contain immutable files

solracsf:
Clone copies the source node wholesale, flags included, so cloning a tree with a `chattr +i` file produces immutable copies. When a clone is interrupted the top directory stays detached, and its cleanup goes through `emptyDir -> BatchUnlink/Rmdir`, which refuse entries carrying `FlagImmutable` or `FlagAppend` (uid 0 does not bypass that). `CleanupDetachedNodesBefore` then failed on the tree on every `gc --delete`, logging an error each run while the tree kept its slice references alive, so those blocks could never be reclaimed.

Mark the context used to remove a detached tree and let the flag checks in `doBatchUnlink` and `doRmdir` stand down for it, in all three engines. A detached tree is unreachable and this cleanup is its only way out.

Removal of reachable files is untouched: the marker is set only by `doCleanupDetachedNode`.

### DIFF
--- a/pkg/meta/base.go
+++ b/pkg/meta/base.go
@@ -3133,6 +3133,21 @@ func (m *baseMeta) cleanupTrash(ctx Context) {
 	}
 }
 
+type ignoreAttrFlagsKey struct{}
+
+// A detached tree is unreachable, so it can only be removed by the cleanup of
+// detached nodes; immutable/append flags on the entries it contains (copied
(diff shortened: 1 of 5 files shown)


solracsf:
@jiefenghuang can you kindly review this PR. Thanks 👍

jiefenghuang (approved):

PR #7408, opened 2026-08-17
vfs: align ioctl flag permissions with the kernel

solracsf:
`FS_IOC_SETFLAGS` carries the whole resulting flag mask, so clearing a protected flag arrives as a missing bit rather than a set one. The guard only rejected a non-root caller whose incoming mask contained `FS_SECRM_FL`, `FS_IMMUTABLE_FL` or `FS_APPEND_FL`, which caught `chattr +i` but not `chattr -i`: a user who could reach a file root had made immutable was able to clear the flag and then modify or delete it.

Compare the requested flags with the current ones and refuse a non-root caller when a protected bit changes in either direction, which is what the kernel does in `fileattr_set_prepare()`. A write that leaves every protected flag as it was stays unprivileged, so re-applying a flag that is already set no longer fails.

Windows is unaffected: `FILE_ATTRIBUTE_READONLY` goes through `ChFlags` in vfs_windows.go, not this `ioctl`.

### DIFF
--- a/docs/en/security/trash.md
+++ b/docs/en/security/trash.md
@@ -123,6 +123,8 @@ It is possible to skip the trash and permanently delete files directly. The 's' 
 
 You will need to enable the mount option `--enable-ioctl` to allow adjusting file attributes using `chattr`.
 
+JuiceFS follows the permission rules the Linux kernel applies to `chattr`: you must own a file to change its attributes, and while the 's' flag only requires ownership, the 'i' (immutable) and 'a' (append-only) flags can only be set or cleared by root.
+
 ## Trash and slices {#gc}
 
 Apart from user deleted files, there's another type of data which also resides in Trash, which isn't directly visible from the `.trash` directory, they are stale slices created by file edits and overwrites. Read more in [How JuiceFS stores files](../introduction/architecture.md#how-juicefs-store-files). To sum up, if applications constantly delete or overwrite files, object storage usage will exceed file system usage.
--- a/docs/zh_cn/security/trash.md
+++ b/docs/zh_cn/security/trash.md
@@ -135,6 +135,8 @@ juicefs rmr .trash/2022-11-30-10/
 
 需要在挂载时启用`--enable-ioctl`选项，才能使用 chattr 命令修改文件属性。
 
+JuiceFS 遵循 Linux 内核对 chattr 的权限规则：只有文件属主才能修改其属性；其中's'属性属主即可设置，而'i'（immutable）和'a'（append-only）属性只有 root 才能设置或清除。
+
 ## 回收站和文件碎片 {#gc}
 
 在回收站里，除了因用户操作而产生的文件，还存在另一类对用户不可见的数据——覆写产生的文件碎片。关于文件碎片是怎么产生的，可以详细阅读[「JuiceFS 如何存储文件」](../introduction/architecture.md#how-juicefs-store-files)。总而言之，如果应用经常删除文件或者频繁覆盖写文件，会导致对象存储使用量远大于文件系统用量。
--- a/pkg/vfs/vfs_unix.go
+++ b/pkg/vfs/vfs_unix.go
@@ -347,31 +347,39 @@ func (v *VFS) Ioctl(ctx Context, ino Ino, cmd uint32, arg uint64, bufIn, bufOut 
 		} else {
 			return syscall.EINVAL
 		}
+		var flags uint8
 		if (iflag & FS_SECRM_FL) != 0 {
-			attr.Flags |= meta.FlagSkipTrash
+			flags |= meta.FlagSkipTrash
 		}
 		if (iflag & FS_IMMUTABLE_FL) != 0 {
-			attr.Flags |= meta.FlagImmutable
+			flags |= meta.FlagImmutable
 		}
 		if (iflag & FS_APPEND_FL) != 0 {
-			attr.Flags |= meta.FlagAppend
-		}
-		if ctx.CheckPermission() && ctx.Uid() != 0 {
-			// chattr sends the whole resulting mask, so clearing a protected flag
-			// arrives as a missing bit rather than a set one; compare against the
-			// current flags to catch both directions.
-			const protected = meta.FlagSkipTrash | meta.FlagImmutable | meta.FlagAppend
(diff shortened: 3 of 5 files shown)


jiefenghuang:
Thanks for catching this — the direction is right, and I pushed a follow-up commit to the branch that extends it. Here is the full picture of how this permission check works, since it depends on the kernel version in a way that is easy to miss.

### How `FS_IOC_SETFLAGS` reaches a FUSE filesystem

The kernel dispatches ioctls by exact command number in `do_vfs_ioctl()`, which gives two very different paths:

**A. The kernel screens the request.** `case FS_IOC_SETFLAGS` → `ioctl_setflags()` → `vfs_fileattr_set()`:

1. `inode_owner_or_capable()` — the caller must own the inode (or hold `CAP_FOWNER`);
2. `vfs_fileattr_get()` — fetch the *current* flags from the filesystem;
3. `fileattr_set_prepare()`:
   ```c
   if ((fa->flags ^ old_ma->flags) & (FS_APPEND_FL | FS_IMMUTABLE_FL) &&
       !capable(CAP_LINUX_IMMUTABLE))
           return -EPERM;
   ```
4. only then is `FS_IOC_SETFLAGS` forwarded to the filesystem.

FUSE joined this path in **Linux 5.13**, when `fuse_fileattr_get`/`fuse_fileattr_set` were added to `fs/fuse/dir.c`.

**B. The request is passed straight through.** Any command number that does not match a `case` falls into `default: → file_ioctl() → fuse_file_ioctl()` and is handed to the filesystem verbatim. **The kernel checks nothing at all.** This happens when:

- the kernel is **older than 5.13** — that covers *every* `FS_IOC_SETFLAGS` on FUSE there (CentOS/RHEL 7 and 8, Ubuntu 18.04/20.04, …);
- a 64-bit caller sends **`FS_IOC_SETFLAGS_32` (`0x40046602`)** on any kernel. The 32→64 translation only exists in `COMPAT_SYSCALL_DEFINE3(ioctl)` and therefore only applies to 32-bit processes, yet we must accept that number because it is the canonical one on a 32-bit kernel.

So on path B **JuiceFS is the only thing standing between an unprivileged user and the immutable flag.**

### What that means for the bug

I reproduced this on Linux 6.8 with an ext4 reference. On 6.8 a plain `chattr -i` takes path A, so the kernel rejects it before we ever see it — the access log shows only the `FSGETXATTR` that fetches the old value, and no `SETFLAGS` at all:

```
[uid:0]    ioctl (2,0x801C581F,...) - OK              # fetch old value
[uid:0]    ioctl (2,0x40086602,[16 0 0 0],[]) - OK    # root sets immutable
[uid:1000] ioctl (2,0x801C581F,...) - OK              # fetch old value
                                                      # kernel returns EPERM; SETFLAGS never arrives
```

Driving the same sequence over `0x40046602` puts it on path B, which is exactly what an older kernel does for a normal `chattr`. There, before this PR:

| | non-root clears immutable | resulting flags | then writes | then deletes |
| --- | --- | --- | --- | --- |
| ext4 | rejected | still set | rejected | rejected |
| JuiceFS (before) | **succeeds** | **cleared** | **succeeds** | **succeeds** |

### Why the original patch was not enough

The new guard sat behind `ctx.CheckPermission()`, and in `pkg/fuse/context.go`:

```go
ctx.checkPermission = fs.conf.NonDefaultPermission && header.Uid != 0
```

A default mount passes `default_permissions` to the kernel and leaves `NonDefaultPermission` false, so `CheckPermission()` is always false and **the whole guard was skipped on the most common mount configuration**. With the patched binary a non-root user could still clear immutable and delete the file on a default mount; it only behaved correctly under `--non-default-permission`.

Two further gaps: there was no ownership check at all (the kernel does that first), and the SET branch built `Attr{}` from scratch while `mergeAttr` assigns `Flags` wholesale, so a `chattr` silently wiped the Windows attributes (`Hidden`/`System`/`Archive`) — and symmetrically a Windows `Chflags` wiped `Append`/`SkipTrash`.

### What the follow-up commit does

Reproduce what `vfs_fileattr_set()` does, unconditionally:

```go
if ctx.Uid() != 0 {
    if ctx.Uid() != cur.Uid {          // inode_owner_or_capable()
        return syscall.EPERM
    }
    if (cur.Flags^flags)&privileged != 0 {  // fileattr_set_prepare()
        return syscall.EPERM
    }
}
attr.Flags = cur.Flags&^mapped | flags
```

Resulting behaviour:

| caller | `s` (skip-trash) | `i` / `a` |
| --- | --- | --- |
| root | allowed | allowed |
| non-root owner | allowed | `EPERM` |
| non-owner | `EPERM` | `EPERM` |

- The comparison is on the **delta**, so it catches setting *and* clearing, and a rewrite that leaves the privileged flags untouched stays unprivileged — same as `fileattr_set_prepare()`.
- `s` is deliberately *not* privileged: the kernel does not gate `FS_SECRM_FL` either, and requiring root would have taken `chattr +s` away from ordinary owners on a default mount, which the trash documentation tells them to use.
- Both `Ioctl` and the Windows `ChFlags` now preserve the flags they do not map.
- The check stays in the ioctl path rather than moving into `mergeAttr`, because `mergeAttr` is shared with WinFsp, where `FILE_ATTRIBUTE_READONLY` maps onto `FlagImmutable` and must remain settable by a normal user.

Tests cover: privileged flags in both directions, the no-op rewrite, `s` being allowed for an owner, a non-owner being refused (including a no-op write), preservation of unmapped flags, and `ENOTSUP` for unsupported bits.

### Known limits

- **TOCTOU**: the kernel holds `inode_lock` across get/check/set; we cannot, so there is a small window between `GetAttr` and `SetAttr`. Exploiting it needs a concurrent privileged writer, and closing it would mean pushing the rule into the metadata transaction, which would change the Windows semantics above.
- **`CAP_FOWNER`**: a non-root process holding it is accepted by the kernel on path A but refused by our ownership check, since capabilities are not visible over FUSE. Rare, and it fails closed.

### Backport

`release-1.4` and `release-1.3` carry the same one-directional guard and are affected the same way. `release-1.3` has no `FS_SECRM_FL` support, so the mapped/privileged masks need adjusting there.

jiefenghuang (approved):

PR #7417, opened 2026-08-19
mount: fix avoid interrupting requests during graceful shutdown

zxh326:
ref: https://github.com/juicedata/go-fuse/pull/58


cmd/mount_unix.go
jiefenghuang, line 686: 不是中断重启流程吗

jiefenghuang (approved):

cmd/mount_unix.go
jiefenghuang: mark: 后续重构, 整个容错处理再梳理

PR #7419, opened 2026-08-19
cache: exclude staged blocks from the cached block count

### LINKED ISSUE #7418: staged blocks are counted as cache blocks, but not as cache bytes
With `--writeback`, `blockcache_blocks` counts staged blocks. `blockcache_bytes` does not.

After writing 200 blocks of 4 KiB, before they are uploaded:

```
juicefs_blockcache_blocks 200
juicefs_blockcache_bytes 0
juicefs_staging_blocks 200
juicefs_staging_block_bytes 819200
```

The same 200 blocks are already reported by `staging_blocks`, so they are counted twice.

The `--cache-items` limit counts staged blocks too. `--cache-size` does not.

With `--cache-items 50` and the same 200 staged blocks, reading 100 files:

```
juicefs_blockcache_blocks 201
juicefs_blockcache_bytes 8192
juicefs_blockcache_evicts 99
juicefs_blockcache_hits 0
juicefs_staging_blocks 200
```

Staged blocks cannot be evicted, so the count never gets below the limit. The cleanup keeps removing read cache blocks, and only 1 of the 100 survived.

Reproduced with 1.4.1 and current main (d178e3d).

Looka149:
Resolves #7418

The cache now keeps its own block count, updated wherever `used` is updated. Neither counts staged blocks.

`blockcache_blocks` reports this count, so it now matches `blockcache_bytes`.

`--cache-items` uses it too, so the limit now applies to read cache blocks only. Staged blocks still take disk space, and `--free-space-ratio` still limits that.

Tested on a kind cluster with `--writeback`, 200 staged blocks and `--cache-items 50`: reading 100 files now keeps 50 read cache blocks instead of 1, and `staging_blocks` stays 200.

### DIFF
--- a/docs/en/reference/p8s_metrics.md
+++ b/docs/en/reference/p8s_metrics.md
@@ -68,8 +68,8 @@ If you haven't yet set up monitoring for JuiceFS, read [monitoring and data visu
 
 | Name                                    | Description                                 | Unit   |
 |:----------------------------------------|---------------------------------------------|--------|
-| `juicefs_blockcache_blocks`             | Number of cached blocks                     |        |
-| `juicefs_blockcache_bytes`              | Size of cached blocks                       | byte   |
+| `juicefs_blockcache_blocks`             | Number of cached blocks (staging excluded)  |        |
+| `juicefs_blockcache_bytes`              | Size of cached blocks (staging excluded)    | byte   |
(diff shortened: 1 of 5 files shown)


CaitinChen:
@jiefenghuang PTAL

Looka149:
@CaitinChen @jiefenghuang I updated the issue and the PR description. Thank you in advance for your review.

PR #7424, opened 2026-08-20
cmd/gc: wait slice deletions before scanning objects

### LINKED ISSUE #7420: optimise gc cleanup
Extract the GC cleanup logic into a separate function, decoupled from ScanSlice, and make the cleanup progress observable.

**Why is this needed**:
The current cleanup process may cause GC to falsely report leaked objects.

zxh326:
fix: https://github.com/juicedata/juicefs/issues/7420

### DIFF
--- a/pkg/meta/interface.go
+++ b/pkg/meta/interface.go
@@ -504,7 +504,7 @@ type Meta interface {
 	Compact(ctx Context, inode Ino, concurrency int, preFunc, postFunc func()) syscall.Errno
 
 	// CleanupSlices performs metadata cleanup and schedules slice deletion tasks.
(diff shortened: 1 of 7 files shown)


pkg/meta/interface.go
```diff
@@ -503,6 +503,10 @@ type Meta interface {
 	// Compact chunks for specified path
 	Compact(ctx Context, inode Ino, concurrency int, preFunc, postFunc func()) syscall.Errno
 
+	// CleanupSlices performs metadata cleanup and schedules slice deletion tasks.
```
jiefenghuang, line 506: 去除参数delete, CleanupPendingSlice?

jiefenghuang (approved):

PR #7428, opened 2026-08-20
S3/gateway: fix DeleteObjects API wrongly delete sibling object named as bucket

### LINKED ISSUE #7426: S3/gateway: deleting an object may remove another object named after its bucket
**What happened**:

The scheduled `gateway-random` workflow found inconsistent behavior between MinIO and the JuiceFS S3 gateway when deleting an object.

When `DeleteObjects` deletes an object located directly under a bucket, it may pass an absolute filesystem path back to `delObj` as an object key. The bucket name is then joined twice. As a result, an unrelated object whose key equals the bucket name may be deleted.

For example, deleting `jmrq/rror` may unexpectedly delete `jmrq/jmrq`.

CI run:
https://github.com/juicedata/juicefs/actions/runs/31273806491

Failed job:
https://github.com/juicedata/juicefs/actions/runs/31273806491/job/93144051225

The minimized test case creates bucket `jmrq`, then creates two objects:

- object `jmrq`, whose key is the same as the bucket name
- object `rror`

After deleting `rror`, the JuiceFS gateway also deletes object `jmrq` unexpectedly. Therefore, removing the bucket succeeds on JuiceFS while MinIO correctly reports that the bucket is still non-empty.

The final assertion was:

```text
remove_bucket:
result1 is returncode:1 output:mc: <ERROR> `minio_admin/jmrq` is not empty.
Retry this command with ‘--force’ flag if you want to remove
`minio_admin/jmrq` and all its contents

result2 is True
```

myangitzh:
Fixes #7426

### DIFF
--- a/pkg/gateway/gateway.go
+++ b/pkg/gateway/gateway.go
@@ -487,29 +487,48 @@ func (n *jfsObjects) DeleteObject(ctx context.Context, bucket, object string, op
 	return info, jfsToObjectErr(ctx, err, bucket, object)
 }
 
+// pruneEmptyDirs removes dir and its now-empty ancestor directories, going up
+// until (but not including) the bucket root. It only operates within the bucket
+// subtree and stops at the bucket root, a still non-empty or already removed
+// directory, or an explicit directory object (marked with atime==0, kept).
+func (n *jfsObjects) pruneEmptyDirs(root, dir string) error {
+	// A path-boundary prefix so a sibling that merely shares a name prefix (e.g.
+	// bucket "foobar" vs root "/foo") is not treated as inside the bucket subtree.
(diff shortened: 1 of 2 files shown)


pkg/gateway/gateway.go
```diff
@@ -544,7 +559,7 @@ func (n *jfsObjects) DeleteObjects(ctx context.Context, bucket string, objects [
 				}
 				return err
 			}
-			if e := n.delObj(bucket, ppath); e != nil {
+			if e := n.pruneEmptyDirs(root, ppath); e != nil {
```
zhijian-pro, line 567: 如果直接原因是路径拼错了，那把路径拼正确应该就可以解决问题

zhijian-pro (approved):

PR #7438, opened 2026-08-22
KEXT-less FUSE support for MacOS: FUSE-T (NFS, SMB, FSKit) + MacFUSE-FSKit backends via cgofuse

anasalzuvix:
JuiceFS can now be mounted on macOS **without any kernel extension** through **FUSE-T**
with three selectable transports (`-o backend=nfs|smb|fskit`), plus the macFUSE-FSKit
path through the same cgofuse adapter (`pkg/fusemac`).

**Procedure, and Technique:**

1. **Path-based frontend over `vfs.VFS`** — a Darwin-only adapter (`pkg/fusemac`) implements
   the cgofuse `FileSystemInterface` on top of the shared VFS layer. cgofuse's callback set
   (no single `SetAttr`; separate `Chmod`/`Chown`/`Utimens`) is translated onto VFS's
   bitmask-style `SetAttr`.
2. **Backend selection & routing** — default mounts stay on macFUSE-KEXT via go-fuse;
   `-o backend=` routes through cgofuse. Runtime library detection refuses impossible
   combinations instead of failing mysteriously mid-mount.
3. **Upstream-contract verification** — each fix was checked against the pinned dependency
   sources (cgofuse v1.6.0 `hostUtimens`, dlopen order, commented-out Lock op) rather than
   assumed from release notes.

### 1. cgofuse dylib selection escape hatch is missing from our pinned version

Upstream cgofuse added `CGOFUSE_LIBFUSE_PATH`, which lets the application choose the FUSE-T
dylib explicitly when both FUSE-T and macFUSE are installed. Because we pin cgofuse v1.6.0
(which predates this), that variable is unavailable — and cgofuse v1.6.0 dlopens macFUSE's
`libfuse.2.dylib` *before* `libfuse-t.dylib`, silently hijacking every `backend=` mount on
dual installs. Until we upgrade cgofuse, `CheckBackend` fails closed on those combinations
and asks users to remove the macFUSE dylibs from `/usr/local/lib`.

### 2. pjdfstest tests - FUSE-T FSKit wins on compliance, but workloads reveal the true performance.

| Backend       | Total Tests | Failed | Passed |
| ------------- | ----------: | -----: | -----: |
| FUSE-T fskit  |        8686 |   4835 |   3851 |
| MacFUSE KEXT  |        8686 |   6919 |   1767 |

MacFUSE KEXT scores lower than FUSE-T in pjdfstest, but remains the battle-tested option
under high concurrency due to its mature kernel-mode implementation, advantages the
still-maturing FSKit backends cannot match yet. FUSE-T NFS handled all high-workload
scenarios correctly, with throughput bounded by the local network interface rather than by
correctness or stability.

#### Recommendation: switching backends for heavy workloads

If high-load work fails with the FUSE-T `fskit` backend, it is recommended to temporarily
switch to the NFS backend (`-o backend=nfs`) or (`-o backend=smb`), perform the high-chunk work there, and then
switch back to `fskit`. While NFS,SMB  is slower than FSKit, it handled all high-workload
scenarios correctly — making it the safer choice for heavy chunk operations until the FSKit
backend matures.

### 3. What the kext-less bridges do not expose

**a) Protocol-level operations** that FSKit/NFS/SMB bridges do not forward:

- flock/fcntl locks (per-host)
- fallocate
- ioctl
- copy_file_range
- SEEK_HOLE / SEEK_DATA
- chflags forwarding

**b) Maintenance subcommands** that depend on the `.control` request/response channel,
whose write-then-read pattern the kext-less bridges cannot service reliably:

> Affected JuiceFS subcommands: `info`, `rmr`, `clone`, `compact`, `summary`, `warmup`.

## Additional Hardening Included

- **Fail-closed backend guard**: cgofuse loads macFUSE's library first; `nfs`/`smb`
  error out on dual installs, and `backend=fskit` requires explicit
  `-o allow_unsafe_fskit_macfuse` because macFUSE-FSKit's write path delivers zeroed
  buffers upstream.
- **xattr parity with go-fuse**: `--enable-xattr` gates the xattr callbacks
  (ENOSYS when off).
- **chflags**: native `UF_IMMUTABLE`/`UF_APPEND` mapping, owner-or-root enforcement,
  preservation of unrelated meta flag bits, round-trip via `st_flags`.
- **utimens**: single-sided `UTIME_NOW`/`UTIME_OMIT` sentinels translated instead of
  written as timestamps (prevents corruption on macFUSE).
- **Unmount safety**: data + meta session flush **before** teardown; a SIGPIPE-killed
  child during FUSE-T cleanup is classified by the supervisor as expected teardown
  (data already durable).

---

## Verification

Tested on macOS platform:

| Item   | Value                                                                                     |
| ------ | ----------------------------------------------------------------------------------------- |
| Arch   | darwin arm64                                                                              |
| Kernel | Darwin 25.5.0; root:xnu-12377.121.10~1/RELEASE_ARM64_T8103 arm64                          |
| OS     | macOS 26.5.2 (BuildVersion 25F84)                                                         |

- ^C durability: unsynced 8 MB write survives immediate unmount + remount
- Supervised signal flow matches the go-fuse/macFUSE behavior end-to-end

## Known Limitations

- Per-host file locks
- No fallocate / ioctl / copy_file_range / SEEK_HOLE on kext-less backends
- Birthtime reported as ctime
- chflags forwarding dormant on FUSE-T (client-side refusal)
- macFUSE-FSKit unusable until upstream fixes its write path [#1187](https://github.com/macfuse/macfuse/issues/1187)

### DIFF
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -649,7 +649,7 @@ func mount(c *cli.Context) error {
 			foreground = os.Getppid() == 1 && !insideContainer()
 		}
 		if foreground {
-			go checkMountpoint(format.Name, mp, c.String("log"), false)
+			go checkMountpoint(format.Name, mp, c.String("log"), false, vfsConf.Fusemac)
 		} else {
 			daemonRun(c, addr, vfsConf) // only stage 0 needs the vfsConf
 		}
--- a/cmd/mount_unix.go
+++ b/cmd/mount_unix.go
@@ -209,7 +209,7 @@ func readThreadStats(pid int, tid string, b *strings.Builder) {
 	}
 }
 
-func watchdog(ctx context.Context, mp string) {
+func watchdog(ctx context.Context, mp string, fusemac bool) {
 	var lastActive int64 = time.Now().Unix()
 	var pid int
 	var agentAddr string
@@ -224,7 +224,8 @@ func watchdog(ctx context.Context, mp string) {
 			var confStat syscall.Stat_t
 			err := syscall.Stat(filepath.Join(mp, confName), &confStat)
 			ino, _ := vfs.GetInternalNodeByName(confName)
-			if err == nil && confStat.Ino == uint64(ino) {
+			// fusemac remaps internal-node inodes; fall back to a regular-file check.
+			if err == nil && (confStat.Ino == uint64(ino) || (fusemac && runtime.GOOS == "darwin" && confStat.Mode&syscall.S_IFMT == syscall.S_IFREG)) {
 				if dev == 0 && runtime.GOOS == "linux" {
 					var st syscall.Stat_t
 					if err := syscall.Stat(mp, &st); err == nil && st.Ino == 1 {
@@ -297,7 +298,7 @@ func parseFuseFd(mountPoint string) (fd int) {
 	return fd
 }
 
-func checkMountpoint(name, mp, logPath string, background bool) {
+func checkMountpoint(name, mp, logPath string, background bool, fusemac bool) {
 	if parseFuseFd(mp) > 0 {
 		logger.Infof("\033[92mOK\033[0m, %s with special mount point %q", name, mp)
 		return
@@ -316,7 +317,8 @@ func checkMountpoint(name, mp, logPath string, background bool) {
 		time.Sleep(time.Duration(interval) * time.Millisecond)
 		st, err := os.Stat(mp)
 		if err == nil {
-			if sys, ok := st.Sys().(*syscall.Stat_t); ok && sys.Ino == uint64(meta.RootInode) {
+			// fusemac may report the root as fileid 2 instead of 1.
+			if sys, ok := st.Sys().(*syscall.Stat_t); ok && (sys.Ino == uint64(meta.RootInode) || (fusemac && sys.Ino == 2)) {
 				// in pod, pid probably the same
 				if csiCommPath == "" && oldConf != nil {
 					_, newConf, _ := loadConfig(mp)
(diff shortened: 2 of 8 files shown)


jiefenghuang:
```
3. What the kext-less bridges do not expose
a) Protocol-level operations that FSKit/NFS/SMB bridges do not forward:

flock/fcntl locks (per-host)
fallocate
ioctl
copy_file_range
SEEK_HOLE / SEEK_DATA
chflags forwarding
b) Maintenance subcommands that depend on the .control request/response channel,
whose write-then-read pattern the kext-less bridges cannot service reliably:

Affected JuiceFS subcommands: info, rmr, clone, compact, summary, warmup.
```

It looks like the current implementation is still relatively immature and may have quite a few compatibility issues, so we’ll leave this pending for now.

This feature would ideally be implemented in the upstream go-fuse layer as well, rather than in JuiceFS. There was a previous PR for this: https://github.com/hanwen/go-fuse/pull/453, but it was closed. We’ll continue to keep an eye on any further progress, although it looks like there haven’t been any updates for about three years.

anasalzuvix:
Yeah, completely agree that it's still pretty immature. To be honest, fixing all of these edge cases on my own isn't really doable right now, my knowledge of JuiceFS core architecture is pretty limited, so I'm hitting a wall there 

Regarding hanwen/go-fuse#453: the main issue with going that route is that Hanwen was looking at implementing a native NFS server directly in go-fuse rather than wrapping FUSE-T, but the effort stalled out years ago.

Even if I opened a new PR against upstream go-fuse today, I don't think it would solve this cleanly. Modern FUSE-T also supports FSKit as a backend now, which feels like it would be a huge headache to implement directly inside go-fuse. That's why I went with winfsp/cgofuse instead, it already handles the abstraction for FUSE-T (all backend), MacFUSE FSkit backends out of the box. I don't have enough deep expertise with go-fuse internals to know if porting all of that over is even possible, since most of my work has been around cgofuse library.

Totally understand putting this on hold for now. But having to reboot into Recovery Mode just to lower security settings for MacFUSE is becoming a massive pain point for macOS users. Whenever you or someone on the core team has time to look deeper into native kext-less support, I'd be more than happy to help test things out or contribute where I can. 🤞

zxh326:
Thanks for the detailed background.

I put together a small proof of concept using go-fuse with macFUSE’s MFMount transport. It confirmed that a kext-less path is technically possible without going through FSKit directly. That said, lifecycle management, error handling, and the overall FSKit service behavior still seem pretty immature.

For now, we’d prefer not to introduce cgofuse as a second FUSE stack in JuiceFS. A better long-term direction would be a mature, actively maintained integration layer — ideally support upstream in go-fuse — so that JuiceFS only needs minimal changes.

We’ll hold off on this PR for now. Thanks again for your work and for offering to help test things when we revisit this.

PR #7440, opened 2026-08-23
security: authenticate distributed sync and SFTP peers

aburan28:
## Summary
- protect the distributed sync manager with per-run TLS 1.3 certificate pinning and bearer authentication
- bound and validate worker HTTP state, keep pprof unreachable, and enforce SSH host verification
- make SFTP host verification fail closed and reject traversal keys

## Verification
- go test ./pkg/sync ./pkg/object
- go vet ./pkg/sync ./pkg/object
- command and documentation compatibility reviewed


jiefenghuang:
This increases the complexity and makes it more difficult to use, so we’ll close this for now.

PR #7441, opened 2026-08-23
security: harden local object and ciphertext boundaries

aburan28:
## Summary
- anchor local object operations to an os.Root boundary so symlink ancestors cannot redirect writes outside storage
- preserve legitimate final-symlink and permission semantics with descriptor-relative operations
- reject malformed AEAD nonce lengths without panicking

## Verification
- go test ./pkg/object
- go test ./pkg/sync
- go vet ./pkg/object ./pkg/sync


jiefenghuang:
duplicate https://github.com/juicedata/juicefs/pull/7425

PR #7442, opened 2026-08-23
security: fail closed on service authentication

aburan28:
## Summary
- reject incomplete WebDAV credential and TLS configurations before service initialization
- restrict the mounted internal control interface to the mount owner or root
- document the updated operator requirements in English and Chinese

## Verification
- go test ./pkg/fs
- go test ./pkg/vfs -run ^TestInternalFile$ -count=1
- go test ./cmd -run ^$
- go vet ./pkg/fs ./pkg/vfs ./cmd


jiefenghuang:
It affects many commands, so we’ll close it.

PR #7443, opened 2026-08-23
security: isolate extracted Java native libraries

aburan28:
## Summary
- stop trusting or reusing predictable native library files from the shared temporary directory
- extract each bundled library into a unique private directory and fail closed if the resource is missing
- add an isolated regression test for uniqueness, symlink resistance, and POSIX permissions

## Verification
- JAVA_HOME=/opt/homebrew/opt/openjdk@21 mvn -Dtest=NativeLibraryDirectoryTest test
- Java main and test sources compile successfully


tangyoupeng:
Sharing the same libjfs library can speed up initialization when running multiple SDK instances on the same machine

PR #7452, opened 2026-08-25
S3/gateway: fix DeleteObjects API wrongly delete sibling object named as bucket

### LINKED ISSUE #7426: S3/gateway: deleting an object may remove another object named after its bucket

myangitzh:
Fix #7426

### DIFF
--- a/pkg/gateway/gateway.go
+++ b/pkg/gateway/gateway.go
@@ -526,6 +526,7 @@ func (n *jfsObjects) DeleteObjects(ctx context.Context, bucket string, objects [
 		p := path.Dir(path.Clean(n.path(bucket, o.ObjectName)))
 		delMap[p] = append(delMap[p], idx)
 	}
+	root := n.path(bucket)
 	var g errgroup.Group
 	g.SetLimit(runtime.NumCPU())
 	for ppath := range delMap {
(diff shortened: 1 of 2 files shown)


myangitzh:
track in #7428，close

PR #6445, opened 2025-11-04
hadoop: support kerberos

tangyoupeng:
Usage:
```shell
./juicefs config redis://127.0.0.1:6379/0 --kerberos-config-file kerb.cfg
```
kerb.cfg
```
# kerberos keytab
dev.keytab={BASE64 KEYTAB}
# delegation token
dev.token.life=604800
dev.token.renew=86400

# superuser and supergroup
dev.superuser=client
dev.supergroup=supergroup

# Mapping from Kerberos principals to OS user accounts
# https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-common/SecureMode.html#Mapping_from_Kerberos_principals_to_OS_user_accounts
dev.mechanism=hadoop
dev.rule=RULE:[2:$1/$2@$0](root/.*@example.com)s/.*/hdfs/
dev.rule=RULE:[2:$1/$2@$0](jerry/.*@EXAMPLE\.COM)s/.*/jerry_map/
dev.rule=RULE:[2:$1/$2@$0](tom/.*@EXAMPLE\.COM)s/.*/client/
dev.rule=DEFAULT

# proxy user settings
# https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-common/SecureMode.html#Proxy_user
# users: user1,user2 or *
dev.proxy.client.users=foo
# groups: group1,group2 or *
dev.proxy.client.groups=foogrp
# hosts: host1,host2 or 192.168.1.1,192.168.1.2 or 192.168.1.1/32 or *
dev.proxy.client.hosts=*
```

tested:
- [x] hadoop fs 
- [x] hive
- [x] spark
- [x] ranger

PR #7454, opened 2026-08-26
hadoop: skip kerberos authentication when server principal is not configured

kyungwan-nam:
### Problem

On a kerberized cluster,`hdfs dfs -ls jfs://{VOL}/` fails right after kinit:

`-ls: Empty nameString not allowed`

Without kinit, the same command works and this does not happen with the 1.3 Hadoop SDK
The code path was introduced in 1.4 by #6445

### Cause

`buildAuthCredential()` decides to build a Kerberos AP-REQ based only on whether the login user holds Kerberos credentials. When `juicefs.server-principal` is not configured, spn is the empty string, so `GSSManager.createName("", GSSName.NT_USER_NAME, ...)` in `KerberosUtil.genApReq()` throws `IllegalArgumentException: Empty nameString not allowed`

### Fix

Require a non-empty SPN as well.
Verified on a kerberized cluster with the patched hadoop sdk

### DIFF
--- a/sdk/java/src/main/java/io/juicefs/JuiceFileSystemImpl.java
+++ b/sdk/java/src/main/java/io/juicefs/JuiceFileSystemImpl.java
@@ -2340,8 +2340,8 @@ public class JuiceFileSystemImpl extends FileSystem {
   }
 
   public AuthCredential buildAuthCredential(String spn) throws IOException {
-    // auth use kerberos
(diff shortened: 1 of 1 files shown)


tangyoupeng (approved):

kyungwan-nam:
@tangyoupeng @jiefenghuang 
Thanks for the review and the merge!

PR #7463, opened 2026-08-26
cmd/warmup: treat malformed range suffix as path

jiefenghuang:
## Summary

- treat warmup targets that do not match byte-range syntax as literal paths
- keep rejecting syntactically valid ranges with invalid bounds
- add a regression test proving a file named `file [abc]` is processed

## Testing

- `pkg/vfs/fill_test.go`: 28 tests passed

### DIFF
--- a/pkg/vfs/fill.go
+++ b/pkg/vfs/fill.go
@@ -47,16 +47,10 @@ type ByteRange struct {
 // targetRe matches "path [start-end;...]"
 var targetRe = regexp.MustCompile(`^(.*\S)\s+\[([0-9]+-[0-9]+(?:;[0-9]+-[0-9]+)*)\]$`)
 
-// bracketSuffixRe detects a trailing bracket group
-var bracketSuffixRe = regexp.MustCompile(`\s\[[^\[\]]*\]$`)
-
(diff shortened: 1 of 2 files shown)


pkg/vfs/fill.go
```diff
@@ -47,16 +47,10 @@ type ByteRange struct {
 // targetRe matches "path [start-end;...]"
 var targetRe = regexp.MustCompile(`^(.*\S)\s+\[([0-9]+-[0-9]+(?:;[0-9]+-[0-9]+)*)\]$`)
 
-// bracketSuffixRe detects a trailing bracket group
-var bracketSuffixRe = regexp.MustCompile(`\s\[[^\[\]]*\]$`)
-
 // SplitTarget splits a warmup target into its path and byte ranges.
 func SplitTarget(target string) (path, spec string, ranges []ByteRange, err error) {
```
simpleqin, line 51: 直接把返回值err去掉吧，下面那个地方解析失败也可能是文件名的内容

jiefenghuang: vfs中解析也会用到, 防止写错了, 写错的概率比这么特殊的文件名大一些.

zhijian-pro (approved):

PR #7467, opened 2026-08-26
cmd/mount: skip the oom_score_adj adjustment inside a container

### LINKED ISSUE #7466: Add an option to keep the mount process killable by the OOM killer

An option for `juicefs mount` that skips setting `oom_score_adj` to -1000. The default stays -1000, so existing behavior does not change. With the option on, the value comes from the environment, for example pod priority and QoS under juicefs-csi-driver.

**Why is this needed**:

`launchMount` sets `oom_score_adj` to -1000 on the supervisor, and the worker inherits it.

https://github.com/juicedata/juicefs/blob/ef98ec6a8bb9dc31e700bdcf061b6e76f58be725/cmd/mount_unix.go#L980

-1000 removes a task from OOM victim selection. If the cgroup holds only these two processes, the kernel has nothing to kill and repeats `Out of memory and no killable processes` instead of reclaiming. A container is the common case, but any cgroup with a limit behaves the same.

Measured in kind, mount limited to 384Mi. A read load through the mount caused OOM, and in a few minutes:

| | value |
|---|---|
| `oom` in the container `memory.events` | 193,817 |
| `oom_kill` | 0 |
| host `/var/log` growth | 1.5 GiB |
| container `restartCount` | 0 |

The container is still reported healthy, and the kernel log affects the whole node.

Reproduced with 1.4.1 and main (ef98ec6a).

Looka149:
Resolves #7466

### DIFF
--- a/cmd/mount_unix.go
+++ b/cmd/mount_unix.go
@@ -495,10 +495,6 @@ func mountFlags() []cli.Flag {
 			Name:  "disable-transparent-hugepage",
 			Usage: "disable transparent huge page to avoid latency spikes caused by kernel's memory compaction",
 		})
-		selfFlags = append(selfFlags, &cli.BoolFlag{
-			Name:  "disable-oom-protection",
-			Usage: "do not set oom_score_adj to -1000, keeping the mount process killable by the OOM killer",
-		})
 	}
 	return append(selfFlags, fuseFlags()...)
 }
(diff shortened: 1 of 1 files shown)


cmd/mount_unix.go
```diff
@@ -977,7 +981,9 @@ func installHandler(m meta.Meta, mp string, v *vfs.VFS, blob object.ObjectStorag
 }
 func launchMount(c *cli.Context, mp string, conf *vfs.Config) error {
 	increaseRlimit()
-	utils.AdjustOOMKiller(-1000)
+	if !c.Bool("disable-oom-protection") {
```
zxh326:
```suggestion
	if os.Getenv("JFS_INSIDE_CONTAINER") != "1" {
	    utils.AdjustOOMKiller(-1000)
	}
```

Looka149: @zxh326 Thank you for the review.

I confirmed that juicefs-csi-driver already injects `JFS_INSIDE_CONTAINER=1` into the mount container.

zxh326:
LGTM

/cc @jiefenghuang

jiefenghuang (approved):

Looka149:
@zxh326 @jiefenghuang 
Thank you for the review and merge.

PR #7469, opened 2026-08-27
feat(object): Add timeout for cos

201341:


jiefenghuang:
Neither Swift’s Timeout nor its ConnectTimeout is fully equivalent to http.Client.Timeout.

201341:
> Neither Swift’s Timeout nor its ConnectTimeout is fully equivalent to http.Client.Timeout.

fix

PR #7471, opened 2026-08-27
vfs: opt checkReadBuffer to avoid panic

201341:


pkg/vfs/reader.go
```diff
@@ -735,15 +735,18 @@ func (r *dataReader) readBufferUsed() int64 {
 func (r *dataReader) checkReadBuffer() {
 	for {
 		r.Lock()
+		fs := make([]*fileReader, 0, len(r.files))
```
jiefenghuang, line 738: They are all protected by r.Lock, so no changes are needed here.

201341: there is a concurrent problem for map iter and write when call r.unlock() So I think this change is necessary. @jiefenghuang

jiefenghuang: I know. You can write a unit test to verify whether there is actually a problem.

201341: @jiefenghuang ok, I will add a unit test; BTW, we have met this on the Test ENV. This is the stack.    
```goroutine 15770284 [running]:
internal/runtime/maps.fatal({...})
    /root/go/pkg/mod/golang.org/toolchain.../runtime/panic.go:1046 +0x20
internal/runtime/maps.(*Iter).Next(0x4101c73748?)
    /root/go/pkg/mod/golang.org/toolchain.../internal/runtime/maps/table.go:792 +0x98
juicefs/pkg/vfs.(*dataReader).checkReadBuffer(0x40bd4b69c0)
    juicefs/pkg/vfs/reader.go:772 +0xb4
created by juicefs/pkg/vfs.NewDataReader in goroutine 94970
    juicefs/pkg/vfs/reader.go:760 +0x2d4```

jiefenghuang: Are you using the community edition, or have you made some custom modifications? Which version is this? Currently, all accesses/modifications to this map are protected by r.Lock.

201341: Hi, I hava add the unit test to explain this issue. https://github.com/juicedata/juicefs/pull/7475

jiefenghuang: 这个pr也不是证明. 我英文可能没说清楚, 我的意思是你可以看下golang对map并发操作的判断. 你可以很简单的写一个并发单测验证的, -race检测. 这个问题我先关闭了.

PR #7479, opened 2026-08-29
feat(gateway): support If-None-Match wildcard

### LINKED ISSUE #6460: s3 gateway does not support CONDITIONAL WRITE

ilcm96:
## Summary

Add S3 conditional-write support for `If-None-Match: *` to JuiceFS Gateway.

- enforce create-only semantics for PutObject, CopyObject, and multipart completion
- publish new directory markers through a fully prepared temporary inode and `RenameNoReplace`
- preserve the target state when a conditional directory-marker request fails
- atomically create markers on implicit directories with `--head-dir=false`; return `412 Precondition Failed` for existing explicit markers or directories exposed by `--head-dir=true`
- clean up temporary directory-marker inodes on every failed publish path
- add unit, race, and S3 integration coverage, including repeated `If-None-Match` header fields

Fixes #6460.

## MinIO dependency

This PR depends on [juicedata/minio#84](https://github.com/juicedata/minio/pull/84).

The current `go.mod` replacement intentionally points to the pushed feature commit in my MinIO fork so this PR and its CI can be tested before the MinIO PR is merged. After #84 is merged, I will update the replacement to the corresponding `github.com/juicedata/minio` commit before this PR is merged.

## Verification

All verification was run in Docker:

```console
go test ./pkg/gateway -count=1
go test ./cmd -run IfNoneMatch -count=1  # in the MinIO repository
bash -n integration/s3gateway_test.sh
```

The live JuiceFS Gateway checks also verified:

- conditional zero-byte PUT and CopyObject against an implicit directory return `200` with `--head-dir=false`, preserving the directory inode and child data; with `--head-dir=true`, they return `412`
- a new directory marker is created successfully and a duplicate conditional write returns `412`
- repeated raw `If-None-Match` fields are combined correctly: an empty field followed by `*` returns `412`, while `*` followed by an ETag returns `400`
- the existing object content remains unchanged after failed conditional requests


pkg/gateway/gateway.go
```diff
@@ -852,21 +903,169 @@ func (n *jfsObjects) putObject(ctx context.Context, bucket, object string, r *mi
...
+	}
+
+	parent, eno := n.fs.Stat(mctx, parentPath)
+	if eno != 0 {
+		return eno
+	}
+	owner := atomic.AddUint64(&directoryMarkerLockOwner, 1)
+	lockCtx := meta.WrapWithCancel(ctx, mctx.Pid(), mctx.Uid(), mctx.Gids())
+	defer lockCtx.Cancel()
+	if eno = n.fs.Meta().Flock(lockCtx, parent.Inode(), owner, meta.F_WRLCK, true); eno != 0 {
```
myangitzh: Do we need this Flock for the create-only path? The temporary inode plus RenameNoReplace already provides an atomic commit point: one concurrent creator succeeds and the others receive EEXIST. The lock also does not serialize against ordinary PUT/mkdir paths that do not acquire it. And It could potentially block the entire bucket.

pkg/gateway/gateway.go
```diff
@@ -852,21 +903,169 @@ func (n *jfsObjects) putObject(ctx context.Context, bucket, object string, r *mi
...
+			return minio.PreConditionFailed{}
+		}
+		return fmt.Errorf("%s is not directory", directoryPath)
+	}
+	isExplicitMarker := isExplicitDirectoryMarker(&attr)
+	if ifNoneMatch {
+		if isExplicitMarker || n.gConf.HeadDir {
+			return minio.PreConditionFailed{}
+		}
+		return minio.NotImplemented{Message: "atomic conditional promotion of an implicit directory is not supported"}
```
myangitzh: why？If only prefix/child exists, conditional PUT prefix/ returns 501 NotImplemented (or 412 with HeadDir)；however, under S3 semantics, shouldn't this return a 200?

myangitzh:
Hi,Could we add destination If-Match: <ETag> support alongside If-None-Match: *? AWS supports it for PutObject, CopyObject, and CompleteMultipartUpload.
If this is too large for the current PR, could we open a follow-up PR and track it?

S3 User Guide: https://docs.aws.amazon.com/AmazonS3/latest/userguide/conditional-requests.html

pkg/gateway/gateway.go
ilcm96: Agreed. I removed the flock. The prepared temporary inode and `RenameNoReplace` already provide the atomic commit point for creating a new directory marker. Concurrent conditional creators have one winner, and the others return `412`. The concurrency tests pass without the lock.

pkg/gateway/gateway.go
ilcm96: You're right that `prefix/child` does not imply an S3 object at `prefix/`. With the default `--head-dir=false`, conditional PUT of an empty `prefix/` object now returns `200` instead of `501`.

The earlier `501` was a temporary limitation: updating the marker and its xattrs separately could leave partial changes on failure. I added an atomic metadata operation across Redis, SQL, and KV engines that checks and creates the marker while preserving the directory inode and children. Concurrent conditional creators have one winner; subsequent requests return `412`. Conditional copies of zero-byte objects use the same operation.

I kept `412` with `--head-dir=true` to stay consistent with its existing behavior: implicit directories are already exposed as objects through the shared lookup used by HEAD, GET, and copy-source reads. My understanding was that this option intentionally treats implicit directories as existing objects, so I thought the same interpretation should apply to conditional writes as well.

ilcm96:
@myangitzh
> Could we add destination If-Match: <ETag> support alongside If-None-Match: *?

Thanks for the suggestion. I'd prefer to keep this PR focused on `If-None-Match: *` and track destination `If-Match: <ETag>` support separately.

The current implementation atomically checks that the destination does not exist. `If-Match` would also need to compare the destination's ETag and commit the replacement atomically, so a concurrent write cannot invalidate the comparison. That needs additional gateway and metadata work, along with tests for PUT, CopyObject, and multipart completion. I think a separate follow-up would make that easier to review and validate.

pkg/gateway/gateway.go
```diff
@@ -616,7 +634,31 @@ func (n *jfsObjects) CopyObject(ctx context.Context, srcBucket, srcObject, dstBu
 	dst := n.path(dstBucket, dstObject)
 	src := n.path(srcBucket, srcObject)
 
+	if dstOpts.IfNoneMatch && strings.HasSuffix(dstObject, sep) {
```
myangitzh: Directory copies are only handled here when `IfNoneMatch` is set. An unconditional zero-byte copy to `prefix/` falls through to the regular-file path, creating a file instead of a directory marker (or failing if the directory exists)

Should this branch handle all trailing-slash destinations and pass `dstOpts.IfNoneMatch` to `putDirectoryObject`?

ilcm96:
MinIO PR juicedata/minio#84 has been merged. I rebased this PR onto the latest `main` and updated the MinIO pin from my temporary fork to the merged upstream commit (`f2266df40f17`).

pkg/gateway/gateway.go
pai-smyun: Good catch. The directory-object path should not be gated on `IfNoneMatch`; the flag should only control conditional-write semantics within `putDirectoryObject`.

I’ve updated the branch to handle all trailing-slash destinations, while continuing to pass `dstOpts.IfNoneMatch`. The regular-file fallback predates this PR, but reusing the directory-marker path makes conditional and unconditional copies consistent.

I also added regression tests for unconditional zero-byte copies to new markers, implicit directories, and existing explicit markers, plus self-copy coverage for metadata and tags. `go test ./pkg/gateway -count=1` passes.

ilcm96: Good catch. The directory-object path should not be gated on `IfNoneMatch`; the flag should only control conditional-write semantics within `putDirectoryObject`.

I’ve updated the branch to handle all trailing-slash destinations, while continuing to pass `dstOpts.IfNoneMatch`. The regular-file fallback predates this PR, but reusing the directory-marker path makes conditional and unconditional copies consistent.

I also added regression tests for unconditional zero-byte copies to new markers, implicit directories, and existing explicit markers, plus self-copy coverage for metadata and tags. `go test ./pkg/gateway -count=1` passes.

zhijian-pro (approved):

PR #7482, opened 2026-08-31
cache: prevent stale pending writes from deleting rewarmed blocks

### LINKED ISSUE #7464: ci: TestWarmup intermittently fails with missing warmed cache file
**What happened**:

The `unittests (test.cmd)` job intermittently fails in `TestWarmup` after the foreground warmup command reports success.

The failures have the same signature:

```text
warmup cache: 1 files (4 B bytes)
warmup_test.go: warmup: open /var/jfsCache/<uuid>/raw/chunks/0/0/1_0_4: no such file or directory; got content
--- FAIL: TestWarmup
```

Recent occurrences:

- 2026-08-26: https://github.com/juicedata/juicefs/actions/runs/32952914227/job/98128178251
- 2026-07-29: https://github.com/juicedata/juicefs/actions/runs/30419993448/job/90474637051
- 2026-07-12: https://github.com/juicedata/juicefs/actions/runs/29201947487/job/86674714440
- 2026-07-10: https://github.com/juicedata/juicefs/actions/runs/29065915405/job/86277237591
- 2026-06-17: https://github.com/juicedata/juicefs/actions/runs/27709059687/job/81965087291

The issue has appeared in both scheduled and push workflows, so it is not isolated to a single runner or workflow trigger.

`TestWarmup` should be stable. Once foreground warmup reports that the file has been warmed, the test should be able to observe the corresponding cache block.

Run the existing `TestWarmup` repeatedly in the same Ubuntu environment used by `test.cmd`.

The test performs this timing-sensitive sequence:

1. Mount a temporary file-backed JuiceFS volume.
2. Write a 4-byte file.
3. Evict its cache immediately.
4. Warm the same file immediately.
5. Wait for and read the expected disk cache block.

Repeating this sequence under CI load increases the likelihood of reproducing the failure.

- Waiting up to 10 seconds for the cache file did not eliminate the flake; the failures on 2026-07-29 and 2026-08-26 still timed out.
- Warmup reports success and there is no cache write error in the log, but the expected cache file never becomes visible.
- This suggests a synchronization race between the asynchronous disk-cache flush, eviction, and the following warmup request for the same block.
- This issue tracks only the `TestWarmup` CI flake.

qingketsing:
## Summary

Fixes #7464.

`TestWarmup` may fail after the following sequence:

1. Write a small file.
2. Evict its cache immediately.
3. Warm up the same file.
4. Warmup reports success, but the final cache file disappears.

## Root cause

Disk cache writes are asynchronous. Eviction removes `pages[key]` but does not cancel an already queued write.

A subsequent warmup can install a new pending page for the same key. When the old write finishes, it only checks whether the key exists and may delete the new page's marker. The new write then treats itself as evicted and removes the final cache file.

The old write could also temporarily publish a final file after eviction, causing warmup to skip downloading it before the old write removes it again.

## Solution

- Give each pending write a unique entry identity.
- Write each entry to a private temporary file.
- Validate the entry before publishing the final file.
- Perform validation, rename, index update, and marker removal in the same critical section.
- Serialize explicit eviction with pending commit using the same mutex.
- A stale entry now removes only its own temporary file and cannot affect a newer entry.

Page data is still written outside the mutex; only metadata operations such as rename and unlink are serialized.

## Tests

Added deterministic regression tests covering:

- old pending write → evict → new same-key write;
- eviction after the old entry has prepared its temporary file;
- final content, pending marker, memory accounting, and page references.

Verified with:

```text
go test ./pkg/chunk \
  -run '^(TestPendingCacheAfterEvict|TestEvictedPreparedCacheIsNotPublished)$' \
  -count=100

go test -race ./pkg/chunk \
  -run '^(TestPendingCacheAfterEvict|TestEvictedPreparedCacheIsNotPublished)$' \
  -count=20

go test ./pkg/chunk -count=1


jiefenghuang:
Since this issue only affects the warmup command and has limited impact, for now we only need to fix the flaky CI test #7488. There is no need for a larger fix.

PR #7490, opened 2026-09-02
sync: tolerate object stores that do not list keys in lexicographic order

### LINKED ISSUE #7489: sync: whole prefixes are silently dropped when the object store does not list keys in lexicographic order (e.g. TOS buckets with hierarchical namespace)
**What happened:**

Syncing from a Volcengine TOS bucket (hierarchical namespace enabled, accessed through its S3-compatible endpoint) with `juicefs sync --list-threads 2 --list-depth 2 --enable-checkpoint`, the final summary reported only a small fraction of the bucket:

```
Found: 99236, excluded: 0 (0 B), skipped: 5 (0 B), copied: 99229 (2.83 TiB), extra: 0 (0 B), failed: 226
FATAL: failed to handle 226 objects
```

while the bucket holds ~83 TiB. The 226 "failed" entries are not objects but **prefixes** whose listing aborted. None of the objects below them are counted in `Found`, so the shortfall is invisible unless you compare with the console.

The `-v` log shows two error shapes:

```
<ERROR>: The keys are out of order: marker "", last "a/b/dir2/" current "a/b/.cache/huggingface/" [listAll@sync.go:290]
<ERROR>: Failed to restart producer for prefix a/b/: list s3://...: found invalid key a/b/pkg-1.0/ from List, prefix: a/b/, marker: a/b/zzz/ [restorePrefixFromCheckpoint@sync.go:1814]
```

and with delimiter listing on a directory with more than 1000 entries:

```
<ERROR>: list prefix a/pkgs/: list s3://... with delimiter: found invalid key a/pkgs/libwebp-1.2.0-h89dd481_0.conda from List, prefix: a/pkgs/, marker: a/pkgs/libwebp-1.2.0-h89dd481_0/ [startProducer@sync.go:1988]
```

In `--dry` mode the same listing error crashes the process instead:

```
github.com/juicedata/juicefs/pkg/sync.startProducer.func2.1  pkg/sync/sync.go:1989
```

**What you expected to happen:**

Either the sync lists and copies everything, or it fails loudly with a count that reflects the objects that were not synced. A dry run should report listing errors, not panic.

**How to reproduce it:**

1. A TOS bucket with hierarchical namespace, containing a directory whose entries include both `x/` and `x.<ext>` (e.g. a conda package cache), more than 1000 entries in that directory, and some nested directories.
2. `juicefs sync --dry s3://bucket.tos-s3-<region>.ivolces.com/prefix/ <dst> --list-threads 1` → `The keys are out of order` immediately.
3. `juicefs sync --dry ... --list-threads 8 --list-depth 64` on the >1000-entry directory → `found invalid key`, then the nil pointer panic in dry mode.

**Root cause:**

TOS (HNS) returns `ListObjectsV2` results in directory order: entries of a directory come before its sub-directories, and directory names are ordered without the trailing `/`, so `foo/` comes before `foo.conda`. `juicefs sync` assumes strict lexicographic order in three places:

1. `pkg/object/s3.go` / `tos.go` `List()` rejects any key below `StartAfter` with `found invalid key`, even on continued pages where the server follows its own continuation token. `listCommonPrefix` passes the last key of the *client-sorted* page as marker, so a directory with >1000 entries fails as soon as a page boundary falls between `x/` and `x.<ext>`.
2. `listCommonPrefix` appends pages in server order; the per-page sort in the backend does not order entries across pages, and `produce()` merges src and dst by key.
3. `failed` is only created when `!config.Dry`, but listing failures call `failed.Increment()` unconditionally.

Additionally, with `--enable-checkpoint`, a prefix that fell back to flat listing is stored with `list_depth: 0` and the checkpoint key hash does not include `--list-depth`, so re-running with a larger depth restores the old depth and fails again until `--checkpoint-force-reset` is used.

**Environment:**

- JuiceFS version: `juicefs version 1.4.1+2026-07-30.0b90c7d`; the code paths are unchanged on current `main`.
- OS: Ubuntu 24.04, x86_64.
- Source: Volcengine TOS, hierarchical namespace bucket, S3-compatible endpoint (`JFS_S3_VHOST_STYLE=1`). Destination: Tencent COS.

I have a fix for the three points above and will open a PR referencing this issue.

SecretSun:
Fixes #7489

## Problem

Some object stores, notably Volcengine TOS buckets with hierarchical namespace, return `ListObjectsV2` results in directory order: entries of a directory come before its sub-directories, and directory names are ordered without the trailing `/`, so `foo/` comes before `foo.conda`. `juicefs sync` assumed strict lexicographic order in three places. With `--list-depth N` every prefix below depth N fell back to flat listing, hit `The keys are out of order`, and was counted as **one failed prefix** while none of its objects were counted in `Found`. In practice a bucket of tens of TiB was reported as a few TiB with a couple hundred "failed objects" and nothing pointed at the missing data.

## Changes

1. **`pkg/object/s3.go`, `pkg/object/tos.go`**: only validate returned keys against `StartAfter` on the first page. A continuation token is the server's own cursor and `StartAfter` is ignored on continued requests, so the `found invalid key` check was rejecting legitimate pages whenever the server's order differs from byte order. The prefix check is kept.
2. **`pkg/sync/sync.go` `listCommonPrefix`**: sort the whole level once all pages are collected. The per-page sort inside the backend does not order entries across pages, and `produce()` merges src and dst by key, so unsorted input leads to wrong skip/extra decisions.
3. **`pkg/sync/sync.go` `Sync`**: create the `failed` counter regardless of `--dry`. Listing errors call `failed.Increment()` in dry mode too, which panicked with a nil pointer on the first listing error instead of reporting it. The summary already guards on `failed != nil`.

## Test

- New `TestListCommonPrefixSortsUnorderedPages`: a fake store serves two pages that are out of order internally and across pages; asserts files and child prefixes come out sorted.
- `go test ./pkg/sync/ ./pkg/object/` passes.
- Verified against a real TOS HNS bucket: the stock binary fails on a >1000-entry directory in 0.15 s with `found invalid key`; the patched binary lists it and continues into its sub-directories. A previously failing prefix now reports `Found: 15 ... copied: 15`.

## Known limitation (not addressed here)

On checkpoint resume inside a >1000-entry directory, the stored `LastListedKey` is in byte order while such a server interprets `StartAfter` in its own order, so a few entries near the page boundary can be skipped. A follow-up `--dry` run catches them. Also, the checkpoint key hash does not include `--list-depth`, so changing the depth requires `--checkpoint-force-reset`; that may deserve a separate change.

### DIFF
--- a/pkg/object/s3.go
+++ b/pkg/object/s3.go
@@ -263,7 +263,11 @@ func (s *s3client) List(ctx context.Context, prefix, start, token, delimiter str
 		if err != nil {
 			return nil, false, "", errors.WithMessagef(err, "failed to decode key %s", rawKey)
 		}
-		if !strings.HasPrefix(oKey, prefix) || oKey < start {
+		// A continuation token is the server's own cursor and StartAfter is
+		// ignored on continued requests. Some backends (e.g. TOS buckets with
+		// hierarchical namespace) also return keys in a non-lexicographic
+		// order, so only validate against StartAfter on the first page.
+		if !strings.HasPrefix(oKey, prefix) || (token == "" && oKey < start) {
 			return nil, false, "", fmt.Errorf("found invalid key %s from List, prefix: %s, marker: %s", oKey, prefix, start)
 		}
 		objs[i] = &obj{
--- a/pkg/object/tos.go
+++ b/pkg/object/tos.go
@@ -172,7 +172,9 @@ func (t *tosClient) List(ctx context.Context, prefix, start, token, delimiter st
 	objs := make([]Object, n)
 	for i := 0; i < n; i++ {
 		o := resp.Contents[i]
-		if !strings.HasPrefix(o.Key, prefix) || o.Key <= start {
+		// See s3.go: StartAfter is only meaningful on the first page, and TOS
+		// buckets with hierarchical namespace list keys in directory order.
+		if !strings.HasPrefix(o.Key, prefix) || (token == "" && o.Key <= start) {
 			return nil, false, "", fmt.Errorf("found invalid key %s from List, prefix: %s, marker: %s", o.Key, prefix, start)
 		}
 		objs[i] = &obj{
--- a/pkg/sync/checkpoint.go
+++ b/pkg/sync/checkpoint.go
@@ -368,7 +368,23 @@ func (m *CheckpointManager) ValidateConfig(current *Config) bool {
 		return false
 	}
 
+	// PrefixState stores the list depth of every pending prefix, so the prefixes
+	// restored from the checkpoint would keep the old listing mode.
+	if effectiveListDepth(old) != effectiveListDepth(current) {
+		logger.Warnf("Checkpoint config mismatch: list-threads/list-depth, old: %d/%d, current: %d/%d", old.ListThreads, old.ListDepth, current.ListThreads, current.ListDepth)
+		return false
+	}
+
 	return true
+}
+
+// effectiveListDepth returns the depth down to which directories are listed
+// with delimiter; see startProducer.
+func effectiveListDepth(c *Config) int {
+	if c.ListThreads <= 1 || c.ListDepth <= 0 {
+		return 0
+	}
+	return c.ListDepth
 }
 
 // GetOrCreatePrefixState gets or creates a prefix state
--- a/pkg/sync/list_unordered_test.go
+++ b/pkg/sync/list_unordered_test.go
@@ -31,12 +31,14 @@ import (
 // trailing "/" and entries of one directory come before its sub-directories.
 type unorderedStore struct {
 	object.ObjectStorage
-	pages [][]object.Object
+	pages       [][]object.Object
+	startAfters []string // StartAfter received by each List call
 }
 
 func (u *unorderedStore) String() string { return "unordered://" }
 
 func (u *unorderedStore) List(ctx context.Context, prefix, startAfter, token, delimiter string, limit int64, followLink bool) ([]object.Object, bool, string, error) {
+	u.startAfters = append(u.startAfters, startAfter)
 	idx := 0
 	if token != "" {
 		idx, _ = strconv.Atoi(token)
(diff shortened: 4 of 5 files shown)


jiefenghuang:
1. listAll still not work.
2. checkpoint may also have issues.
3. The original listCommonPrefix could already consume too much memory. Maybe we should consider a more complete solution using external sorting. cc @zxh326

SecretSun:
Thanks for the review, agreed on all three points.

1. Right, this PR only fixes the delimiter path (`--list-depth`); `listAll` still requires strict order and will abort on such backends. I should have stated that explicitly.
2. Yes. With a server that does not list in byte order, resuming from `LastListedKey` via StartAfter is not reliable, and the checkpoint key hash also ignores `--list-depth`.
3. Agreed that buffering a whole level in `listCommonPrefix` is a pre-existing memory problem that sorting does not address.

For a complete solution I would propose a sorted-listing layer on top of the existing `pkg/utils/extsort` sharded sorter: consume the listing in server order (no order checks, continuation token only), spill to disk, then emit a sorted stream. Both `listAll` and `listCommonPrefix` would use it, which also removes the in-memory `total`. Since it delays copying until listing finishes, it should be opt-in (e.g. `--sort-list`) or triggered per prefix when out-of-order keys are detected. In that mode checkpoint resume would be per prefix rather than per key, and `--list-depth` should be part of the checkpoint hash.

Would you prefer to keep this PR as the minimal fix (the continuation-token check and the `--dry` panic are independent of the bigger design) and take the sorted-listing work as a follow-up? I am happy to implement it if you agree on the shape, or leave it to @zxh326 if you would rather own it.

zxh326:
Thanks, this fix makes sense. I think it could be merged before the external-sorting work, but the checkpoint issue should be handled first. One option would be to list the prefix from the start again, sort it, and skip keys up to LastListedKey on the client side, rather than using LastListedKey as StartAfter.

SecretSun:
@zxh326 thanks, done in the second commit:

- `listCommonPrefix` no longer passes the resume point as `StartAfter`. It lists the level from the beginning, sorts, and drops keys up to `LastListedKey` on the client side (`sort.Search` on the sorted slice). Later pages still carry the page marker so backends that paginate by marker keep working.
- Every `PrefixState` stores the `list_depth` it was created with, so a checkpoint would restore prefixes in the old listing mode after `--list-threads`/`--list-depth` change. `ValidateConfig` now treats a change of the effective list depth as a config mismatch and starts fresh, consistent with the other option mismatches. Raising `--list-threads` alone (still > 1) does not invalidate the checkpoint.
- Tests: `TestListCommonPrefixResumeSkipsOnClientSide` (server order puts `a/x/` before `a/x.conda`; resume after `a/x.conda` must still yield `a/x/` and `a/x.txt`, and the first `List` call gets an empty `StartAfter`) and `TestCheckpointValidateConfigListMode`.

The flat `listAll` path is unchanged; on such backends it still aborts with "keys are out of order", and I would leave that to the external-sorting follow-up.

SecretSun:
Friendly ping — this has been green and conflict-free for a while.

The checkpoint concern @zxh326 raised is addressed in the second commit (client-side skip on resume, plus list-depth in the checkpoint config check), all 34 checks pass, and the branch still merges cleanly onto current main (#7498 touched a different part of sync.go).

Happy to rebase, split it further, or adjust anything. Just let me know what would make it easier to land.

jiefenghuang:
Sorry for the delayed response. After further discussion, we’ve decided to provide unified and complete support through external sorting in the next release. Therefore, we won’t merge this PR for now.

PR #7498, opened 2026-09-03
sync: avoid excess bandwidth reservations for small objects

### LINKED ISSUE #6627: cmd/sync: ratelimt has the issue of amplification
https://github.com/juicedata/juicefs/blob/e1a269deaf9137aacf8331cfdac801cbc14f6ccb/pkg/sync/sync.go#L709
The `b` is a pool with a size of 1M, and there is a situation of throttling waste when the actual object is less than 1M.

qingketsing:
## What problem does this PR solve?

Sync can reserve 2 MiB of bandwidth tokens when copying a 1-byte object through a destination's 1 MiB buffer: the data read and the final EOF read are both charged by the requested buffer size. Short source reads can also cause repeated reservations.

Fixes #6627.

## What is changed?

- Bound read requests by the known object/part size with `io.LimitReader`.
- Keep unused reservations across short reads and wait only for additional bytes, before reading from the source.
- Perform a single underlying `Read` and preserve its returned byte count and error.

A completed, known-size transfer reserves exactly its size. As with the existing pre-read limiter, unused reservations are not refunded if the source ends early or fails.

## Verification

Two regression tests cover real mem-to-file copying through the 1 MiB buffer, pre-read token reservation, short reads, data returned with an error, and the final EOF. Running these tests against the original implementation, the post-read variant, and the `ReadFull` variant reproduces their respective failures.

Passed:
- Targeted regression tests: `-count=100`.
- Regression and existing limiter failover tests with `-race -count=10`.
- `go test ./pkg/sync -count=1`.

`make test.pkg` was attempted but could not build the Gluster backend because this local environment lacks `pkg-config`.

### DIFF
--- a/pkg/sync/sync.go
+++ b/pkg/sync/sync.go
@@ -758,10 +758,10 @@ type withProgress struct {
 }
 
 func (w *withProgress) Read(b []byte) (int, error) {
-	if limiter != nil {
-		limiter.Wait(int64(len(b)))
-	}
 	n, err := w.r.Read(b)
+	if limiter != nil && n > 0 {
+		limiter.Wait(int64(n))
+	}
 	if copiedBytes != nil {
 		copiedBytes.IncrInt64(int64(n))
 	}
--- a/pkg/sync/sync.go
+++ b/pkg/sync/sync.go
@@ -749,35 +749,19 @@ func doCopySingle0(src, dst object.ObjectStorage, key string, size int64, calChk
 	}
 	r := &chksumReader{in, 0, calChksum}
 	defer in.Close()
-	err = dst.Put(ctx, key, &withProgress{r: r, remaining: size})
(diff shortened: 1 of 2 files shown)


zhijian-pro (approved):

pkg/sync/sync.go
```diff
@@ -758,10 +758,10 @@ type withProgress struct {
 }
 
 func (w *withProgress) Read(b []byte) (int, error) {
-	if limiter != nil {
```
jiefenghuang, line 761: This causes the source-side rate limiting to become ineffective.

qingketsing: 感谢指出。已调整为在读取源端前等待令牌，并通过 io.LimitReader 和 io.ReadFull 避免小对象、短读及 EOF 导致的重复计费，同时保证源端限速仍然有效。麻烦再帮忙看一下。

jiefenghuang (approved):

PR #7503, opened 2026-09-03
utils: avoid racing callback results in WithTimeout

jiefenghuang:
## Summary

Pass callback errors through the completion channel instead of sharing an `err` variable between goroutines. A callback completing after a timeout or cancellation could otherwise race with `WithTimeout` and overwrite the selected error, potentially reporting a canceled operation as successful.

The original `TestTimeout` changes have been reverted. The PR now adds a focused `TestWithTimeoutCancelResultRace` regression test that lets cancellation return first and then allows the callback to publish its late result.

With the pre-fix implementation, the race detector reports concurrent writes from the cancellation branch and callback. With the channel-based implementation, callback results no longer share the selected error and the test passes.

## Tests

- `go test ./pkg/utils`
- `go test -race ./pkg/utils -run "^TestWithTimeoutCancelResultRace$" -count=50`
- Pre-fix verification: the targeted race test fails with `WARNING: DATA RACE` as expected

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -108,25 +108,22 @@ func FindLocalIPs(allowedInterfaces ...string) ([]net.IP, error) {
 }
 
 func WithTimeout(pCtx context.Context, f func(context.Context) error, timeout time.Duration) error {
-	var done = make(chan int, 1)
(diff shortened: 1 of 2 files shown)


myangitzh (approved):
lgtm

PR #7514, opened 2026-09-04
cmd/changelog: support applying changelogs

zxh326:


.github/scripts/changelog_e2e.py
```diff
@@ -0,0 +1,572 @@
+#!/usr/bin/env python3
```
jiefenghuang, line 1: license header

PR #7538, opened 2026-09-15
meta: rebuild counters from load binary backup records

### LINKED ISSUE #7537: Binary load can restore stale volume usage counters
TKV/SQL clients buffer volume usage changes in memory and flush them periodically. A binary backup may therefore contain updated inode metadata while usedSpace and totalInodes still reflect an earlier state.

Binary load restores these counters as-is, leaving usage statistics inconsistent with the restored metadata. JSON load already recalculates them from the restored inodes.

zxh326:
fix: https://github.com/juicedata/juicefs/issues/7537


pkg/meta/base.go
```diff
@@ -4095,6 +4098,21 @@ func (m *baseMeta) LoadMetaV2(ctx Context, r io.Reader, opt *LoadOption) error {
 			return err
 		}
 
+		loaded.updateFromSegment(seg)
```
jiefenghuang, line 4101: loaded.updateFromSegment(seg)放到workerFunc最后处理

pkg/meta/base.go
```diff
@@ -4095,6 +4098,21 @@ func (m *baseMeta) LoadMetaV2(ctx Context, r io.Reader, opt *LoadOption) error {
...
+		if seg.typ == segTypeCounter {
+			for _, c := range seg.val.(*pb.Batch).Counters {
```
jiefenghuang, line 4103: 这段放到updateFromSegment, 然后在errors.Is(err, errBakEOF)时塞入最后的一个counter处理

jiefenghuang (approved):

PR #7539, opened 2026-09-15
Revert "feat(gateway): support If-None-Match wildcard"

jiefenghuang:
Reverts juicedata/juicefs#7479


zhijian-pro (approved):

ilcm96:
Hi, I noticed this revert PR was opened shortly after #7479 was merged. Could you share the reason or concern that prompted the revert?

If there’s an issue with the implementation, scope, or compatibility that I missed, I’d be happy to address it. Knowing the specific concern would also help me understand what should be changed before revisiting this feature.

myangitzh:
Thank you for the contribution and for the work you put into this implementation.

After discussing it within the team, we are concerned that adding SetDirMarker to the metadata layer may increase the long-term maintenance burden and couple the generic metadata interface too closely to object-storage-specific behavior, especially since it is currently used only by the Gateway.

We are exploring whether this functionality can be implemented without modifying the metadata layer. For now, we have temporarily reverted the related changes while we evaluate alternative approaches.

Thank you again for your contribution and understanding.

myangitzh:
We feel the scope of this change is larger than warranted for the problem we’re addressing. Our approach is to solve the common cases with code that is as simple and efficient as possible. We’re choosing not to support some rare corner cases that would require substantially more complexity.

We appreciate the thoroughness of your implementation and the work you’ve put into it.

myangitzh:
After discussion, we would prefer to keep this PR smaller and avoid adding an object-storage-specific primitive to the Meta interface.

Atomically promoting an existing implicit directory to a directory object—for example, `PUT a/b` followed by `PUT` or `CopyObject` to `a/`—does not appear to have a simple implementation using the existing Meta APIs without distributed locking. Supporting this corner case introduces disproportionate complexity and maintenance cost.

For this PR, please keep the temporary-directory + `RenameNoReplace` approach only for destinations that do not exist, and leave the existing implicit-directory promotion case unsupported. If `a/` already exists as an implicit directory, the request should fail explicitly and leave the existing directory and its children unchanged. Please also add a test documenting this limitation.

We can revisit atomic implicit-directory promotion separately if there is sufficient real-world demand.

Note: The current Gateway has limited support for directory objects and does not support directory objects with non-zero content. We may improve these semantics in a future update, but this is currently a low-priority use case, as we have not received any customer requests requiring this behavior.

@ilcm96

myangitzh:
We have introduced an in-process object lock to serialize concurrent writes to the same object within a Gateway instance.

Our design assumption is that all mutating requests for the same object key are consistently routed to the same Gateway instance. This significantly reduces the complexity of distributed consistency, since concurrent operations on the same object can be coordinated locally instead of through the metadata layer.

Based on this design, `SetDirMarker` is no longer necessary. We would prefer to remove it and keep the Meta interface independent of Gateway-specific directory-object semantics.

Concurrent operations on the same key routed to different Gateway instances are outside the scope of this PR. 

Deployments that require this behavior should use sticky routing or consistent hashing to preserve the same-key routing guarantee.

PR #7546, opened 2026-09-16
meta: support changelog scan options

zxh326:
Add From/Follow/Seen options across metadata engines, supporting rewind deduplication and --follow=false.


pkg/meta/sql.go
```diff
@@ -1115,7 +1116,10 @@ func (m *dbMeta) ScanChangelog(ctx Context, last int64, handler func(ver int64,
 		}
 		logger.Infof("last version is %d", last)
 	}
-	seen := make(map[int64]struct{})
+	seen := make(map[int64]struct{}, len(opt.Seen))
```
jiefenghuang: ?

zxh326, line 1119: 这是为了将来 load 备份里保存的数据

jiefenghuang (approved):

PR #7548, opened 2026-09-18
meta: record source engine in binary backups

jiefenghuang:
## Summary

- Record the source metadata engine in binary backup footers.
- Keep dumped counters for Redis backups and rebuild them for SQL/KV backups.
- Preserve existing behavior for older backups without source metadata.

## Tests

- Targeted metadata backup/load tests
- pkg/meta/pb tests
- Lite build


jiefenghuang:
Follow-up to #7538: record the source metadata engine so Redis backups keep their dumped counters while SQL and KV backups rebuild them.

zxh326 (approved):

PR #7570, opened 2026-09-22
meta: fix nlink of cloned directory without recounting

zxh326:
## Problem

When a sub-directory disappears while `juicefs clone` is running, `cloneEntry`
fixes up the nlink of the cloned directory with:

```go
attr.Nlink -= skipped
m.en.doRepair(ctx, ino, &attr)
```

The subtraction is dead code: every engine's doRepair resets attr.Nlink = 2
and recounts by scanning the entries of the directory.

The attr filled by doCloneEntry is also the source's, not the one written
for the clone, so this repair writes the source's parent/uid/gid/mode/times
onto the clone.


pkg/meta/base.go
```diff
@@ -3555,10 +3551,10 @@ func (m *baseMeta) cloneEntry(ctx Context, srcIno Ino, parent Ino, name string,
 		eno = syscall.EINTR
 	}
 
-	if eno == 0 && skipped > 0 {
-		attr.Nlink -= skipped
-		if eno := m.en.doRepair(ctx, ino, &attr); eno != 0 {
-			logger.Warnf("fix nlink of %d: %s", ino, eno)
+	if n := atomic.LoadUint32(&nlink); eno == 0 && n != attr.Nlink {
```
jiefenghuang: eno == 0 && nlink != attr.Nlink ?

pkg/meta/sql.go
```diff
@@ -5507,6 +5509,7 @@ func (m *dbMeta) doCloneEntry(ctx Context, srcIno Ino, parent Ino, name string,
 		if n.Type == TypeFile && n.Nlink > 1 {
 			n.Nlink = 1
 		}
+		m.parseAttr(&n, attr)
```
jiefenghuang: move to the end?

jiefenghuang (approved):

PR #7573, opened 2026-09-22
object/nfs: check Mount error before using the target

### LINKED ISSUE #7572: object/nfs: nil pointer when NFSv3 mount fails
`newNFSStore` uses the result of `mount.Mount` before checking its error.

`github.com/vmware/go-nfs-client` returns `(*Target, error)` and the target is nil when the NFSv3 MOUNT procedure fails (`MNT3ERR_ACCES`, `MNT3ERR_NOENT`, protocol not supported, and so on). The current code writes `target.Config.DirCount` and `target.Config.MaxCount` first, so `juicefs format --storage nfs` panics:

```
github.com/juicedata/juicefs/pkg/object.newNFSStore
    pkg/object/nfs.go
```

This hides the real error, which should be `unable to mount <host>:<path>: ...`. `DialMount` already checks its error before use; `Mount` should do the same.

Seen with juicefs 1.3.0 (`1.3.0+2025-07-03.30190ca`) and still present on main. The kernel can mount the same export as NFSv4, while this backend only speaks NFSv3, so a v3 rejection becomes a panic instead of an error.

AlanJager:
## Summary

`mount.Mount` returns a nil `*nfs.Target` when the NFSv3 MOUNT call fails. `newNFSStore` set `target.Config` before checking that error, so `juicefs format --storage nfs` panicked instead of returning `unable to mount`.

Move the error check ahead of the `target.Config` writes. `DialMount` already follows this order.

Fixes #7572

## Test plan

- [ ] `juicefs format --storage nfs` against an export that rejects NFSv3 prints `unable to mount` and exits, with no nil-pointer panic
- [ ] `juicefs format --storage nfs` against a working NFSv3 export still formats


jiefenghuang (approved):

jiefenghuang:
Hi, please sign the cla.

AlanJager:
Manual check of this patch (`f68cf9e`), built locally as `juicefs version 1.5.0-dev+unknown`.

NFSv3 mountd is up, but the export is not allowed for the client, so `mount.Mount` returns `MNT3ERR_ACCES`. Before this change that nil target was dereferenced and `format` panicked. With this change:

```
$ juicefs format --storage nfs --bucket 127.0.0.1:/tmp/nfs-export sqlite3:///tmp/jfs-nfs-manual.db manualtest
2026/09/22 18:19:01.737621 juicefs[820798] <INFO>: Meta address: sqlite3:///tmp/jfs-nfs-manual.db
2026/09/22 18:19:01.744900 juicefs[820798] <FATAL>: object storage: unable to mount 127.0.0.1:/tmp/nfs-export: MNT3ERR_ACCES
```

Exit code 1. No panic.

PR #7574, opened 2026-09-22
chunk: stop patching getDiskUsage while cache loops are running

AlanJager:
## Summary

`unittests (test.pkg)` on #7573 failed in `pkg/chunk`, which that PR does not touch. `TestCleanupFullDoesNotBlockLoad` rewrites `getDiskUsage` with gomonkey while `checkFreeSpace` loops from earlier `newDiskCache` calls are still inside it. The process then faults:

```
unexpected fault address 0x2e3be4f
fatal error: fault
[signal SIGSEGV: segmentation violation code=0x2 addr=0x2e3be4f pc=0x2e3be4f]
github.com/juicedata/juicefs/pkg/chunk.getDiskUsage
    pkg/chunk/utils_unix.go:40
```

Call `getDiskUsage` through an `atomic.Value`, and have the tests replace that value. Callers for other cache directories keep the real `statfs`.

## Test plan

- [x] `go test -count=1 -run 'TestCleanupFullDoesNotBlockLoad|TestUploadStagingToFreeCalculation|TestUploadStagingInodeToFree|TestSpaceToFreeNoAction|TestDiskCacheState' ./pkg/chunk/`


jiefenghuang:
It’s just a flaky unit test. We’ll address these issues together later.

